PR #2, opened 2021-09-13
Swedish translation

mickemartinsson:


wills106 (approved):
Thanks for the help with translations.

PR #8, opened 2022-01-04
Fix run mode code for Off mode + fix typo

martin-klima:
I found a mistake in run mode. When run mode code == 3, then the inverter is Off.
Number 3 is the correct value for Feedin Priority work mode (Charger Use Mode), not for inverter status.

The screenshot shows the wrong run mode when I switched the inverter off.

image


wills106:
I'll have to double check "Run Mode == 3" as according to the Modbus Protocol Doc's for Gen2, 3 & 4 they all show it should be Fault Mode. Although looking back at my Packages I have it as "Off"
Not sure where the "Feedin Priority" came from? Unless I copied it from Charger Use Mode and forgot to edit the name?

PR #9, opened 2022-01-07
Fix Inverter power scale for each phase

martin-klima:
This PR fixes the inverter power scale for each phase.

Values for:
- Inverter Power R (register address 0x6C or 108)
- Inverter Power S
- Inverter Power T
loaded from Modbus registers are in Watts and should not be scaled by 0.1.

After this fix the formula `Inverter power R + S +T =  Inverter Power` is valid 
(`grid_power_r + grid_power_s + grid_power_t = inverter_load`).

This PR has been tested on SolaX Hybrid X3.

image

### DIFF
--- a/solax_modbus/__init__.py
+++ b/solax_modbus/__init__.py
@@ -761,8 +761,11 @@ class SolaXModbusHub:
         grid_current_r = decoder.decode_16bit_int()
         self.data["grid_current_r"] = round(grid_current_r * 0.1, 1)
         
+        # @todo Rename variable as it this is the invertor power on phase R, not the grid power.
+        #   The grid power is currently named as feedin_power_(rst)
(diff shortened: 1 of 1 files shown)


wills106:
@mickemartinsson Are you able to check if your Inverter Power is reporting correctly / incorrectly? Your Inverter is a Gen3 isn't it?

@martin-klima Is your inverter a Gen 2 or 3? As I know on the X1 they scale different between Gen 2 and Gen 3

mickemartinsson:
@wills106 I have a X3-Hybrid-G1/G2 and the Inverter Power is, what I can see, not correct. It is too low and the sum doesn't add up.

martin-klima:
@wills106 My model is X3-Hybrid-8.0-D-E, it is reported as X3-Hybrid-G1/G2 type in the Solax Cloud.

PR #11, opened 2022-01-21
Fix user limit for export control

martin-klima:
The value in the "Export control user limit" Modbus registry is stored as (Power / 10), for example, 5200 W is stored as 520.
Therefore we must multiply the registry value by 10 to put the correct human-readable value to the sensor.solax_export_control_user_limit sensor.

Tested on SolaX X3-Hybrid-G1/G2.
I'm not sure if other SolaX types have the same problem.

### DIFF
--- a/solax_modbus/__init__.py
+++ b/solax_modbus/__init__.py
@@ -351,7 +351,7 @@ class SolaXModbusHub:
         self.data["export_control_factory_limit"] = round(export_control_factory_limit * 0.1, 1)
         
         export_control_user_limit = decoder.decode_16bit_uint()
-        self.data["export_control_user_limit"] = round(export_control_user_limit * 0.1, 1)
+        self.data["export_control_user_limit"] = round(export_control_user_limit * 10, 1)
         
         eps_mutes = decoder.decode_16bit_uint()
         if eps_mutes == 0:


wills106:
This is a difference between Gen2 and Gen 3.
I will need to take an approach as used https://github.com/wills106/homsassistant-solax-modbus/blob/8ff2ab2c592c45a242600aadc4606860b9c1e253/solax_modbus/__init__.py#L238-L241

But with the value of 10 and 0.1 depending on the GenX version with a new variable name.

wills106:
Yea I need to implement variables more, especially now we have Gen 4.
Saves having so much repeated code, see below.

https://community.home-assistant.io/t/solax-inverter-by-modbus-no-pocket-wifi-now-a-custom-component/140143/337

martin-klima:
Reg. variables for different models) I'm not a Python programmer, I'm PHP (and Drupal) developer, so I don't know the best practices in Python. From my experience, I would consider using some kind of YAML config files for each Gen.

We could have Modbus register addresses, multiplicators, friendly names for all sensors in the YAML file. Parse the file to an array and work with the array in the Python code. In this way, we would maintain and modify mostly only YAML configs in the future when we add some new sensors or react on some changes. Also, we would be ready for the next generation.

Example of YAML config for specific Gen:
```yaml
input sensors:
  export_control_user_limit:
    name: "Export Control User Limit"
    native_unit_of_measurement: POWER_WATT
    register: 0xFF 
    scale: 10
    enabled: False
    some_extra_property: 12345 
```

Reg. export_control_factory_limit) I will check it ASAP and let you know.

martin-klima:
The "SolaX Export Control Factory Limit" with 0.1 factor looks good.
image

wills106:
I was trying to avoid any config files or extra yaml setup.
I was thinking of keeping the variables in the _init__.py / const.py file and then depending on which Tickbox the user selects at the config setup determines what's visible and what multipliers to use for the given sensors, so they read correctly depending on the Inverter version they have.

Sadly I seem unable to pull versions out of the inverter. The registers that are meant to provide some of the info don't return anything.

martin-klima:
It looks like the inverter type is propagated by type ID on SolaxCloud. In my case the type = 5.
image

SolaxCloud has a JSON library of all inverter types. There is the following part for value = 5 and it matches the type displayed on the SolaxCloud page (X3-Hybrid-G1/G2).

```json
{
            "id": 546530971697020928,
            "sid": "546530971697020928",
            "dictCode": "INVERTER_TYPES",
            "lang": "en_US",
            "code": "X3-Hybrid-G1/G2",
            "value": "5",
            "sort": 5,
            "remark": ""
        },
```

So, maybe we could try to search registers (or something else) for the type ID.

I pasted the whole Inverter type list here: https://pastebin.com/UVCygynh

martin-klima:
I found how it works. The SolaxCloud recognizes the invertor type from its serial number.

It POSTs the serial number to URL: https://www.solaxcloud.com/phoebus/device/getTypeModelBySN with payload `inverterSN=H3UE0xxx...`

It responses:
```json
{
    "id": 5,
    "inverterSN": null,
    "inverterType": 5,
    "phaseFlag": 3,
    "battery": 1,
    "pvCount": 2,
    "classify": 0,
    "epsStatus": 1,
    "roleFlag": 0
}
```

It means they probably have all serial numbers with basic configuration data in their database and that's why inverters don't have it in the Modbus registers.

We could simulate the communication above, but it requires an authentication token and it is generated during successful login to the SolaxCloud. A little bit difficult, but not impossible.

wills106:
Is the 10th digit of your inverter 5 then?
Just trying to work out if there is any relation to the numbers in the Serial number.

martin-klima:
Unfortunately, my inverter SN doesn't contain the number 5. I had the same idea yesterday :disappointed:

wills106:
This should be corrected in 0.4.0 if you are still having wrong scaling could you please open a new pull based on the new code. Or let me know which lines need correcting.

wills106:
When you setup this component, do you select Gen3 X3?
I might need to add in a Gen2 X3 option for this sensor.solax_export_control_user_limit

martin-klima:
> When you setup this component, do you select Gen3 X3?

Yes, I think so. If I remember, it was only one option with X3, that I understand as a 3-phases inverter.

wills106:
Can you try again on an up to date release.
If your "export_control_user_limit" is still out I might need to add an "Gen2 X3" option, which is a bit annoying just for one register.

wills106:
If you use the standard integration setup as Gen3 X3 are you still having scaling issues for the "export_control_user_limit" sensor?

Where 5200W is represented as 520W?

PR #12, opened 2022-01-21
Add Battery Awaken control entity

martin-klima:
This PR creates a select box for controlling Battery Awaken.
I'm not sure it is the best approach for controlling this entity, maybe some switch or button would be better, but this simple solution works for me and could be useful for others.

image

Tested on SolaX X3-Hybrid-G1/G2.

### DIFF
--- a/solax_modbus/const.py
+++ b/solax_modbus/const.py
@@ -147,6 +147,14 @@ SELECT_TYPES = [
             3: "Both Allowed",
         }
     ],
+    ["Battery Awaken",
+        "battery_awaken",
+        0x56,
+        {
+            0: "Disable",
+            1: "Enable",
+        }
+    ],
 ]
 
 @dataclass


wills106:
I'm going to close this PR down as the code has changed that much since this was raised.

Have you had this reliably wake the inverter up when the battery is discharged?
I am sure I have tried this register in the past and it didn't do anything. If it has worked for you I'll add it in. Perhaps it only works on certain models / Firmware versions?

martin-klima:
> Have you had this reliably wake the inverter up when the battery is discharged?

Yes, I use it very often and it works. When I set the minimum battery level and the inverter reaches the level, it goes to "Idle mode". If I want to wake up the inverter when the sun is not shining, I must switch the Battery Awaken to Enable for some seconds.

Try it yourself.

I found the Modbus register for battery awaken when I analyzed the communication between the internet browser and SolaxCloud server.

PR #50, opened 2022-04-07
Phase 1 of code simplification : sensor entities

infradom:
This is a major code restructuring, but should keep exactly the same semantics. It has been tested on Gen4, so verification on other generations is necessary.
In this proposal, all sensor entities are declared in one long flat list, but each entity has an additional allowedtypes attibute.
This allowedtypes attribute declares to which inverter types the attibute applies (combination of generation, X1orX3, HybridOrAc, EPS)
By using this method, and a matchInverterWithMask function, the sensor.py code has also been greatly simplified

In a later phase, number.py and select.py can be simplified in a similar way


infradom:
Will replace this pull request whith an updated one (as I forgot some EPS flags)

PR #51, opened 2022-04-08
Restructure code to avoid code duplication (phase 1: sensor entities)

infradom:
This is a rather major code restructuring, but should keep exactly the same semantics. It has been tested on Gen4, so verification on other generations is necessary.
In this proposal, all sensor entities are declared in one long flat list, but each entity has an additional allowedtypes attibute.
This allowedtypes attribute declares to which inverter types the attibute applies (combination of generation, X1orX3, HybridOrAc, EPS)
By using this method, and a matchInverterWithMask function, the sensor.py code has also been greatly simplified

In a later phase, number.py and select.py can be simplified in a similar way


infradom:
I added a small change that allows multiple inferters to coexist on the same ip address; 
this works as long as the modbus address or tcp port is different

wills106:
I'll have a proper look over this in the next day or two and try it out.

Might need to have a think about the X1-AC as it doesn't have the PV registers. It's essentially a battery charger / manager only.

infradom:
As you will see, the framework is prepared for the AC, we just need to specify the  HYBRID or AC flag for the affected sensors

infradom:
One question may be relevant for the future work on this: from which piece of code would you read the serial number? In my opinion, the unique_id should ideally be derived from the serial number and not from the IP address as we do now. I had to extend this already a bit to support multiple inverters behind the same IP (change 10c16f3)

wills106:
I'll push this out as a beta / prerelease then people can try it out and still have the option roll back to 0.4.13a and at the same time know it's not "Production Ready"

infradom:
Thanks ! ,
by the way:  I am almost ready with the similar restructuring of numbers and select declarations

wills106:
I have just moved over and it all seems ok so far.

PR #53, opened 2022-04-09
final phase of code restructuring, fix button bug

infradom:
This should go first into the prerelease version 
adapt button.py code to new model
fix inverter unlock and battery awake problem (the removal of the comma should also go into the main line to fix a serious bug)
adapt init.py to new inverter model detection method


wills106:
Just cloned your fork into my HA and all the buttons, number, selects & sensors appear to be exactly the same as before.
So I can't see anything accidently broken on a Gen3 X1

PR #57, opened 2022-04-11
declare most inverters as hybrid and remove old code

infradom:


infradom:
I cannot explain what you experienced, but the code should look like in this new PR

PR #68, opened 2022-04-26
Correct error in newly created PV bit: the hex value should only contain one bit.

infradom:
Correct error in newly created PV bit: the hex value should only contain one bit.
0x0600 contains 2 bits  (0x0400 and 0x0200)which will lead to problems.
I restructured somewhat to resolve this issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Any manual updates / HACS updates require a restart of Home Assistant to take ef
 - Any major changes might require deleting the Integration from the Integration page and adding again. If you name the Integration exactly the same including the Area if set, you should retain the same entity naming bar any name changes in the release. (Refer to the release notes for any naming change)
 
 If the Integration fails to load with the following error in your log "unrecognized inverter type - serial number : {your_serial_number_here}"
-Please see #26 providing the details asked for.
+Please see [Discussion #26](https://github.com/wills106/homsassistant-solax-modbus/discussions/26) providing the details asked for.
 
 <B>Inverter never connected to the Cloud / possibly after a firmware update:</B>
 - If you can read values, but unable to adjust select / number you need to press the "Unlock Inverter" button.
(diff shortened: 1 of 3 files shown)


wills106:
Sorry for messing up again. I just thought things were spaced out by 200 or 400 so it was easy to read.

I need to go and read how bitmasks actually work.

infradom:
No problem, I feel a bit guilty for using such an 'old style' construct. (But it is very efficient)
Things would have been easier to understand if I had expressed the values in binary instead of hexadecimal:

The same declaration would then look like:

```
HYB_GROUP_BITS = 0b0001110000000000 # hex 0x1C00
PV             = 0b0000010000000000 # hex 0x0400
AC             = 0b0000100000000000 # hex 0x0800
HYBRID         = 0x0001000000000000 # hex 0x1000
```

Hope this helps, let me know if it is not clear enough ..

wills106:
So would this be used as a very condensed set of rules or commands for example a "Bad Guy" sprite would stick to.

Sprite can fly up and down
but not right to left
sprite can shoot red bullets but not blue

But if you hit a check point one of the bits would get changed so it now can shoot blue bullets instead of red?

So this would be used on early 8bit or 16bit games to save disk space.

infradom:
Sorry, I cannot follow ;-) , but yes, I started programming over 45 years ago, and although python has for almost everything a nice approach, as far as I now, for these bit flags, the old fashioned way looked nicer to me.

If you want, I can restructure and put these attributes in a python dictionary, but that will not be very efficient and the __init__.py code might look uglier when we check the type of a device:

wills106:
I think the Inverter uses a similar approach to some of the error codes.

If you read a register and you get 0000 at the binary level you get no error
0001 = error A
0010 = error B
0100 = error C
1000 = error D

But you never get an error like 0110
The Inverter can only respond with a single error

wills106:
Ok that makes sense.

Item A = 0100
Item B = 1000

Item A doesn't match Item B

But you can call Item A & B by combining them together into 1100

I am sorry if that doesn't translate. But I think I am starting to understand how it works and the relations between numbers.

infradom:
A correction to your example: 

Item A = 0100
Item B = 1000

A | B  is equal to 1100 ; this is a bitwise or
A&B is equal to 0000 ; this is a bitwise and

Another example:

MASK = 1110
C = 0001
A & MASK is equal to 0100
B & MASK is equal to 1000
C & MASK is equal to 0000

PR #100, opened 2022-08-08
Correct sleeping MIC behaviour, correct button indentation error

infradom:
Just a suggestion for MIC, feel free to solve it differently (see @kars remarks)
Correct button indentation error

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -1292,50 +1292,21 @@ class SolaXModbusHub:
         )
         
         pv_voltage_1 = decoder.decode_16bit_uint()
-        self.data["pv_voltage_1"] = round(pv_voltage_1 * 0.1, 1)
-        
         pv_voltage_2 = decoder.decode_16bit_uint()
-        self.data["pv_voltage_2"] = round(pv_voltage_2 * 0.1, 1)
-        
         pv_current_1 = decoder.decode_16bit_uint()
-        self.data["pv_current_1"] = round(pv_current_1 * 0.1, 1)
-        
         pv_current_2 = decoder.decode_16bit_uint()
-        self.data["pv_current_2"] = round(pv_current_2 * 0.1, 1)
-        
         grid_voltage_r = decoder.decode_16bit_uint()
-        self.data["grid_voltage_r"] = round(grid_voltage_r * 0.1, 1)
-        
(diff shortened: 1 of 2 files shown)


wills106:
Do you think this will also help with [this](https://github.com/wills106/homeassistant-solax-modbus/issues/88#issuecomment-1207287934) or is that more likely because the USB device can't communicate with anything? So it ends up spamming the USB Port?

wills106:
Sorry pressed wrong button 🤣

infradom:
> Do you think this will also help with [this](https://github.com/wills106/homeassistant-solax-modbus/issues/88#issuecomment-1207287934) or is that more likely because the USB device can't communicate with anything? So it ends up spamming the USB Port?

I dont think it is related to the USB/serial ussue

custom_components/solax_modbus/__init__.py
```diff
@@ -1345,32 +1316,40 @@ def read_modbus_input_registers_mic(self):
...
+            self.data["feedin_power"] = output_power
+
+            self.data["feedin_power_r"] = output_power_phase_r
+            self.data["feedin_power_s"] = output_power_phase_s
+            self.data["feedin_power_t"] = output_power_phase_t
+            self.data["pv_power_1"] = pv_power_1
+            self.data["pv_power_2"] = pv_power_2
+            self.data["pv_total_power"] = pv_power_1 + pv_power_2
+            self.data["total_yield"] = round(total_yield * 0.0001, 2)
+            self.data["today_yield"] = round(today_yield * 0.001, 2)
```
Kars-de-Jong, line 1353: This one keeps updating, even when the inverter is in run mode Waiting. I see it go to 0 at around 0:00.

infradom: Thanks Kars, but can you be more precise? About which one are you talking? Only today_yield or also other variables? We disabled them all for the time being, and in the beta7  we event dont try to read them if run_mode is not "waiting"

Kars-de-Jong: I indeed meant `today_yield`, but there are indeed more values that are also measured when the inverter is waiting, or are at least OK when showing 0, these are all of them:

- Today’s yield
- DC voltages
- DC and AC currents
- DC and AC powers

Which leaves measurements to be disabled when waiting:
- AC voltages
- AC frequency
- Total yield

I think I have them all now…

PR #107, opened 2022-08-14
enhance exception handling for modbus errros

infradom:


wills106:
Would it be best to print out what the actual exception is? We might be able to work out what the error is, assuming it returns something that makes sense?

infradom:
> Would it be best to print out what the actual exception is? We might be able to work out what the error is, assuming it returns something that makes sense?

This code will already print out the exeption details, it is not recommended to pass the exception object to the logging call as this may fail in case of non-asciii characters

see https://riptutorial.com/python/example/19233/logging-exceptions

PR #128, opened 2022-10-05
Fix typo when logging host validation errors.

ppetru:


wills106:
Thanks for spotting the error.

PR #224, opened 2022-11-29
Add some GEN4 modbus power control related read registers

infradom:


wills106:
It will be returning the value of 6A which is reserved.

infradom:
ok, sorry I was not aware I changed this ...

infradom:
You are right, the 0x disappeared

Issue #234, opened 2022-12-08
[BUG] Can't update config of the integration (@wills106 Potential related to 2022.12 config flow)


johny-mnemonic:
I wanted to change the polling frequency from 15 to 10 seconds, so I opened the integrations page and clicked the config button, but when I pressed send button it failed with "unknown error occurred".

**Mandatory details**
1. Integration Version: 0.6.4
2. HA Version: 2022.12.0
3. Inverter Brand: Solax X3 Gen4 10kW
4. Plugin Used:
5. First 6 digits of Serial: H34A10
6. Connection Method: ModBus TCP using WaveShare RS485 TO ETH (B)

**Detailed Error Log**
```
Logger: aiohttp.server
Source: helpers/schema_config_entry_flow.py:165
First occurred: 20:03:38 (6 occurrences)
Last logged: 20:08:00

Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib/python3.10/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.10/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 60, in security_filter_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/forwarded.py", line 100, in forwarded_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 28, in request_context_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 81, in ban_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 236, in auth_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/view.py", line 136, in handle
    result = await result
  File "/usr/src/homeassistant/homeassistant/components/config/config_entries.py", line 241, in post
    return await super().post(request, flow_id)
  File "/usr/src/homeassistant/homeassistant/components/http/data_validator.py", line 73, in wrapper
    result = await method(view, request, data, *args, **kwargs)
[...]

PR #236, opened 2022-12-09
Adapt config_flow to support 2022.12 as well as older versions

infradom:


infradom:
should solve issue #234

PR #265, opened 2022-12-19
First version with Gen4 remote power control - probably still immature

infradom:
Also: allow partial reconfiguration of integration


infradom:
Documentation: see wiki page

Issue #269, opened 2022-12-21
Missing radiobuttons on configuration


gelse:
**Describe the bug**
When creating a new config (after removing previous installation / upgrade), the radio buttons for the inverter type are missing, thus leading to invalid user input:
`User input malformed: value must be one of [] for dictionary value @ data['plugin']`

**Mandatory details**
1. Integration Version: 2022.12.12 (also tried 2022.8 and main)
2. HA Version: 2022.12.7

**Detailed Error Log**
there is no error log, it is an UI error.

<img width="260" alt="image" src="https://user-images.githubusercontent.com/49122643/208943649-71a518e5-472e-4fea-9494-3e800a29fce9.png">

gelse:
as a fallback it would be nice to have a how-to somewhere how i can configure the integration via YAML.

wills106:
Have you upgraded from 0.6.4 or earlier without removing the Integration from the Integration page, restarting HA and adding it back to the Integration page?

gelse:
no, i followed the instructions by the book.  
i removed the integration from HA
i removed the integration from HACS
i restarted HA
i added the integration in HACS
i restarted HA again
i tried to add he integration in HA

gelse:
and also, as i said in the description above, there are NO errors of solax in the logs of HA.

wills106:
You are somehow missing the whole section from the config flow where you select which plugin you want.

On the Integration page have you tried forcing a refresh?
Ctrl + F5
Then try adding it again?

gelse:
yes, that was the first thing i tried (i am a developer myself ;-) ) - i just re-tried it, same behaviour.

wills106:
I have just done a fresh HACS download on a separate machine and I can't replicate this.

Has HACS downloaded the plugin_sofar plugin_solax files etc?

gelse:
after removal via HACS:
```
root@omv:~/links/homeassistant/config/custom_components# ls
alexa_media  edgeos  elasticsearch  hacs  monitor_docker  myjdownloader  openmediavault  spotcast  xiaomi_cloud_map_extractor  xiaomi_miot
```
then adding it via HACS:
```
root@omv:~/links/homeassistant/config/custom_components# ls
alexa_media  edgeos  elasticsearch  hacs  monitor_docker  myjdownloader  openmediavault  solax_modbus  spotcast  xiaomi_cloud_map_extractor  xiaomi_miot
```
and the content of solax_modbus seems fine to me.

[...]

PR #271, opened 2022-12-22
made paths to plugins calculated instead of fixed

gelse:


gelse:
bugfix for https://github.com/wills106/homeassistant-solax-modbus/issues/269

infradom:
I believe we need to more, and avoid that the persistent config_entry variables contain paths.
I am trying a version that probably doesnt need paths .., but if that fails, your proposal could certainly be part of the solution

infradom:
This has been implemented in a PR that also modifies the config_flow entries to  avoid paths in the entry
Thanks !

wills106:
I'll close down as #272 resoles the same issue.

PR #272, opened 2022-12-22
make plugin method more path-independent 

infradom:
Thanks to [@gelse](https://github.com/gelse)
Also made it non-breaking change (config_flow)
fixes [#269](https://github.com/infradom/homeassistant-solax-modbus/issues/269) and replaces PR [#271](https://github.com/infradom/homeassistant-solax-modbus/issues/271)

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -49,11 +49,11 @@ from .const import (
     DEFAULT_PORT,
     DEFAULT_BAUDRATE,
     DEFAULT_PLUGIN,
-    PLUGIN_PATH,
+    #PLUGIN_PATH,
     SLEEPMODE_LASTAWAKE,
 )
 from .const import REGISTER_S32, REGISTER_U32, REGISTER_U16, REGISTER_S16, REGISTER_ULSB16MSB16, REGISTER_STR, REGISTER_WORDS, REGISTER_U8H, REGISTER_U8L
-from .const import setPlugin, getPlugin, getPluginName
+#from .const import setPlugin, getPlugin, getPluginName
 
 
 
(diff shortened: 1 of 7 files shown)


custom_components/solax_modbus/__init__.py
```diff
@@ -88,13 +88,20 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
-    plugin_name = getPluginName(plugin_path)
-    plugin = importlib.import_module(f".plugin_{plugin_name}", 'custom_components.solax_modbus') 
-    if not plugin: _LOGGER.error(f"could not import plugin {plugin_name}")
-    setPlugin(name, plugin)
+    _LOGGER.info(f"Ready to load plugin {config[CONF_PLUGIN]}")
+    plugin_name = config[CONF_PLUGIN]
+    # convert old style to new style plugin name here - Remove later after a breaking upgrade
+    if plugin_name.startswith("custom_components") or plugin_name.startswith("/config") or plugin_name.startswith("plugin_"):
+        plugin_name = plugin_name.split('plugin_', 1)[1][:-3] 
+        _LOGGER.warning(f"converting old style plugin name {config[CONF_PLUGIN]} to new style short name {plugin_name}")
```
gelse, line 96: shouldnt you also change it in the config after conversion ?

infradom:
@gelse Thanks for this suggestion, I know how to change config when we are in module config_flow.py, but when a user just upgrades the integration, afaik, we never enter the config_flow.py code. From within __init__.py, I am unsure how to modify the config_entries. 
This is why I left it as it is and a cleanup can be done when we have a real breaking update.
Feel free to suggest how we can update config entries from within __init__.py however ..

gelse:
i think it is described here:
https://developers.home-assistant.io/docs/config_entries_config_flow_handler/#config-entry-migration

infradom:
Thanks a lot !! I was not aware of that. I will adapt and retest (retesting is the most annoying part)

PR #280, opened 2022-12-27
Misc smaller changes (unrelated to each other)

infradom:
Feel free to keep the old TIME_OPTIONS code


wills106:
I'm happy with the time changes, it makes it easy to change the resolution ie 5mins.
Obviously you end up with a very long select list!

PR #296, opened 2023-01-14
Add PV to Heatsink Temperature 1

darkrain-nl:
Heatsink Temperature 1 is supported on my PV device, therefore add PV to the allowedtypes


wills106:
Thank's the PV support was a little bit guess work, so there may be other sensors applicable to your PV Only Inverter.

PR #304, opened 2023-01-23
Add 2 remotecontrol modbus power  modes: Self_Use and Feedin_Priority emulations

infradom:


custom_components/solax_modbus/__init__.py
```diff
@@ -227,14 +227,14 @@ def saveLocalData(self):
 
     def loadLocalData(self):
         try: fp = open(self._hass.config.path(f'{self.name}_data.json'))
-        except: pass
+        except: _LOGGER.error(f"cannot open config file {self.name}_data.json, will retry later")
```
wills106, line 230: Had to change this to _LOGGER.info was filling my error log with a new entry every second.

infradom: Ok thanks, I understand, will search a better solution  for non-gen4 systems

Issue #307, opened 2023-01-25
Suggestions for improvements on Solis inverters


i-jk:
Hello, I'm using `plugin_solis` and have noticed a minor bug, plus have some suggestions to make...

**Bug:** The Inverter Storage Control sensor doesn't show the name of the mode name, just the number. Perhaps this is due to a type mismatch (should be `fmt = "i"`)? Note this same issue also affects the mode select widget, which is blank (or rather shows the placeholder text).

**Suggestion: fix charge and discharge current limitation of 70A**
The `native_max_value` of several number sliders for the charging and discharging current are limited to 70A. This is obviously dependent on the specification of the inverter, e.g. my 5kW 48V hybrid should be 100A. I wonder if it's possible to either:
1. establish the inverter specs from the serial and set these accordingly
2. use the appropriate `BMS` value for charge max and discharge max, when available, or fallback to safe 70A.

**Suggestion: number fields instead of sliders?**
TBH the sliders are rather fiddly to use on larger current numbers like 100A due to the limited space HA provides such widgets... A simple number field might be better, is this possible?

If you agree with the above, perhaps we can give the hour/minute timed charge and discharge values the same treatment? Can we group the hour and minute fields together to display like a proper time? This is a nice to have, obviously :slightly_smiling_face:  

Thanks again for such an awesome integration! :raised_hands:

wills106:
> Inverter Storage Control sensor doesn't show the name of the mode name, just the number.

Looks like there is a potential mismatch between the rear sensor and the select entitie.

> charging and discharging current are limited to 70A

This was only set to 70A as that's what's in the Modbus docs.
If we get enough data I can use an approach like used on the SolaX Inverters:

https://github.com/wills106/homeassistant-solax-modbus/blob/4264bc6e0c84e96b9232c1611f0dcf7e0d064aeb/custom_components/solax_modbus/plugin_solax.py#L242-L244

> TBH the sliders are rather fiddly to use on larger current numbers like 100A

HA itself automatically configures sliders or number boxes on it's own.
I think there is an option to override this automatic allocation, I can have a look into it.

> Can we group the hour and minute fields together to display like a proper time?

There isn't actually support for the equivalent to an input.date_time for Integrations.
You should be able to create a time helper and have the Integration use that I believe. But I haven't found an example on how to do this yet.

i-jk:
>> Inverter Storage Control sensor doesn't show the name of the mode name, just the number.

[...]

PR #362, opened 2023-03-20
fix incorrect bms battery current scale

wawyed:
Fixes https://github.com/wills106/homeassistant-solax-modbus/issues/307#issuecomment-1475333256


wawyed:
Thank you! <3

PR #369, opened 2023-03-30
Solis status codes

kyuzumaki:
I added all documented status codes for solis inverters. It seems to work for me I don't have any way to test them all but hopefully will flag up any faults that occur. I've copied the wording from the documentation.


wills106:
Thanks, I'm going to clean the naming a bit so they are more consistent as there is a mix of abbreviations and upper / lower case in the Solis docs.

PR #380, opened 2023-04-12
Add Grid monitoring into Solax plugin

johny-mnemonic:
There is no Grid state and voltage monitoring in the integration. This kind of info is very useful for automations that should act when there is Grid outage.


johny-mnemonic:
I am not sure about the `allowedtypes` value. It works with my X3-Hybrid-Gen4 10kW, but I don't know which other Solax lines are compatible, hence I limited it only to the tested one.

wills106:
I'll have a review tomorrow to see which other Inverters this applies to.

wills106:
Just had a look and these registers are only on the Gen4.
Aren't they a duplicate of "Inverter Voltage", "Inverter Current" though? Other than being able to see the individual phases? Although I can imagine seeing the individual phases being useful if you need to load balance across the phases.

johny-mnemonic:
No, these are basically values from the grid meter. So apart from the "Inverter Voltage" which shows zero when the inverter goes idle, these voltages still show voltage in the grid.
Voltage per phase is definitely nice-to-have kind of addition for few edge cases like you describe.
I know that few times we experienced grid outage only on single phase, so I wanted to have this covered, but if there is any downside adding them I can remove them.

Issue #264, opened 2022-12-19
[FR] Support solarmanv5 protocol


alienatedsec:
**Describe the bug**
Cannot make it working for some reason.

**Mandatory details**
1. Integration Version: 2022.12.08
2. HA Version: 2022.12.4
3. Inverter Brand: RHI-6K-48ES-5G Single
4. Plugin Used: What is that?
5. First 6 digits of Serial: 603105
6. Connection Method: Wi-Fi (TCP)

**Detailed Error Log**
```2022-12-19 14:24:27.150 INFO (MainThread) [homeassistant.setup] Setting up solax_modbus
2022-12-19 14:24:27.151 INFO (MainThread) [homeassistant.setup] Setup of domain solax_modbus took 0.0 seconds
2022-12-19 14:24:27.202 INFO (MainThread) [custom_components.solax_modbus.config_flow] detected HA core version 2022 12
2022-12-19 14:24:27.203 INFO (MainThread) [custom_components.solax_modbus.config_flow] starting configflow - domain = solax_modbus
2022-12-19 14:24:27.231 INFO (MainThread) [custom_components.solax_modbus.plugin_solis] Solaxv01: trying to determine inverter type
2022-12-19 14:24:32.245 WARNING (MainThread) [custom_components.solax_modbus.plugin_solis] Solaxv01: reading serial number from address 0x80ec failed; other address may succeed
2022-12-19 14:24:32.249 INFO (MainThread) [custom_components.solax_modbus.plugin_solis] Read Solaxv01 0x80ec serial number: None, swapped: False
2022-12-19 14:24:32.254 ERROR (MainThread) [custom_components.solax_modbus.plugin_solis] Solaxv01: cannot find serial number, even not for other Inverter
2022-12-19 14:24:32.254 ERROR (MainThread) [custom_components.solax_modbus.plugin_solis] unrecognized Solaxv01 inverter type - serial number : unknown
2022-12-19 14:24:32.277 INFO (MainThread) [homeassistant.components.sensor] Setting up sensor.solax_modbus
2022-12-19 14:24:32.305 INFO (MainThread) [custom_components.solax_modbus.sensor] computedRegs: {}
2022-12-19 14:24:33.262 INFO (MainThread) [homeassistant.components.button] Setting up button.solax_modbus
2022-12-19 14:24:33.270 INFO (MainThread) [homeassistant.components.number] Setting up number.solax_modbus
[...]

PR #413, opened 2023-05-21
Add compatibility for Modbus RTU over TCP / Ascii over TCP

mludolph:
This change introduces the possibility to connect to RS485 adaptors that do not translate Modbus RTU to Modbus TCP but instead to Modbus RTU over TCP, such as the USR-TCP232-304 mentioned in the wiki. This was necessary to make the adapter work.


wills106:
Thanks, I was expecting USR-TCP232-304 to only provide a virtual com port when I briefly looked at it. Unless that's Windows only?
Have you come across any adaptors using the ModbusAsciiFramer ?

wills106:
Do you have access to any other RS485 adaptors?
I'm just wondering how selecting ModbusRTUFramer on a Waveshare for example would behave.
I only have an Ethernet connection on my Inverter, so I have no need for an adaptor and nothing to connect one to.

I'll need to be clear with the documentation in the Wiki so users don't select the wrong configuration now there is 3 TCP options.

mludolph:
I did get the virtual COM port to work under windows but wasnt able to do so for ubuntu (tried using socat and mounting it into the docker container did not work). 

I do not have access to other adaptors nor have I seen one using ASCII over TCP but I figured since the application that I initially used for testing (Modbus Poll) provided these options, I might aswell include that option in case someone comes across such a device.

I do believe however that using Modbus RTU over TCP would not work e.g. using a Waveshare device since these seem to provide real Modbus TCP although I cannot test this.

Let me know in case there's anythihg I should change to make it less confusing for other users.

alienatedsec:
> I do believe however that using Modbus RTU over TCP would not work 

Modbus RTU over TCP is the SolarMan v5 format that is used with Solis DLS-W (now obsolete) data loggers. There is great documentation about it, and initially I was trying this method with SolaX integration.

jmccrohan:
> > I do believe however that using Modbus RTU over TCP would not work
> 
> Modbus RTU over TCP is the SolarMan v5 format that is used with Solis DLS-W (now obsolete) data loggers. There is great documentation about it, and initially, I was trying this method with SolaX integration #264
> 
> `[EDIT]` More about SolarMan protocol FYI - https://pysolarmanv5.readthedocs.io/en/latest/solarmanv5_protocol.html

Solarman loggers don't use Modbus RTU over TCP. The Solarman V5 protocol is different. However, I've just tidied up and published something I wrote a while back to encapsulate Modbus RTU over TCP frames in Solarman V5 frames. If it is of use, I'll look to package it up as a Home Assistant Addon.

https://github.com/jmccrohan/pysolarmanv5/blob/main/examples/solarman_proxy.py

PR #535, opened 2023-09-18
Fixed Long term statistics in error case

wrfz:
I have a Sofar hyd 15ktl-3ph and Waveshare RS485 to RJ45 Ethernet in use.
Occasionally I get "read failed at ..." errors (especially when the polling frequency for modubus registers is set to 15 seconds or lower).

This PR does not solve the cause, but the consequence errors!

In case of a failed communication (**read failed at ... 0x{descr.register:02x} ...**), **val = 0** was set.
This caused the long term statistics for entities with **state_class** = **total_increasing** to get messed up.
Statistics->**Sum** was calculated incorrectly. 
The result was that the energy dashboard was also always messed up.

grafik

The return value **None** now ensures that HA receives the value **Unknown** and no longer falsifies the **Sum**.

grafik

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -445,6 +445,7 @@ class SolaXModbusHub:
 
     def treat_address(self, decoder, descr, initval=0):
         return_value = None
+        val = None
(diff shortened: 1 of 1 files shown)


wills106:
Thanks.
This should also hopefully fix the similar issue Solis owners get, when setup with multiple Modbus devices to allow cloud and local connection.

PR #561, opened 2023-10-16
Modify firmware entries

cewbdex:
Distinguish between ARM and DSP version and add major version from G4 document. If those changes are not intended for previous generations, I will make them only for G4.


cewbdex:
Found it's relevant only for G4, now it's probably ready to merge.

PR #562, opened 2023-10-17
Add shadow fix option to Solax

cewbdex:


wills106:
I think the name and key should match with:

https://github.com/wills106/homeassistant-solax-modbus/blob/317455e04d8d17211869e1437ebc958de5531ea0/custom_components/solax_modbus/plugin_solax.py#L2238-L2248

So the values update when HA restarts. Although it looks like key to 0x104 should be updated to match the working of the name.

cewbdex:
Right, I did not know it works like this, thanks.

wills106:
Even though the sensor is disabled (register = 0x9C,) HA still reads that to populate the select on start up of HA, or if you manually change something on the Inverter itself. So the name / key needs to match. I'm not 100% sure which is the critical one that needs to match.

cewbdex:
image
image

wills106:
Oops got the read and the write register back to front in my reply.

But in the Integration the names need to match.

PR #563, opened 2023-10-17
Add minimal suported homeassistant version requirement to hacs.json

cewbdex:


wills106:
Been a while since I last looked at the HACS document, didn't realise you could specify a minimum HA version.

PR #576, opened 2023-10-23
Refactor README.md III

cewbdex:
In my subjective view it looks better, but it's up to you to decide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,25 +78,6 @@ SolaX Power
     </ul>
 </ul>
 
-You can have multiple instances of this Integration, just change the default Prefix from SolaX to something else. Ie. `SolaX Main` or `SolaX Southwest`.
-
-<details>
-<summary>
-SolaX - PocketLAN & PocketWiFi Readme
-</summary>
-
-**(Pocket LAN does not provide a Modbus connection at all and the Pocket WiFi v1 & v2 does not provide a reliable Modbus connection! Trouble shooting for Pocket WiFi v1 or v2 will not be provided)**
-
-#### Pocket WiFi 3.0 with Firmware V3.004.03 and above is only [officially supported](https://kb.solaxpower.com/data/detail/ff8080818407e2a701840a22dec20032.html). SolaX only mentions Gen4 Hybrid, other inverters may work?
-- **Contact SolaX for latest version.**
-
-(⚠I still don't recomend the PocketWiFi. If you loose all entites after normal operation, try power cycling your Inverter.
-
-Another approach to fixing previously working PocketWiFi installs is to restart your rooter and then reload the integration in Home Assistant. If that doesn't work you can unplug PocketWifi usb for 30 seconds and plug it in again, then reload the integration. 
-
-Updating / downgrading the integration or Home Assitant won't help, you have lost the internal Modbus connection between the Inverter and PocketWiFi and I am unable to assist with this issue.⚠)
-</details>
-
 <details>
 <summary>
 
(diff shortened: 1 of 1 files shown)


wills106:
I think the PocketLAN / PocketWiFi info should stay near the top. The FAQ isn't exactly read very often.

cewbdex:
I was thinking that it would be better to have something like github.io pages, now something is in readme and something in wiki and it's kinda confusing

wills106:
github.io isn't something I have looked into.
But if it allows for other people to submit PR's for doc changes then it would be good. That's the main problem with the wiki there is only two of us that can update it and neither of us own non SolaX Inverters.
So there is limited documentation on non SolaX equipment.

cewbdex:
Wiki could be moved to readthecods, [see here my test repo](https://rtdtestingelfs.readthedocs.io/en/latest/index.html)

wills106:
I got as far as https://github.com/wills106/homeassistant-solax-modbus/tree/main/docs
Which can be found https://wills106.github.io/homeassistant-solax-modbus/

But at the time, I ran out of time to actually populate anything.
I haven't dealt with github.io or readthedocs.io before, which you do think offers an easier way for others to contribute?

cewbdex:
I believe readthedocs is better, because is has a search functionality, which can make things easier. GitHub pages might be easier for contributors to contribute, as it uses markdown instead of restructuredtext, but the syntax is really simple too. Both have sources hosted on GitHub. I could try to move our existing wiki if you wish and then transfer ownership of the repo to you, we can have a new GitHub organisation to keep things organized.

cewbdex:
Wiki is there: https://testmdp2.readthedocs.io/en/latest/intro.html

cewbdex:
I am closing this, please review https://testmdp2.readthedocs.io/en/latest/ entire wiki is there, only some fixes needed, if you like it, we can move it to a new address and repo. Only install tutorial is missing there.

wills106:
Noticed you have a repository of Firmwares for the Various models.

Have you come across anymore Changelogs for the X1 Hybrid Gen3 other than the changelogs in the Excel sheets?

wills106:
Is the Syntax used on that readthedocs example compatible with GitHub / github.io or does it need conversion?

I think short term it might be best to put it in https://github.com/wills106/homeassistant-solax-modbus/tree/main/docs
Then we can at least get some contributions from owners of other Inverters. The biggest issue with the current Wiki is that people can't contribute. So the likes of Solis, Growatt etc have zero documentation.

I need to have a think on longer term branding of the Integration as it isn't SolaX exclusive anymore so a bit missleading.

cewbdex:
I have done some investigation and I found readthedocs supports markdown, so it's almost 1:1, with some addition, I think it would be better to use RTD rather than GH pages, question is, do you want it to be hosted in this or separate repo? This is the code, you can see it's very simple, ignore repo name: <https://github.com/cewbdex/example-jupyter-book>

wills106:
I think for the time being it's best staying in this repo. At least people can contribute to it for now.
Giving people the ability to easily submit doc updates is more important, as it will improve peoples experience with the Integration.

Like I say longer term I will look into a new Integration name and how to host things better. The Integration name needs to be more universal.

cewbdex:
I missed your first comment. The repository is not mine, it was told to me that SolaX owns it, but I can't guarantee it, but multiple people tested it with multiple inverters and no problems so far.

wills106:
No worries, I'll reach out to SolaX and see if there is any documentation on the changes.
I'm curious to know what's been changed in the various releases.

cewbdex:
Maybe we could host somewhere a repo with all modbus documents we have, so it's easier to collaborate.

PR #590, opened 2023-11-02
Add Battery Charge and Discharge Limits

fboundy:
Hi - this PR just adds a couple of registers to Solis inverters that were missing previously.


wills106:
@fboundy I am guessing this is a limit you can actually change on the Inverter? If so I would expect a corresponding number entity to enable you to change it.

Isn't this a duplication of

https://github.com/wills106/homeassistant-solax-modbus/blob/aa628317c6392eb342791b8b241d758c12f66f72/custom_components/solax_modbus/plugin_solis.py#L629-L658

Or do they return a different value?

wills106:
If they are a static value of the max you can set them too I am wondering if they could be used to dynamically build:

https://github.com/wills106/homeassistant-solax-modbus/blob/aa628317c6392eb342791b8b241d758c12f66f72/custom_components/solax_modbus/plugin_solis.py#L169-L177

PR #595, opened 2023-11-06
Add SolaX bug form

cewbdex:
Forms provide richer experience over templates. They are easier to fill by users, can provide them more information and will force them to fill required information. Tags need to be created in order to be applied. My vision is to have per brand issue form, so it can be tailored exactly, but that's just my point of view and I could provide a generic one instead.

### DIFF
--- a/.github/ISSUE_TEMPLATE/solax.yml
+++ b/.github/ISSUE_TEMPLATE/solax.yml
@@ -3,7 +3,7 @@ description: File a bug report about Solax inverter
 title: "[Bug]: "
 labels: ["bug", "triage", "Solax"]
 projects: []
-assignees:
+assignees: []
 body:
   - type: markdown
     attributes:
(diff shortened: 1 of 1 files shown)


wills106:
I think at the moment the form can stay generic. The only thing that would need changing is:

`label: First 6 digits of Serial`

`label: First 6 digits of Serial (Growatt Firmware Version)`

and perhaps make the `label: Inverter firmware versions` as I don't have any relevant information on other Brands.

Or another option is to keep solax.yml and have a second generic.yml that's a bit more looser for other Brands?

I think it's good though that you can set mandatory and optional Fields. I didn't realise this was possible.

cewbdex:
I guess one for Solax and one generic for others would be fine. If I understand you correctly, the serial number is only relevant for Growatt? Or what do you mean by the suggested change? Mandatory and optional fields are only possible in forms, not templates (the current new issue approach). Here is how it will look like: https://github.com/cewbdex/homeassistant-solax-modbus/issues/new?assignees=&labels=bug%2Ctriage%2CSolax&projects=&template=solax.yml&title=%5BBug%5D%3A+

wills106:
SolaX, Solis and Sofar use Serial Number to identify.
Growatt uses the Firmware version

> Mandatory and optional fields are only possible in forms

I was referring to your new form sorry, I loaded it up to look and it looks good.

wills106:
Yea just done a bit of housekeeping.

PR #597, opened 2023-11-07
Enable RTC and Battery Power

fboundy:
I'm working to integrate the Solis inverters with BatPred and it needs these two entities enabled. It can be done manually but having them enabled by deault would be helful.


cewbdex:
Your PR does not contain any changes, what is the benefit for normal user for enabling it by default? IMHO just unnecessary clogging of recorder.

cewbdex:
> Hi @cewbdex
> 
> You either new to the community or taking your role too serious.

Hello @alienatedsec 
I don't know which role do you mean, I don't have any. I have just reviewed a PR of integration I use, as I wish it works as good or better as now.

> The PR says what it's for and @fboundy has an idea, which would be beneficial to the community without additional steps. Take it or leave it but it's already in the release.

Ready for release? Did you even look at the changes? There are none. I have just asked what are the changes for, I can see he stated some BatPred, but I have no idea what it is and that's why I asked the exact need of these entities. RTC just seems arbitrary for me. "beneficial" is a relative term, it may be beneficial for him, but not for people that will not use it (probably the majority for now?), as it will eat more of their resources. I am not saying that this contribution is useless or anything like that, only that I don't see personally any benefit over the downsides. Also I don't see anything to the topic in your comment? In the end my review may not matter anything, as it's up to the integration authors to decide, so even if I said that this should not be merged, I don't see a problem there.

alienatedsec:
> RTC just seems arbitrary for me. "beneficial" is a relative term, it may be beneficial for him, but not for people that will not use it (probably the majority for now?)

@cewbdex I am not sure how to respond. 

I will only say that I am looking forward to @fboundy BatPred integration with SolaX as I have no time for the work to commence, and it was on my list to do for a while now. 

I am happy to share some other work that integrates Sunsynk-power-flow-card (frontend energy dashboard) with SolaX. 

image

fboundy:
Thanks for highlighting that I had reverted the commit before submitting the PR. I’ve closed this and will submit another.

Given how often Solis click times drift or don’t follow DST changes I would say that RTC is of general use but if you are trying to automate charging with time-based tariffs it is very useful.

And battery power seems like a generally useful thing to have. If you are concerned about DB size disable them or reduce your update frequency. Or customise Recorder

PR #598, opened 2023-11-07
Default Enable RTC and Batter Power on Solis

fboundy:
I am attempting to integrate this integration with [BatPred]( https://github.com/springfall2008/batpred ) which requires both inverter clock and battery power entities. These are currently disabled by default. This PR simple makes them enabled by default.

Given how often Solis inverter time drifts or does not reset with DST I think this is useful in of itself. Similarly it seems to me that Battery Power should be a basic sensor that is available.


wills106:
Looking at the other PR I believe and agree RTC should remain disabled by default. It's a Database clogger which doesn't really provide any value.

I would advise an automation to resync the RTC Weekly or Monthly would be a better option.

I can look to see if you can add an entity that doesn't write to the Database, as it's been a while since I have looked at entity configurations. But I still wouldn't class it as an entity that needs to be enabled by default.

"Battery Power" Could have been disabled by mistake or I thought it was a duplicate. It's hard to know what's useful and what's not when I don't own a particular inverter.

fboundy:
Fine. I’ll add a note to the config file for BatPred to say that it will need to be enabled. Not sure you can have entities that don’t populate the database unless you filter them separately in Recorder.

Battery Power - it’s disabled and current is enabled: personally I’d have it the other way round but also not a big issue.

When I get a chance I’ll have a check and see if there any others that are missing that I’ve previously pulled in directly using the Core ModBus service. If they’re not critical I’ll add them as Disabled.

cewbdex:
I guess battery power is disabled, because there are already battery input and output energy, however I don't really understand why energy has watt unit, am I missing something or is it a bug? If that is the case I think battery power should be enabled (looked in the file and didn't find other similar entity) and unit of energy changed.

@fboundy 
Does HA energy dashboard work for you with the battery?

```
    SolisModbusSensorEntityDescription(
        name = "Battery Power",
        key = "battery_power",
        native_unit_of_measurement = UnitOfPower.WATT,
        device_class = SensorDeviceClass.POWER,
        state_class = SensorStateClass.MEASUREMENT,
        register = 33149,
        register_type = REG_INPUT,
        unit = REGISTER_S32,
        allowedtypes = HYBRID,
        icon = "mdi:home",
    ),
    SolisModbusSensorEntityDescription(
        name = "Battery Input Energy",
        key = "battery_input_energy",
        native_unit_of_measurement = UnitOfPower.WATT,
        device_class = SensorDeviceClass.POWER,
        state_class = SensorStateClass.MEASUREMENT,
        value_function = value_function_battery_input_solis,
        allowedtypes = HYBRID,
        icon = "mdi:battery-arrow-up",
    ),
    SolisModbusSensorEntityDescription(
        name = "Battery Output Energy",
        key = "battery_output_energy",
        native_unit_of_measurement = UnitOfPower.WATT,
        device_class = SensorDeviceClass.POWER,
        state_class = SensorStateClass.MEASUREMENT,
        value_function = value_function_battery_output_solis,
        allowedtypes = HYBRID,
        icon = "mdi:battery-arrow-down",
    ),
```

fboundy:
I’ve only just started using this integration - previously I just read the inverter directly using the core ModBus service. I’ll have a look at the Energy dashboard and check.

wills106:
Just had a look at this.
"Battery Power" was named that as that's how it was in the docs, something like Battery Energy Flow or something would have been better.
(When I do the initial setup of a new plugin I just go off the docs as I have no way of knowing exactly what any of them do. I'm also reluctant to rename sensors after a while incase it breaks peoples automations / messes up people's long term stats, unless a request to change a name is early into the development of a plugin.)

It's a bi-directional sensor and a lot of the Energy Dashboards wanted oneway sensors so "Battery Input/Output Energy" was created and "Battery Power" was set to disabled by default as it then became a duplicate.

fboundy:
OK - that makes sense. What confused me was that the existing entities are called Energy when they are actually Power... I see that they have `device_class` power and `unit_of_measurement` of W so I think they should really be renamed to `battery_input_power` and `battery_output_power`

wills106:
I think I named the input/output after the SolaX naming convention, which came from the Modbus docs for SolaX.

I think if the key e.g. `key = "battery_input_energy",` remains the same you can change `name = "Battery Input Energy",` without having the entities becoming unknown and having to delete them.

But any changes will probably break peoples long term statistics. So we would have to think very carefully how to reduce that risk.
But I'll be honest some of the entity naming convention isn't the greatest. It's also difficult to harmonise across the various plugins when different Inverter Brands may name them something different on the Inverter interface.

fboundy:
Understood. I can update the config for PredBat to use the existing entities so I'll close this PR.

PR #599, opened 2023-11-07
Add more entities for SolaX G4

cewbdex:
Don't merge them, published this for review while I test it. I won't implement safety code write register, as there is no reliable way to obtain all IDs. Not sure why are two shadow fix functions present, added the second.


cewbdex:
There is a second EPS frequency field, this is the one used currently:
image
This is the one I add:
image
The second is `0x03:Read Holding Register` and the currently used is `0x04:Read Input Register`. What is the difference between them?

I will probably remove the second GMPPT, as I have enabled it on inverter and the one currently used got value changed.
Currently used addresses:
image
image

The `2`:
image
image

The other entities I added:
image
image
image

wills106:
Which safety code are you missing?

cewbdex:
This work is a bit stale, as I believe some entities I added/want to add need newer inverter FW.

It's not correct, just a mistake, because I added a new eps frequency entity and found there is one already at different address and the test was supposed to be the one I added.

Safety code my inverter returned was `42`. I believe that as they get upgraded (Czech got from 2021 to 2022), new number is used, but can't say for sure yet.

wills106:
If the Inverter is missing a new register from a Firmware update it should just return zero. It shouldn't cause the Integration to fail.

So I can add this in and clean it up.

When you read `27 07 26` Indicated [here](https://github.com/wills106/homeassistant-solax-modbus/blob/main/docs/solax-faq.md#how-to-find-firmware-version)
Which entity returns the 27 & 26 I'm going to add formatting so it's 1.27 and 1.26

cewbdex:
It does not fail, it returns zero and the description of the modbus address is not clear to me and I don't know what does it really mean and check it now.

I don't want to merge this yet, but I didn't find how to make this a draft (is it just a name?). I plan to add some more and remove the extra ones. Also we should make a wiki section for developers and put our knowledge here, so it's easier to catch up for newcomers.

<https://github.com/wills106/homeassistant-solax-modbus/blob/main/custom_components/solax_modbus/plugin_solax.py#L1845>
<https://github.com/wills106/homeassistant-solax-modbus/blob/main/custom_components/solax_modbus/plugin_solax.py#L1782>
And we can then remove the major, as people report that it returns zero.

wills106:
I have added in "Safety code" into my commit.

Will look at the rest later.

cewbdex:
Where did you get the other codes?

wills106:
I was able to view them, but it's not a document I can share unfortunately.

cewbdex:
Ah, is it some unobtanium directly from SolaX? Or if I ask them, I can get one too, if I won't share it? 
This is kinda tricky, my inverter has 2021 and it returns 2022, may be better to just return Czech?

wills106:
Perhaps on your firmware the Safety code was a first revision like a draft? Czech Government body revised it in 2022 and in later firmware SolaX they updated the naming to comply with the final version?

I'm not an installer so it doesn't really mean much to me. I know some countries have regulations on Anti-Islanding and so on...

cewbdex:
I have removed the extra stuff and now only the mergeable things remain. I would like to test the EPS frequency, how it differs from the one we have already. There is an issue requesting a second GMPTT entity, but there was no response, so I don't know if it solves the problem (inverter I have access to either does not have this entity or the FW is too old). Also I have found that there is GMPTT entry for Gen3, however they have different values, so unifying them is probably not a good idea?

wills106:
What do those registers represent? Remaining Battery capacity?

HA-Supi:
> What do those registers represent? Remaining Battery capacity?

Remaining Charge or Dischargeable Battery Capacity, just as the Name of the Entitity. I read in Modbus Doku 3.21 about them. 

And because I had an RMA Qcells Hyp G3 Battery T30 (Entity "SolaX Battery Input /Output Energy Today" reports only 5.0 kWH usable Battery instead beforde 5,5-6) i just want to monitor as much Entitys as possible.

By the way:  Many Thanks for your Addon!

martintothgit:
grafik
I want to add some information regarding ShadowFixFuncEnable2. I have a Solax X3 Hybrid G4 Inverter and it has 2 MPPT Trackers. So i would assume there's for each Tracker a dedicated setting - as you see in the screenshot from Inverter Settings (above).

wills106:
The first one is already added it's called "Shadow Fix Function Level (GMPPT)"

HA-Supi:
> The first one is already added it's called "Shadow Fix Function Level (GMPPT)"

Hello Wills106, 

Yes. Thats correct. As in Modbus Documentation  Register 0x009C mGlobalMPPTFunc(X1). We just need a newer one than V3.29.  
Because with solax app you can setup all strings separately. 

And at the moment, when i enable Shadow Fix the status of this register "solax_shadow_fix_function_level_gmppt" changes in  "unknown".
image








https://kb.solaxpower.com/solution/detail/ff80808185aa327a0186396bc59d018a

HA-Supi:
> The first one is already added it's called "Shadow Fix Function Level (GMPPT)"

Exactly, PV1 shadefix is change with modbus setting via HA addon.
So we have to figure out what command the solax app sends to the inverter for PV2. 

In 2022, the solax app and/or inverter X3 Hybrid g4 showed only shadefix as one setting for both strings. With later firmware both PV1/PV2 could be set.

martintothgit:
There's also a bug in the enumeration for shadow_fix_enable. Entry "High" should have value 3. Now "Middle" and "High" have the same value (2).:

```
        name = "Shadow Fix Function Level (GMPPT)",
        key = "shadow_fix_enable",
        register = 0x104,
        scale = { 0: "Off",
                  1: "Low",
                  2: "Middle",
                  2: "High", },
        entity_registry_enabled_default = False,
        allowedtypes = GEN4,
    ),
```

Maybe you can also add ShadowFixFuncEnable2 (0x114) with one of the next releases?

wills106:
I have just released [2024.01.3](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.01.3) with these changes

martintothgit:
I updated to 2024.01.3 and the 2 items for shadow fix are working fine now. Thanks a lot!
(I changed parameters from solaxcloud and compared values in home assistant and vice versa. I tested all enumerations.)

wills106:
Thanks for the feedback.

HA-Supi:
Can also confirm that PV 1 and PV2 can be set separately. 
@wills106 : you have the newest Solax documentation 3.29?

wills106:
> Solax documentation 3.29?

Yes I do.

I added in the battery capacity entities by the way. Not sure if you noticed?

HA-Supi:
> > Solax documentation 3.29?
> 
> Yes I do.
> 
> I added in the battery capacity entities by the way. Not sure if you noticed?

Thanks, I noticed these new entities. Although the Photovoltaic community in Germany rethinks these entities as not relevant regarding battery health. 

But nevertheless, 

thanks for this great HA Addon. 


https://www.photovoltaikforum.com/thread/214494-fehlende-max-kapazität-und-kapazitätsverlust-beim-q-home-ess-hyb-g3-3p/?pageNo=1

PR #601, opened 2023-11-07
Use dropdown menu for plugins in bug templates

cewbdex:
Since there are only a few options, it might be more comfortable to have a list of them. Also in my opinion labels should be lowercase, changing them is still an option, because they do not exist yet, thus no duplication.


cewbdex:
The latter commit is a proposal, I think more people will see it there and the section from `README.md` could be removed.

PR #607, opened 2023-11-10
Solis modes

fboundy:
Adds some Modes to the Solis Energy Control Switch that were missing

### DIFF
--- a/custom_components/solax_modbus/plugin_solis.py
+++ b/custom_components/solax_modbus/plugin_solis.py
@@ -709,11 +709,13 @@ SELECT_TYPES = [
         option_dict =  {
                 1: "Selfuse - No Grid Charging",
                 3: "Timed Charge/Discharge - No Grid Charging",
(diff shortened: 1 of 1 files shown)


wills106:
Out of curiosity have you come across any newer docs than the Coghlan one? Quite hard to imagine Solis haven't moved on from a 2018 document.

fboundy:
> Out of curiosity have you come across any newer docs than the Coghlan one? Quite hard to imagine Solis haven't moved on from a 2018 document.

Nothing official from Solis AFAIK

fboundy:
> These new entries need adding to:
> 
> https://github.com/wills106/homeassistant-solax-modbus/blob/aa628317c6392eb342791b8b241d758c12f66f72/custom_components/solax_modbus/plugin_solis.py#L1236-L1252
> 
> Otherwise if you make a change on the Inverter, HA won't correctly update the select option with what it's now set to.

OK - added to the PR

PR #608, opened 2023-11-10
Add readthedocs (wiki) files

cewbdex:
You need to sign in for RTD, I don't know if there is any compability with GH pages.


cewbdex:
It's compatible, but I didn't find any good GH pages skin for a nice wiki, also additional configuration would be probably needed, with RTD it's most plug and play now.

Btw. are the images in images folder used anywhere or can be cleaned up?

cewbdex:
Worked days and nights and the wiki is now feature complete. Awaiting feedback.

wills106:
Thanks for taking the time to convert the wiki into a document format that will be easier for others to contribute to.

cewbdex:
Without you, there would be no need for wiki ;)
Is there anything else you would wish to help with? I am still learning Python, so more complex stuff maybe next time. I am now working on changing icons of the entities, so they are not all the same.

Question is if you want to use RTD or GH pages, RTD is preferred <https://readthedocs.org/accounts/signup/>
If you setup RTD and you see stable branch to fail, don't worry, as it's just because the latest repo tag does not contain the files yet.

wills106:
I have created the RTD page, I'm just making a few changes as the formatting is off on some of the lists. Then the main readme will point to RTD instead of the Wiki.

I'm also adding in some placeholders so people can see how to contribute to missing information on other Inverters.

PR #611, opened 2023-11-12
Revamped icons in plugin_solax.py

cewbdex:


cewbdex:
I thought about entity `Phase Power Balance` either mdi:scale-balance or mdi:scale-unbalanced, any thoughts? The entity name is different from what inverter calls it (`Phase Unbalanced`), in my head both names mean the opposite, what do you think?

wills106:
> I thought about entity `Phase Power Balance` either mdi:scale-balance or mdi:scale-unbalanced, any thoughts?

I only have a single phase Inverter. So it's not something I can visualise or verify.
Pick what you think suits best.

PR #614, opened 2023-11-12
Added SofarSolar FAQ

cschlipf:
Wrote a FAQ for SofarSolar


cschlipf:
Please do not merge yet

wills106:
Just let me know when you are happy with it.

wills106:
Is it worth adding in the examples in #612 into the FAQ or even sofar-automations.md ?
Assuming we can keep the formatting?

cschlipf:
> Is it worth adding in the examples in #612 into the FAQ or even sofar-automations.md ? Assuming we can keep the formatting?

Already did that before your comment :). Had the same idea

cewbdex:
I stick to a guideline that FAQ has only two level headers - questions (Just like I did in SolaX). Your page looks to me more like an article than FAQ (complex header structure). Your way there are only two items in the side menu and they need to be clicked to display their content and so it's not visible at first sight.

"How to connect xyz?" should have a separate page imho, as it's a standard thing and not something that might happen.

Just my opinion. It's up to Will if he sets any common criteria.

cschlipf:
Good points. Let me know the decision and I will change it accordingly.

wills106:
I think for now the main thing is to get the documentation.

We can look at formatting and where certain items best fit later.
I think sections like modbus-adaptor-setup.md needs a complete rewrite, it's aimed at SolaX with random bit's of Sofar and Solis thrown in there.

cewbdex:
Yes, it's nice to see that after my work people join. I didn't make any major changes to existing pages, just made fixups and added stubs.

cschlipf:
OK, I have removed one level as @cewbdex suggested. Ready for review/merge now. Thank you very much for your inputs.

cewbdex (approved):
Format looks user friendly to me.

cschlipf:
>  I think sections like modbus-adaptor-setup.md needs a complete rewrite, it's aimed at SolaX with random bit's of Sofar and Solis thrown in there.

Agree. Maybe moving Solax specific information into the Solax section? Also with regards to Compatible adaptors: Not every adaptor works or is tested with every inverter. But that is really a bigger revamp of the doc, where I am not feeling comfortable enough yet.

cewbdex:
I would explicitly state which inverter does and does not each adaptor support. The same for setup/install instruction, I would add a substep for every step that needs something extra for specific inverter. Having specific page for everything would be a huge amount of work without any major benefit.

cschlipf:
> I would explicitly state which inverter does and does not each adaptor support. The same for setup/install instruction, I would add a substep for every step that needs something extra for specific inverter. Having specific page for everything would be a huge amount of work without any major benefit.

Depends on how much duplicate information we have for each inverter. Maybe a general section for the generic parts? Right now however the generic part on the top is mainly focused on Solax and does not work for other inverters.

wills106:
> But that is really a bigger revamp of the doc, where I am not feeling comfortable enough yet.

Don't worry I wasn't asking you to rewrite it.

> Not every adaptor works or is tested with every inverter.

Yes that's something I need to workout based on Feedback. There is even a Modbus adaptor listed as unsupported that actually works now thanks to #413 but I haven't got round to updating the "Supported Adaptors list"

> Yes, it's nice to see that after my work people join.

I appreciate the work you have done. I have been in contact with @cschlipf for a while about adding more specifics for Sofar, but with the limitations of the Wiki I wasn't sure the best way to get contributions into the Wiki.

cewbdex:
I think tested adaptors page is a bit too much and also modbus setup, it has a lot of common stuff, so one page would be the best. What differs is entities, as every vendor has a different set.

I don't believe that there would be much IFTTT, as differences are exception rather than rule. Look at Waveshare B, only like one section has subsection.

cschlipf:
> I think tested adaptors page is a bit too much and also modbus setup, it has a lot of common stuff, so one page would be the best. What differs is entities, as every vendor has a different set.

OK, depends on what you see as ModBus setup. What I wrote in the FAQ is about the wiring and that is totally specific, right?

cschlipf:
I any way: Please suggest a structure and I can start working on that. I agree that the current structure is a bit chaotic for non-Solax inverters as you constantly need to figure out, what does not apply to my inverter.

wills106:
> What I wrote in the FAQ is about the wiring

That's probably something that does need to go into "Sofar Installation"

Like I say I'm not too bothered about structure at the moment. Once I have enough content for each Inverter I can understand the differences and how they work and rearrange accordingly.

cewbdex:
> > I think tested adaptors page is a bit too much and also modbus setup, it has a lot of common stuff, so one page would be the best. What differs is entities, as every vendor has a different set.
> 
> OK, depends on what you see as ModBus setup. What I wrote in the FAQ is about the wiring and that is totally specific, right?

Yes, looks fine to me. Maybe if we have a dedicated page for wiring connection at one day, one answer could get merged there.

wills106:
> That's probably something that does need to go into "Sofar Installation"

Or a page dedicated to wiring like @cewbdex mentions:

> a dedicated page for wiring connection

cewbdex:
> In any way: Please suggest a structure and I can start working on that. I agree that the current structure is a bit chaotic for non-Solax inverters as you constantly need to figure out, what does not apply to my inverter.

Test every common page and if anything differs, publish it and if not, you can just state it works for xyz too. If you have any ideas, I am open, I am not an authority here. If you propose anything, I will tell you my opinion and it's up to Will to decide.

cschlipf:
> That's probably something that does need to go into "Sofar Installation"

See #616

PR #615, opened 2023-11-12
Remove duplicate stuff from README.md

cewbdex:
Feels a bit lonely now... All of this should be in the wiki now. If that would help, section could be kept with a link to wiki page instead of its contents.

P.S. not enough PRs already, sorry for busy Sunday ^_^


wills106:
Closing down as I have amended the Reame now.

PR #625, opened 2023-11-13
Match externalgen entities

cewbdex:
I don't have any status for this entity and I found there is no matching sensor entity:
obrazek
Looked in the sheet and this looks like a match, please have a look and correct me if i am wrong.


cewbdex:
There is also:
obrazek
I don't have a state for - the sensor entity has DCB allowedtype, but the number does not.

wills106:
I was trying to move all of this Generator stuff to start with the same naming convention.

I intended for them all to start Generator xyz

It's probably the number & select that need's updating and the `config_maxexport_entity` updating.
I probably need to review the DCB tags as well so it's harmonised.

cewbdex:
Do you think my changes will solve the problem correctly? I could look into the naming convention, update and send for review. Sadly that will probably break the API, but I guess this feature has no to a few users, so the change will have more positive than negative effect.

cewbdex:
I don't have any Modbus document regarding external generator, so I can't check it, but the entities I matched were in my document. Now probably ready for review and merge.

cewbdex:
I also believe that `generator_control` and `generator_max_charge` should have the `DCB` tag, so they are not normally exposed, as I don't believe they have any function without DCB. I can include it in my changes if you wish.

wills106:
> I also believe that `generator_control` and `generator_max_charge` should have the `DCB` tag

Yes I agree

cewbdex:
Done, ready for review.

PR #677, opened 2023-12-22
Update plugin_solis.py

dorianwoolger:
Correction for serial starting 160F is a 5Kw not 3.6 as per image from my inverter.

image

### DIFF
--- a/custom_components/solax_modbus/plugin_solis.py
+++ b/custom_components/solax_modbus/plugin_solis.py
@@ -170,7 +170,7 @@ MAX_CURRENTS = [
     ('0602',  62.5 ), # 3kW 48v
     ('0102',  62.5 ), # 3kW 48v AC Only?
     ('110F',  62.5 ), # 3.6kW 48v
-    ('160F',  62.5 ), # 3.6kW 48v
+    ('160F',  100 ), # 5kW 48v
     ('1031',  100 ), # 5kW 48v
     ('6031',  100 ), # 6kW 48v
     ('110C',  25 ), # 10kW HV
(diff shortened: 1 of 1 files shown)


wills106:
We possibly also need the next two digits. So the first 6 digits as there is a 3.6kW starting the same https://github.com/wills106/homeassistant-solax-modbus/discussions/188#discussioncomment-4619175 so would need to be able to tell the two apart.

wills106:
@jone9618 on your 3.6kW Inverter are the next two digits after '160F' 32 or something different?

dorianwoolger:
> We possibly also need the next two digits. So the first 6 digits as there is a 3.6kW starting the same [#188 (comment)](https://github.com/wills106/homeassistant-solax-modbus/discussions/188#discussioncomment-4619175) so would need to be able to tell the two apart.

Interesting, I did try to google the info for identifying by serial but you don't get a lot. The bit of info that I did get seems to infer that the first 3 digits are the brand label or region then the next 2 are the model.

In my case
160 = Solis
F3 = Model

Is it worth contacting Solis to ask the question, I'm happy to do that if that road has not been explored already.

jone9618:
HelloThanks again for all of the work on the integration.First 8 digits of the serial number are 160F5221Let me know if you need anything else.Simon
    ***@***.***
  

 From: Dorian Woolger ***@***.***>Sent: Friday, December 22, 2023 23:32To: wills106/homeassistant-solax-modbus ***@***.***>Cc: jone9618 ***@***.***>; Mention ***@***.***>Subject: Re: [wills106/homeassistant-solax-modbus] Update plugin_solis.py (PR #677) 

We possibly also need the next two digits. So the first 6 digits as there is a 3.6kW starting the same #188 (comment) so would need to be able to tell the two apart.

Interesting, I did try to google the info for identifying by serial but you don't get a lot. The bit of info that I did get seems to infer that the first 3 digits are the brand label or region then the next 2 are the model.
In my case
160 = Solis
F3 = Model
Is it worth contacting Solis to ask the question, I'm happy to do that if that road has not been explored already.

wills106:
I'm going to close this PR down. I have addressed it myself.
I have '160F3' for the 5kW and 1'60F5' for the 3.6kW

Ther serial's don't much sense though. I don't know if the register at 33000 would help at all?
I have released [2024.01.1](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.01.1)

PR #685, opened 2024-01-02
Add X3 Hybrid G4 FIT variant

beezly:
Adds support for an AC only variant of the X3 Hybrid G4.


wills106:
Does this Inverter need:
https://github.com/wills106/homeassistant-solax-modbus/blob/3a055669c9dabceaa2a9f537f3f405611a0e2383/custom_components/solax_modbus/plugin_solax.py#L262
https://github.com/wills106/homeassistant-solax-modbus/blob/3a055669c9dabceaa2a9f537f3f405611a0e2383/custom_components/solax_modbus/plugin_solax.py#L291
Setting, or is the default sufficient?

wills106:
Found the datasheet for this. Looks like it should be 30A, so I'll add that in before releasing it.

PR #726, opened 2024-02-07
Add Sofar SA3 series inverter support

kkubicki:
SA3 series are part of SOFAR 1.1K...3.3KTL-G3 product lineup.


kkubicki:
Proof:
obraz

PR #728, opened 2024-02-08
Update plugin_sofar_old

kkubicki:
- Remove special characters from serial number
- Update entities parameters

These changes I had to make to have full support from HA energy dashboard with my Sofar 2200TL-G3.
I've noticed that the inverter reports serial number with leading two special characters (same for SB1 and SM1E series I suppose):
<00><0C><53><41><33><45><53><32><32><32><4E><33>
I mitigate this issue by stripping seriesnumber from special characters.

@wills106 If you feel this is too extensive change, I can extract it to separate plugin.

### DIFF
--- a/custom_components/solax_modbus/plugin_sofar_old.py
+++ b/custom_components/solax_modbus/plugin_sofar_old.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from dataclasses import dataclass
 from homeassistant.components.number import NumberEntityDescription
 from homeassistant.components.select import SelectEntityDescription
@@ -50,6 +51,9 @@ ALLDEFAULT = 0 # should be equivalent to HYBRID | AC | GEN2 | GEN3 | GEN4 | X1 |
 
 # ====================== find inverter type and details ===========================================
 
+def remove_special_chars(input_str):
+    return re.sub(r'[^A-z0-9 -]', '', input_str)
+
 def _read_serialnr(hub, address, swapbytes):
     res = None
     try:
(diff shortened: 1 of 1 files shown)


wills106:
`seriesnumber.count` was only intended as a quick fix for https://github.com/wills106/homeassistant-solax-modbus/discussions/522#discussioncomment-7165117

I didn't realise the energy dashboard wasn't fully working on sofar_old as no one has reported it.

kkubicki:
Yeah, it is probably Product code that is always sent prior serial number, I tried to get it from other address (0x2003, 0x2004) but the result is the same.

kkubicki:
@wills106 I have seconds thoughts regarding ActivePower change from KILO_WATT_HOUR to KILO_WATT (this change applies only to X3 inverters, for X1 ActivPower was already KILO_WATT). I do not know what is the correct unit for this parameter, in my case (X1 inverter) it seems to be actual power sent out so kW makes more sense. 
There seem to be some confusion regarding this, some plugins use kWh, other ones kW...

wills106:
kW tends to be used for Grid connected like import / export.

kWh used for transfer to and from battery.

I haven't pushed it out as a release yet, so you can relook at it and see what fits best.

wills106:
On my SolaX some realtime:

image

Daily totals / historic:
image

image

Does the Sofar LCD give some pointers?

kkubicki:
It is a very simple inverter, LCD has only a couple of lines. Here's excerpt from manual: 

Standard display
This shows a rolling display of the DC voltage/current (PV1), energy
yields (today/total), grid voltage and current, as well as the status.

I can push another commit that restores previous settings for X3 inverters, and leaves X1 changes only. That would be safest thing to do, I do not want you to have extra work because of my changes. On the other hand, unit of Power is kW, not kWh after all.

PR #730, opened 2024-02-08
plugin_sofar_old: restore X3 settings

kkubicki:
This restores some settings for X3 inverters to original form.


kkubicki:
Since no one complains on 3 phase inverters, here's reversing commit.

PR #739, opened 2024-02-13
Fix select device issue

isorin:
Looks like select wasn't updated and all "select" entities ended up in a different device in 2024.02.3


wills106:
Yes I missed uploading the file 🤦‍♂️

PR #745, opened 2024-02-15
Allow to set desired grid power along with minimum and maximum battery power.

cschlipf:
### PREVIOUS BEHAVIOR
When setting battery power, desired grid was always set to 0 and battery min and max charge to the value of passive_mode_battery_power. This does not allow the full flexibility of the passive mode

### CHANGE
With this addition users can set desired grid power, minimum charge/discharge and maximum charge/discharge:
image

When setting these values the registers are set accordingly:
image

Now this allows finer control like this:

- Passive Mode Grid Power: 0
- Passive Mode Max Battery Power: 2500
- Passive Mode Min Battery Power: -5000

Try to keep grid power at zero like in self use, but limit charging of the battery to 2.5kW, while still allowing discharging to 5 kW

- Passive Mode Grid Power: 2000
- Passive Mode Max Battery Power: 5000
- Passive Mode Min Battery Power: 0

Force charging the battery with 2 kW while preventing discharging. If sun delivers more allow to charge with up to 5kW.

### OUTDATED

I am trying to modify plugin_sofar.py to set upper and lower limits for battery charge/discharge in passive mode. So far it is not working yet. Values do not get committed. Not sure what I am doing wrong here as a total novice.

Registers I am trying to write are the green registers here:
image

UI is showing the new controls already:
image

### DIFF
--- a/custom_components/solax_modbus/plugin_sofar.py
+++ b/custom_components/solax_modbus/plugin_sofar.py
@@ -104,7 +104,7 @@ class SofarModbusSensorEntityDescription(BaseModbusSensorEntityDescription):
 # ====================================== Computed value functions  =================================================
 
 def value_function_passivemode(initval, descr, datadict):
-    return [ (REGISTER_S32, datadict.get('passive_mode_battery_power', 0)),
+    return [ (REGISTER_S32, datadict.get('passive_mode_grid_power', 0)),
             (REGISTER_S32, datadict.get('passive_mode_battery_power_min', 0)), 
             (REGISTER_S32, datadict.get('passive_mode_battery_power_max', 0)),
            ]
@@ -200,8 +200,8 @@ NUMBER_TYPES = [
     #
     ###
     SofarModbusNumberEntityDescription(
-        name = "Passive Mode Battery Power",
-        key = "passive_mode_battery_power",
+        name = "Passive Mode Grid Power",
+        key = "passive_mode_grid_power",
         native_unit_of_measurement = UnitOfPower.WATT,
         device_class = NumberDeviceClass.POWER,
         unit = REGISTER_S32,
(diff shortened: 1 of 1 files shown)


cschlipf:
One possible reason: Registers needs to be written together in one batch with 1187. One button to commit all values together would be better anyway. I assume I did not understand the value_function_passivemode correctly. Maybe this is doing this already if I pass passive_mode_battery_power as first, passive_mode_battery_power_min as second and passive_mode_battery_power_max as third parameter like this?

```python
def value_function_passivemode_min(initval, descr, datadict):
    return [ (REGISTER_S32, datadict.get('passive_mode_battery_power', 0))
            (REGISTER_S32, datadict.get('passive_mode_battery_power_min', 0)), 
            (REGISTER_S32, datadict.get('passive_mode_battery_power_max', 0)),
           ]
```

sahomm:
sounds good, big thanks for your inspiriance and go to far

wills106:
Will probably need to update https://homeassistant-solax-modbus.readthedocs.io/en/latest/sofar-operation-modes/ to explain how to use this mode.

I forgot there is still "ro_" entities in plugin_sofar.py

I'll need to couple them back up as per Solis mentioned
https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2023.05.2
https://homeassistant-solax-modbus.readthedocs.io/en/latest/solis-operation-modes/#timed-charge-discharge

cschlipf:
These ro_ entities are great to check whether the values have been written or check the actual current state in an automation.

Yes, Documentation needs to be updated. I plan to do that, but before I am doing this I would like to add two more entities for controling the passive mode timeout (also very important for reliable passive mode) and I would also like to suggest some renamings. Will post a new discussion on this soon.

I believe then we rather need a more complete documentation on all the entities of Sofar as there are so many specialities to consider here.

wills106:
The entities will still be there instead of "ro_passive_mode_lower" they will just be "passive_mode_lower"
It allows the Inverter to feedback to the Integration if you make a change on the Inverter or Cloud app.

cschlipf:
Thank you. Any other changes needed for this PR?

PR #751, opened 2024-02-17
Documentation: Added Sofar modes of operations and RS485 termination

cschlipf:
Added documentation about Sofar Modes of operations also as follow up to #748.
Added resistor termination to RS485 installation.


wills106:
Do you actually need termination resistors on the Sofar Inverters?
I know on the SolaX people don't bother.

sahomm:
review looks good, @cschlipf grate job, thx!

cschlipf:
> Do you actually need termination resistors on the Sofar Inverters? I know on the SolaX people don't bother.

Yes, they are required. Just learned a few days ago from @sahomm that these were the cause for my unreliable communication over RS485. After all this was also the reason why I switched to the LSE-3 in the end.

sahomm:
That's how it is, but in my opinion the LSE-3 is the first option if you have the choice!!!

Issue #437, opened 2023-06-10
[BUG] maximum Export Control User Limit too low with other inverters active


blb4github:
**Describe the bug**
The Maximum Export Control User Limit for my Gen2 X1 SK-SU (U30) inverter is set to 3000. the PV system connected to this inverter is 2250 Wp but I have more system with total Wp of 6022 Wp. This limit of 3000 is limiting my output to Grid.

I have changed  the MAX_EXPORT value in plugin_solax.py for my inverter ('U30') from 3000 to 6000 and this seems to do the trick. 
MAX_EXPORT = [
    ('L30E',    3000 ), # Gen2 X1 SK-TL
    ('U30',     6000 ), # Gen2 X1 SK-SU

**Mandatory details**
1. Integration Version: 2023.06.7 (latest)
2. HA Version: 2023.4.0
3. Inverter Brand: Solax
4. Plugin Used: modbus
5. First 6 digits of Serial: U30EDB 
6. Connection Method: Ethernet

**Additional context**
It would be great if this limit can be upgraded to higher values, taking into account other inverters exporting to GRID as well.
it's not a problem for the interver; I can change this value to 30.000 via the Solax (local connected) app.

infradom:
@wills106 Maybe we could add a config_entity like we did for export_control_user_limit_readscale
This would make sense for some non-standard setup's
This entity could be disabled by default. Any opinion on that ?

wills106:
So is it treating this value as a global combined export then? Sot it's a sum of both Inverters output. As you are exceeding 3000w it's starting the throttle the output of the SKSU?
As this Inverter should only be able to export 3000w on it's own. That's where this value came from.

blb4github:
It's measuring the value from the modbus kWh meter connected to the GRID wires. The inverter is indeed throttling the output when the kWh meter reports > 3000 W. Today this didn't happen as I did change th.e (max export) value to 6000W

wills106:
I have just released [2023.06.8](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2023.06.8)
There is a new number entity "Config Max Export" which you can enable to override the max limit the sliders will go to.

blb4github:
Great, thanks!! I have upgraded to 2023.06.08, activated and set the Config Max Export entity to a value which works for my setup (6000 W). perfect!!

blb4github:
This issue can be closed.

wills106:
I have updated the wiki for:
[...]

PR #754, opened 2024-02-17
Added support for Passive Mode Timeout

cschlipf:
- Sofar: Add support for setting passive mode timeout along with the time out action
- Sofar: Added RO: prefix back to read only entities for controlling, whether values are set. Prefix helps to differentiate in UI between the temporary values and the read values from the inverter
- Moved more entities to diagnostics, where these entities are Read-Only.
- Sofar: More reasonable value boundaries for desired grid, maximum battery power and minimum battery power
- Doc: Fixed a formating problem in sofar-operations-modes
- Doc: Documentation for new settings


custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -221,8 +234,8 @@ def value_function_toumode(initval, descr, datadict):
         device_class = NumberDeviceClass.POWER,
         unit = REGISTER_S32,
         fmt = "i",
-        native_max_value = 15000,
-        native_min_value = -15000,
+        native_max_value = 10000,
```
sahomm, line 237: is that right? I'm missing the 15 KTL (see next step), we need to observe a range of 5 to 20 KTL

wills106: It is possible to base that on inverter size using something like the serial number. Although we don't inspect enough of the serial number at present.

https://github.com/wills106/homeassistant-solax-modbus/blob/f63bf50a32e35e6d902ffd63f13f3e362a1b1769/custom_components/solax_modbus/plugin_solax.py#L313-L316

https://github.com/wills106/homeassistant-solax-modbus/blob/f63bf50a32e35e6d902ffd63f13f3e362a1b1769/custom_components/solax_modbus/plugin_solax.py#L710-L725

Or we could just use another number entity so maximum is user definable?

cschlipf: > is that right? I'm missing the 15 KTL (see next step), we have to observe a range of 5 to 20 KTL

Yes, this is the maximum charge/discharge of the battery. A fully extended BTS-5K battery with 20 kWh has a maximum of 10 kW charge/discharge.

cschlipf: > It is possible to base that on inverter size using something like the serial number. Although we don't inspect enough of the serial number at present.
> 
> https://github.com/wills106/homeassistant-solax-modbus/blob/f63bf50a32e35e6d902ffd63f13f3e362a1b1769/custom_components/solax_modbus/plugin_solax.py#L313-L316
> 
> https://github.com/wills106/homeassistant-solax-modbus/blob/f63bf50a32e35e6d902ffd63f13f3e362a1b1769/custom_components/solax_modbus/plugin_solax.py#L710-L725
> 
> Or we could just use another number entity so maximum is user definable?
> 
> There may be cases when you want the value to be larger than what the Inverter in theory can handle #437 .

And yes, you are right: More could be required if multiple inverters are chained. But then with the current implementation we have a hard limit at a bit more than 32kW anyway, as we are writing just the lower 16bit register of the int32, correct?

So my thought was to support a single inverter, maxed out spec only.

However note that even if a user goes beyond that these limitations are not critical. Battery charge/discharge is only concerning the batteries that are connected to this single inverter. So it doesn't matter how many inverters you chain together.

Desired grid power just sets the desired buy/sell from/to the Grid. I believe even 20 kW is already a very, very rare case that someone will ever set.

cschlipf: Just saw that one inverter can be maxed out to 40 kWh BTS-5K batteries (2x BTS-E20), which would equal to a max charge/discharge of 20kW. I will increase the values again.

wills106: We can always manually override the limits as needed in the future.

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -2889,36 +2933,44 @@ def value_function_toumode(initval, descr, datadict):
         allowedtypes = HYBRID,
     ),
     SofarModbusSensorEntityDescription(
-        name = "Passive Mode Timeout",
-        key = "passive_mode_timeout",
+        name = "RO: Passive: Timeout",
+        key = "ro_passive_mode_timeout",
```
wills106, line 2937: As you have added the `prevent_update = True,` to the data only number you don't need `ro_` in front of the key name.
This will allow the Integration to read the correct "passive_mode_timeout" register and update the data only one upon change on the Inverter or on restart of HA.
The default `prevent_update = True,` gives you 120seconds to make all of your changes and press the button before the Integration reverts back to what's stored in the register.

docs/sofar-operation-modes.md
```diff
@@ -58,25 +58,20 @@ _TBD: Describe how to use Timing Mode with the Integration_
...
 
-- **Passive: Maximum Battery Power**: _Number in Watt_. 
+- **Passive: Maximum Battery Power**: _Number in Watt_. Limit the maximum battery power. Must be greater than or equal to "Passive Minimum Battery Power". Positive values indicate charging, negative values indicate discharging. Note that the displayed value in this field does not necessarily reflect the current value that is set in the inverter. See the entity "RO: Passive: Maximum Battery Power" for checking the current value.
 
-  Limit the maximum battery power. Must be greater than or equal to "Passive Minimum Battery Power". Positive values indicate charging, negative values indicate discharging.
+- **Passive: Minimum Battery Power**: _Number in Watt_. Limit the minimum battery power. Must be less than or equal to "Passive Maximum Battery Power". Positive values indicate charging, negative values indicate discharging. Note that the displayed value in this field does not necessarily reflect the current value that is set in the inverter. See the entity "RO: Passive: Minimum Battery Power" for checking the current value.
```
wills106, line 66: The above comment will make this statement about ro_ no longer required. As the data only number and the sensor entity will be coupled together correctly. While still allowing you to change the values.

custom_components/solax_modbus/plugin_sofar.py
wills106: Ignore comment. I thought these were numbers and not selects.

docs/sofar-operation-modes.md
wills106: Ignore comment. I thought these were numbers and not selects.

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -105,10 +105,15 @@ class SofarModbusSensorEntityDescription(BaseModbusSensorEntityDescription):
...
-            (REGISTER_S32, datadict.get('passive_mode_battery_power_max', 0)),
+             (REGISTER_S32, datadict.get('passive_mode_battery_power_min', 0)), 
+             (REGISTER_S32, datadict.get('passive_mode_battery_power_max', 0)),
            ]
 
+def value_function_passive_timeout(initval, descr, datadict):
+    return  [ ('passive_mode_timeout', datadict.get('passive_mode_timeout', 0), ),
+              ('passive_mode_timeout_action', datadict.get('passive_mode_timeout_action', 0), ), 
+            ]
+
```
wills106, line 116: I think this would be better, should return the value again on restart and not just blank.
```
def value_function_refluxcontrol(initval, descr, datadict):
    return  [ ('reflux_control', datadict.get('reflux_control', datadict.get('ro_reflux_control')), ),
              ('reflux_power', datadict.get('reflux_power', datadict.get('ro_reflux_power')), ), 
            ]
```

cschlipf: Thank you. You are right. I have updated that.

PR #769, opened 2024-02-22
added highest and lowest battery temp for Growatt SPH

mani951:


wills106:
You need to re-pull my branch and add in the extra registers.
If I merged this it would revert the async changes.

mani951:
I just forked your repo 5 minutes ago, so there should not be any async changes

mani951:
Great, thanks for adding!

Issue #784, opened 2024-03-02
[Bug]: Sofar - 3.3-12KTLX-G3 PV Power Total no longer correct after firmware update of inverter


darkrain-nl:
### Describe the bug

After updating my inverter to firmware version: V100011-10-14 (sorry don't know the old firmware version) the PV Power Total value is changed from scale 0.01 to 0.1 (like the hybrid models). Another change I noticed is it doesn't show 10-90W values anymore, only 100W and above, so lower then 100W is shown as 0W.
![afbeelding](https://github.com/wills106/homeassistant-solax-modbus/assets/24763370/a5abf76b-85cd-4816-8802-006ea7ab7ad8)

The scale I can fix locally, and I could do a pull request for it, but that would probably be a breaking change on inverters that are still on the older firmware... Or is there a way to have it be depending on firmware?

What are the options here, is there something that can be done to show the 10-90W values as well, can the scale be adjusted just for this firmware version?

### Integration Version

2024.02.6

### Homeassistant core version

2024.2.5

### Inverter brand

Sofar

### Plugin used

plugin_sofar.py

### Serial prefix

SS2ES1

### Connection Method

Waveshare adapter

wills106:
I have seen your PR.

I think we either need to detect Firmware version through the Integration. Or have the option to manually override the scaling.

darkrain-nl:
That makes sense, I do not have time to work on this anytime soon... but feel free to close the current PR.

Can come back to it later, or somebody else might have a go...

wills106:
I'll leave it open as a reminder.

We have done similar overrides on SolaX. So I can't see it being a  big issue.

d4ystek:
I have the same problem only with all entities (data is not correct). Did you manage to solve the problem? Software version 370

PR #785, opened 2024-03-02
Change scale for PV Power Total for PV inverters with latest firmware potentially fixes wills106/homeassistant-solax-modbus#784

darkrain-nl:
Please only  merge if we are certain it won't create issues for others on older firmware...


PatrikTrestik:
@darkrain-nl if it is difference between firmware versions you have to add more work for it to be compatible.

1) in plugin add new flax which identifies new FW. Use completely new flag group if it does not fit to existing one.
2) in async_determineInverterType detect new firmware (hope it is possible from known values) and assign flags.
3) in matchInverterWithMask add your new flag
4) duplicate sensor and use flags so only one of two is instantiated depending on firmware.

wills106:
In #784 we mentioned we needed to look into detecting Firmware version.
But not got any further with it yet

darkrain-nl:
Oops I closed this by accident...

Issue #786, opened 2024-03-02
[Bug]: Solis - RHI-3P10K-HVES-5G: Setting Engery Control Switch Setting not working


kai-seipp:
### Describe the bug

Hi Guys, 

**Motivation / goal:** 
Since I now have an electric car I am very much interested in being able to alter some battery storage settings to avoid draining my home battery when chargeing my car. 
Therefore I now take some interest in these settings. 

**Issue description:** 
_energy_storage_control_switch: 
selecting different setting will not work as expected: 

- OK: Choosing "Selfuse - No Grid Chargeing" is working as expected. 
- Bug: Choosing "Timed Charge/Discharge - No Grid Charging" - is triggering "Timed Charge/Discharge - No Grid Charging" - But is actually allowing grid charging
- Bug: Choosing "Backup/Reserve - No Grid Charging" switches the inverter back to "Selfuse - No Grid Charging" 
- Bug / redundant: Choosing "Selfuse" switches the inverter to "Selfuse - No Grid Charging" 
- Bug / redundant: Choosing "Timed Charge/Discharge" - is triggering "Timed Charge/Discharge - No Grid Charging"
- unclear: "Battery Awaken" will go back to Selfuse - no Grid Charging. I don't know what the expected behavior would be
- Bug: "Choosing Backup/Reserve - No Timed Charge/Discharge" results in "Selfuse - No Grid Charging"
- Bug: Choosing "Backup/Reserve" as value results the Inverter to be set to "Timed Charge/Discharge - No Grid Charging" 

I can't test "Off-Grid Mode" as I don't have my house of grid

**Expected:** 
It would be super helpful being able to switch to Backup/Reserve Mode to prevent battery drainage. 
I would also suggest to strip down the possible options to those we can working. 

**Enviroment**
I am Running on 2024.01.8 again, as 2024.2.x kept messing up my statistics (will have to address within an other issue). 
Running on current HA version 

### Integration Version

2024.1.8

### Homeassistant core version

2024.2.5

### Inverter brand

Solis

### Plugin used

plugin_solis.py

### Serial prefix

110CA2

### Connection Method

Waveshare Adapter

### Detailed Error Log

There are no errors within the log

i-jk:
Yes, I'm not sure where all these new values for Energy Control Switch come from? They seemed broadly correct before the recent updates but now there's loads and they don't align with the inverter's actual behaviour. 

[...]

Issue #724, opened 2024-02-06
[Bug]: Data from some sensors resetting at 2358 and a few seconds to zero e.g. sensor.solax_today_s_solar_energy


UK-JustinSmall:
### Describe the bug

I seem to have a issue with some data resetting values to zero at a fraction before 00:00:00 hours resulting in last value being reported for the day as zero. e.g. for daily solar generation sensor.solax_today_s_solar_energy I get zero values in my chart all throughout January.

![solax 1 hour error](https://github.com/wills106/homeassistant-solax-modbus/assets/129121587/d96a1654-8977-46a9-9fc3-f21d361da629)
![2024-02-07_00-45-20](https://github.com/wills106/homeassistant-solax-modbus/assets/129121587/8289f8b8-9116-45d1-8bbd-c2c4e1449fa9)
![Snag_1bd701a](https://github.com/wills106/homeassistant-solax-modbus/assets/129121587/a74925ea-ddff-4113-87f9-071d88e651dc)

Date is correct timezone on ubuntu server and on the inverter

### Integration Version

2023.12.1

### Homeassistant core version

20231208.2

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H4752A

### Inverter firmware versions

3.001.02

### Connection Method

Waveshare Adapter

### Dongle firmware

12.01

UK-JustinSmall:
Appears some entities are updating at a 23:58 and a few seconds each night and resetting the values at that time to zero. So then the history is then showing the value at 00:00:00 any chance this could be set so that the time these entities update are a few seconds after midnight so it captures the value for end of day in my apex charts.

type: picture-elements
image: /local/solar/card-medium-03.png
elements:
  - type: custom:tabbed-card
    style:
      top: 25%
      left: 50%
      width: 97%
      height: 50%
    tabs:
      - attributes:
          label: Daily
        card:
          type: custom:apexcharts-card
          experimental:
            brush: true
          brush:
            selection_span: 30d
          graph_span: 180d
          span:
            end: day
          header:
            show: true
            title: ''
            floating: false
            show_states: false
            colorize_states: true
            standard_format: true
          apex_config:
            chart:
              height: 380px
          all_series_config:
            type: column
            stroke_width: 1
            show:
              legend_value: false
            statistics:
              type: state
              period: day
              align: end
          yaxis:
            - min: 0
              max: 60
              decimals: 0
          series:
[...]

PR #795, opened 2024-03-08
A few minor Solis improvements, add RTC sync offset

isorin:
A few minor improvements for Solis:
- A bit more consistency in the "Self-use" mode names
- Fixed "Timed Discharge End Minutes" not being disabled by default (like all the other Timed charge/discharge sensor entities)
- Fixed Battery Charge/Discharge Current rounding
- Added the ability to specify a "sync offset" when syncing the RTC clock, as the RTC have a tendency to either gain or lose time it makes the RTC syncs less frequent (for instance, sync the time with a "-10" seconds offset and let the RTC gain time until it gets to "+10" seconds, then sync again)


wills106:
Thanks, the offset might be useful for other Inverters as well such as #724

danielknox:
Looking at the commit it looks like modes such as 'Timed Charge/Discharge' got removed. For me this broke a number of automations referencing it which were controlling timed charge and discharges based on energy pricing. Its not necessarily clear what these have been replaced with - is Battery Awaken + Timed Charge/Discharge" the replacement? Can the documentation listing Solis modes (currently listed as a WIP) be updated with this?

isorin:
Hi, sorry about the break. I know as I also use names in my automations and if any of these names would change I would have to update my automations to reflect the change.

There are two main modes of operation in Solis: "Self-use" and "Feed-in priority".
The change aimed to make the naming a bit more consistent across these two modes, so now we have the following two sets named in a similar way:

"Self-Use"
"Self-Use - No Timed Charge/Discharge"
"Self-Use - No Grid Charging"

"Feed-in priority"
"Feed-in priority - No Timed Charge/Discharge"
"Feed-in priority - No Grid Charging"

If you are looking for the old 'Timed Charge/Discharge' this is now called "Self-Use"

"Self-Use" and "Feed-in priority" modes have both 'Timed Charge/Discharge' and 'Grid Charging' activated.
If you are looking for a mode without "Timed Charge/Discharge" then use the one which says "- No Timed Charge/Discharge"
if you are looking for a mode without both then use the "- No Grid Charging" mode.

I hope this helps! Let me know if my explanation is clear.

isorin:
Added info for Energy Storage Control Switch names:
https://github.com/wills106/homeassistant-solax-modbus/pull/823

danielknox:
> Hi, sorry about the break. I know as I also use names in my automations and if any of these names would change I would have to update my automations to reflect the change.
> 
> There are two main modes of operation in Solis: "Self-use" and "Feed-in priority". The change aimed to make the naming a bit more consistent across these two modes, so now we have the following two sets named in a similar way:
> 
> "Self-Use" "Self-Use - No Timed Charge/Discharge" "Self-Use - No Grid Charging"
> 
> "Feed-in priority" "Feed-in priority - No Timed Charge/Discharge" "Feed-in priority - No Grid Charging"
> 
> If you are looking for the old 'Timed Charge/Discharge' this is now called "Self-Use"
> 
> "Self-Use" and "Feed-in priority" modes have both 'Timed Charge/Discharge' and 'Grid Charging' activated. If you are looking for a mode without "Timed Charge/Discharge" then use the one which says "- No Timed Charge/Discharge" if you are looking for a mode without both then use the "- No Grid Charging" mode.
> 
> I hope this helps! Let me know if my explanation is clear.
> 
> I will look into updating the documentation as well.

No problem, explanation makes sense and automations all updated. Timed Charge/Discharge - No Grid Charging is still listed in there as an available option which I originally selected as a fix (which ironically begins a time charge/discharge pulling from the grid, but I think  #786 mentions that issue).

isorin:
> No problem, explanation makes sense and automations all updated. Timed Charge/Discharge - No Grid Charging is still listed in there as an available option which I originally selected as a fix (which ironically begins a time charge/discharge pulling from the grid, but I think #786 mentions that issue).

Now that you mention it, I think that particular option was missed when doing the renaming, it should probably be called:
"Self-Use + Timed Charge/Discharge - No Grid Charging"

Maybe we should consider refactoring this a bit in the future and use a set of individual boolean flags instead (e.g. "Grid charging", "Time of use", etc.) which can be set or reset before calling update on the "Energy Storage Control Switch".

Iinfo for the "Energy Storage Control Switch" options:
https://github.com/wills106/homeassistant-solax-modbus/pull/823

PR #797, opened 2024-03-10
Sofar: Added support for 8 batteries

cschlipf:
Especially users with Amass GTX-3000 batteries, where every module is reported as a single battery, needs support for more than 4 batteries.

With this PR the integration will support 8 batteries. It's possible to support up to 16 batteries, but I hope that for now 8, which will support 20 kWh of capacity, will be enough for most users.


custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -2191,7 +2191,6 @@ def value_function_sync_rtc_ymd_sofar(initval, descr, datadict):
         native_unit_of_measurement = UnitOfElectricCurrent.AMPERE,
         device_class = SensorDeviceClass.CURRENT,
         register = 0x605,
-        newblock = True,
```
cschlipf, line 2194: As the previous register already has set 'newblock' to true, it shouldn't be necessary on the next register. Instead I have added 'newblock' to the first parameters of each battery.

cschlipf:
@wills106 - It looks like with these changes the battery display on the top right is gone:
image

Can you help me to find out what I broke here and what I need to do to fix that?

wills106:
It's possibly because there are now multiple `device_class = SensorDeviceClass.BATTERY,` try commenting them out on battery 2-8 or is there a total battery capacity you can assign as `device_class = SensorDeviceClass.BATTERY,`

cschlipf:
> It's possibly because there are now multiple `device_class = SensorDeviceClass.BATTERY,` try commenting them out on battery 2-8 or is there a total battery capacity you can assign as `device_class = SensorDeviceClass.BATTERY,`

Thanks will try that out. Yes, there is a total battery capacity.

Issue #686, opened 2024-01-03
[FR] Add support for Growatt SPF


wills106:
Have same problem: ERROR (MainThread) [custom_components.solax_modbus.plugin_growatt] unrecognized growatt inverter type - firmware version : 067.01068.01 , with Growatt spf 5000es inverter , spf inverter is not listed, how to setup?

_Originally posted by @Asterix229 in https://github.com/wills106/homeassistant-solax-modbus/issues/673#issuecomment-1863296639_

              Growatt SPF 5000 ES conected to local, usb-usb or rs485 modbus converter, work on pc program pvkeeper, on solaxpower recive some data if inverter put srne

_Originally posted by @Asterix229 in https://github.com/wills106/homeassistant-solax-modbus/issues/673#issuecomment-1874429399_

Growatt SPF is an Inverter aimed at off grid use, while having the capability to charge off the Grid or Generator etc
Uses a different Modbus register set as outlined in "OffGrid-Modbus-RS485RS232-RTU-Protocol-V0.14-20210420.pdf"

Asterix229:
My inverter hibrid Growatt SPF 5000 ES 1x solar PV, main, load, battery, software and hardware 067 01 1.04 and 068 01 1.04, DTC device type code 034xx OffGrid SPF 3-5K, where to change register? Try plugin_growatt.py change HIBRID number but no work, think Modbus protocol v0.14 is my inverter protocol. Where to change? Firmware version recognize.
Error log:

2023-12-17 21:51:02.439 WARNING (MainThread) [homeassistant.util.async_] Detected blocking call to sleep inside the event loop by custom integration 'solax_modbus' at custom_components/solax_modbus/__init__.py, line 344: return self._client.read_holding_registers(address, count, **kwargs), please create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues
2023-12-17 21:51:02.504 WARNING (MainThread) [homeassistant.util.async_] Detected blocking call to sleep inside the event loop by custom integration 'solax_modbus' at custom_components/solax_modbus/__init__.py, line 344: return self._client.read_holding_registers(address, count, **kwargs), please create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues
2023-12-17 21:51:02.556 ERROR (MainThread) [custom_components.solax_modbus.plugin_growatt] unrecognized growatt inverter type - firmware version : 067.01068.01

wills106:
[...]

PR #806, opened 2024-03-15
Add ability to change battery source and charge source modes, Extra informational registers, support for 500.* 5kW inverters and correct register 27

PrimusNZ:
This PR adds in two changeable options
- **State Power** (Where the inverter is mainly powering the loads from
- **State Charge** (Where the inverter is prioritising charging from)

Fixes Registers:
- **27: Inverter Load** - This is inverter load and is expressed as a percentage, not Battery SOC (Which is on register 18)
 
Also adds registers for:the following: (All disabled by default)
- **Max Charge Current** - How much current the inverter can draw to charge the battery
- **Max AC Charge Current** - How much current the inverter can draw from shore power
- **Battery Type** - Information to show what battery type the user has set in the inverter
- **Fault Code** - Displays the current fault code in the inverter as a number - May update this to human readable in future
- **Warning Code** - Displays the current warning code in the inverter as a number - May update this to human readable in future
- **Constant Power OK** - Unsure of this one, I think it displays whether shore power is considered stable or not
- **Battery Overcharge** - Shows if the battery is or has been overcharged.

Adds support for SPF inverters with firmware **500.10**

Work for Feature Request: #686


PrimusNZ:
Also fixed scaling on Grid Frequency and Output Frequency.
Added scale and rounding to Battery Power Charge.

Issue #805, opened 2024-03-14
[Bug]: Solax EV Charger not working values


Philipp-X1:
### Describe the bug

As mentioned here https://github.com/wills106/homeassistant-solax-modbus/pull/772 I want to report not working values related to the EV Charger

### Integration Version

2024.03.01

### Homeassistant core version

2024.3.0

### Inverter brand

Solax X3-EVC-11K

### Plugin used

plugin_solax_ev_charger.py

### Serial prefix

C31103

### Inverter firmware versions

N/A

### Connection Method

Waveshare RS485 to ETH

### Dongle firmware

N/A

### Detailed Error Log

The Entity name = "Charge Added Total" shows 111.457.075,2 kWh at my charger but should be 1.700,7 kWh

Most other entities seem to be ok. 
I think we wont see any Grid related values ever because there cant be any data source for them ( No CT Clamps and no connection to Inverter...)?!

Furthermore i can only check the values for 1 Phase since my Car only supports 1 Phase charging. But maybe there will be a user in the near future to check them out.

Bigger problem seems to be that I cant restart charging after I put the charger use mode to "stop" while I am charging.
So if I want to go back to Fast Eco or Green nothing is happening.
When I am using the homeassistant-solax-http plugin I can start the charing again. Maybe @PatrikTrestik could take a look at this somewhen :)

Thanks!

PatrikTrestik:
Seems related to REG32 value parsing.
I'll try to fix, but have to find time to "rewire" my garage for it. Possibly next week.

PatrikTrestik:
Solved in https://github.com/wills106/homeassistant-solax-modbus/pull/829

Philipp-X1:
Sorry. My fault mixing this up.

The question that should be posted here:

Could you also take a look at the Start Stop issue? 
If you change the Charger to Stop in HA it is in possible to change to Fast Green or Eco anymore.
Thank you :)

johny-mnemonic:
Starting the charging also doesn't work for me through Modbus.

Also I am not sure what was the root cause, but since I added the EVC as a second device to the modbus integration my Inverter started behaving really weird. It was randomly charging and discharging the battery with high power (like 8-10kW !!!) and then I noticed it also reports really high consumption and even production of the house. It was showing peaks more than 10-12kW of consumption and up to 8kW production, while there is nothing in my house which could produce any electricity.

[...]

PR #829, opened 2024-03-29
Moving "Charge added total" to HOLDIG reg 0x619.

PatrikTrestik:
For unknown reason (bug of SolaX) input register 0x10 is reg 32 BIG_ENDIAN.
Fortunately there is holding register 0x619 with same meaning which is correct.

Better to use correct register than fixing non standard encoding.

Solves #805


Philipp-X1:
Works! Thank you!

PR #869, opened 2024-04-23
make scan_group configurable

gec75:
I saw you accepted my changes... 👍

But I wanted to play myself with different intervals, and I realized that modifying source is not convenient (especially for users); and a full HA restart is also not nice...
Therefore I did some refactoring moving from explicit scan_interval to scan_group, and added 3 configuration options (default, medium, fast).
This way we can assign sensors (descriptors) to groups, but the scan interval is now user configurable (defaults **not** changed)

I also applied some small code refactoring, read config directly in hub constructor instead of passing lots of parameters.

Please review...

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -72,6 +72,7 @@ from .const import (
     DEFAULT_PLUGIN,
     # PLUGIN_PATH,
     SLEEPMODE_LASTAWAKE,
+    SCAN_GROUP_DEFAULT,
 )
 from .const import (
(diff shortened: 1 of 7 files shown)


gec75:
And btw, can you add a mention for my Inverter? Or should I add it to my fork?

wills106:
Sorry I have been busy past couple of weeks. I had a quick look at it locally and was thinking about adding the scan times to the config flow and noticed you have made that change.

gec75:
Now you should slowly try to set/assign sensors in other plugins to groups. Then users could be able to try faster scanning.

Issue #813, opened 2024-03-18
[FR] Add support for SolaX Gen5 X1-IES / X3-IES / X3-Ultra Hybrid Inverters


wills106:
Solax X3 Ultra 25k > `H3BC25...`

_Originally posted by @simowNgithub in https://github.com/wills106/homeassistant-solax-modbus/discussions/520#discussioncomment-8790368_

Add support for SolaX Gen5 X1-IES / X3-IES / X3-Ultra Hybrid Inverters

2024.03.5 adds support for Battery 2 and MPPT 3

**It looks like quite a few of the Selects and Number entities have moved to different registers on the Gen5 Inverters.
So I wouldn't make any changes to those yet.**

_Originally posted by @wills106 in https://github.com/wills106/homeassistant-solax-modbus/discussions/520#discussioncomment-8829590_

wills106:
I have just released [2024.03.6](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.03.6) which corrects the Generator Control as the registers have moved to a different location.

I have removed the Following from the Gen5:
https://github.com/wills106/homeassistant-solax-modbus/blob/176d15b889b29635aa1980ecf5c03690c03ee733/custom_components/solax_modbus/plugin_solax.py#L3259-L3382
I'm not sure if they are present on the Gen5.

simowNgithub:
@wills106 i updated and checked your new implementation, following entities getting wrong values: 

The example with the battery capcatiy is really confusing, because you map the same register to solax_battery_1_capacaty and solax_battery_capacaty. One mapping is working and the other mapping is not working.

<img width="373" alt="image" src="https://github.com/wills106/homeassistant-solax-modbus/assets/117570480/acd39e40-5f1d-418b-ac0f-266b166cda50">

wills106:
It looks like you have a mix of new and old entities there.
Did you manually add your Inverter like https://github.com/wills106/homeassistant-solax-modbus/discussions/520#discussioncomment-8814618
I would remove the Inverter from the Integration page, restart HA and add it back in again.
it should then configure itself correctly. We can then look at which entities need scaling correctly.

simowNgithub:
I always add a second Inverter„SolaxTest“ with the same IP address. I think then everything is on a clean setup, correct?

Krawuzzi:
removed inverter, restarted and added ti back again, 

my battery is at 20% but shows 158%
[...]

PR #877, opened 2024-05-02
- skip duplicate gen4 / gen5 entities

simowNgithub:
- fixed for me the reboot issue, that you always have to drop and add the inverter to the solax integration after home assistant was resetartet
- fixed for me the issue with the wrong values, see details here https://github.com/wills106/homeassistant-solax-modbus/issues/813#issuecomment-2005136546


simowNgithub:
@wills106 i found, that your integration gets issues if you use the same entity like "battery_power_charge" in two or more code blocks for the same gen4 / gen5 definition. 

"battery_power_charge" should be a function, which get's calculated by battery_1_power_charge and battery_2_power_charge. Inverters with only one battery slot always got two values, the battery_1_power_charge and the summary battery_power_charge.

I was not deep enough in the code to get this work. Sorry.

wills106:
Your correct the same register shouldn't be read twice. I didn't realise this error was there as no way to test.

> "battery_power_charge" should be a function, which get's calculated by battery_1_power_charge and battery_2_power_charge. Inverters with only one battery slot always got two values, the battery_1_power_charge and the summary battery_power_charge.

That's quite easy to do, as it's what we do for "PV Power Total"

PR #890, opened 2024-05-11
Move import to async executor

PatrikTrestik:


wills106:
Thanks for working on this. I have been too busy lately to look at changing the way the plugin is loaded.

PR #905, opened 2024-05-20
Wait for determine inverter type

Zwer2k:
Waits for the inverter if Modbus communication is disrupted when Homeassistant is started

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -297,11 +297,16 @@ class SolaXModbusHub:
         self.entry = entry
         _LOGGER.debug("solax modbushub done %s", self.__dict__)
 
-    async def async_init(self, *args: Any) -> None:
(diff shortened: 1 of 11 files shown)


wills106:
Were you getting any of the currently open issues before making this change? Just wondering who to ask to test if it helps with their issue as well.

Zwer2k:
Not directly, but I have had the problem several times that the communication with Sofar inverters gets stuck due to a faulty firmware, if I restart HA and then the inverter, the communication does not work.
The current specification is that the inverter must be in operation when Homeassistant is restarted. If the inverter is not running when HA is started, no communication is established until the integration has been restarted. 
There are several messages about communication problems, it is possible that the context is not recognized.

wills106:
I forgot that Inverters go offline at night. I normally remind people to make sure they aren't restarting HA during the night.

But as mine never goes offline I forgot.

PR #906, opened 2024-05-21
X1 sofar_old: enable temperature sensors

kkubicki:
Corrected type allows single phase inverters to monitor temperatures.


wills106:
I'm closing this down and replacing it with:
`allowedtypes = PV,`

As I would imagine Single and 3Phase to both have this sensor.

PR #926, opened 2024-06-14
Add battery pack info as extra devices (currently only sofar BTS-5k)

Zwer2k:
- Support for battery pack information as extra devices (can be easily configured in plugins)
image
 
- Measured values can be read out from individual cells
image

- Number of battery packs is automatically recognized (currently only one battery)
- Support can be activated/deactivated in the configuration (Deactivated by default. Only offered if the plugin supports the function) 
image

- With BTS-5k, the battery pack must be selected before each readout; a register must be written for this purpose

- Display of the model, SW version and HW version implemented under device information for inverters and batteries  
image

- In the case of an error, 3 attempts are made to establish a connection to the inverter before an error message is issued

Translated with DeepL.com (free version)


Zwer2k:
Another addition:
Reading out the BTS-5k memory does not work very reliably. Despite correct selection and subsequent verification that the selection was successful, it happens again and again that the values are supplied by an incorrect battery pack. I suspect an error in the firmware of the inverter. 
To avoid incorrect values, I compare the serial number with the serial number stored in the device info at the end of each read process for a device group. If the serial number does not match, the values are discarded. This has worked very well so far. This procedure assumes that the first determination of the serial number was correct. To ensure this, the first determination is carried out until the serial number has been read correctly 5 times in succession.

wills106:
I was getting the following error when trying the PR
```
Error while setting up solax_modbus platform for sensor
  File "/usr/local/lib/python3.12/site-packages/homeassistant/helpers/entity_platform.py", line 356, in _async_setup_platform
    await asyncio.shield(awaitable)
  File "/config/custom_components/solax_modbus/sensor.py", line 108, in async_setup_entry
    entityToList(hub, hub_name, entities, groups, newgrp, computedRegs, hub.device_info,
  File "/config/custom_components/solax_modbus/sensor.py", line 212, in entityToList
    newdescr.name = name_prefix + newdescr.name
                    ~~~~~~~~~~~~^~~~~~~~~~~~~~~
TypeError: can only concatenate str (not "UndefinedType") to str
```
I got around this by changing `sensor.py` Line 212 and Line 213 from:

```
                newdescr.name = name_prefix + newdescr.name
                newdescr.key = key_prefix + newdescr.key
```
to
```
                newdescr.name = newdescr.name
                newdescr.key = newdescr.key
```

As Line 210 will only run on Non Sofar Inverters or when you disable Battery Module reading on Sofar.

Zwer2k:
Strange, the following code should actually do this correctly. Lines 104-106.
_inverter_name_suffix = ""
    if hub.inverterNameSuffix is not None and hub.inverterNameSuffix != "":
        inverter_name_suffix = hub.inverterNameSuffix + " "_

inverter_name_suffix is now defined in the settings and should simply extend the name in the view

image


The ELSE part (line 210) almost always also runs with Sofar and batteries. The IF (line 204) only runs for value_series
          name = "cell **{}** voltage",
          key = "cell_**{}**_voltage",
          native_unit_of_measurement = UnitOfElectricPotential.VOLT,
          device_class = SensorDeviceClass.VOLTAGE,
          register = 0x9051,
          scale = 0.001,
          rounding = 3,
          allowedtypes = BAT_BTS,
          **value_series = 16**
    ),

Zwer2k:
> I got around this by changing `sensor.py` Line 212 and Line 213 from:
> 
> ```
>                 newdescr.name = name_prefix + newdescr.name
>                 newdescr.key = key_prefix + newdescr.key
> ```
> 
> to
> 
> ```
>                 newdescr.name = newdescr.name
>                 newdescr.key = newdescr.key
> ```
> 
> As Line 210 will only run on Non Sofar Inverters or when you disable Battery Module reading on Sofar.

Please try instead in the file __init__.py line 267
```
self.inverterNameSuffix = config.get(CONF_INVERTER_NAME_SUFFIX)
```
to replace by
```
self.inverterNameSuffix = config.get(CONF_INVERTER_NAME_SUFFIX, None)
```

wills106:
Will the change break the Integration on your Inverter?
If it does I can add an if statement for Sofar Inverters to use your method and mine for other Inverters.

I have had to some something similar as this change:

https://github.com/wills106/homeassistant-solax-modbus/pull/926/files#diff-d11f1d1259d121f7a64f91141c1e0b4b6c0fd3af99faf1ed1e83a620b7a5236eL96-L107

https://github.com/wills106/homeassistant-solax-modbus/pull/926/files#diff-d11f1d1259d121f7a64f91141c1e0b4b6c0fd3af99faf1ed1e83a620b7a5236eR90-R102

Broke reading Hardware and Software Version

I corrected it with:
https://github.com/wills106/homeassistant-solax-modbus/commit/3e85eea65e477c6a573bc99fbb689efed1862c5f#diff-ab4036f64aa303016443a3889c83d1f4215c2bbb7e86b5fc0e28fb956e6261dc

wills106:
> Please try instead in the file **init**.py line 267

I didn't see this before posting my reply. I'll try it.

wills106:
I get the same error

Zwer2k:
> I get the same error

Strange. I'll have to try to reproduce it.
Does it also happen if the suffix is set in the configuration? 
image
Should actually be optional.
 
I have actually tried to implement it in such a way that no exceptions are required. 
With the change I suggested, it should actually also work with other inverters.
If someone wants to implement the battery extension for other inverters, they just need to extend the corresponding plugin.

wills106:
I found you need to remove the Inverters from the Integration before updating to your version as the changes in the config_flow break the Integration and you can't delete them.

So when doing a fresh install it defaults to "Inverter"

I'll have to add to the Release notes it's a breaking change, so you need to remove the Inverter(s) from the Integration before updating.

I have replaced:
https://github.com/wills106/homeassistant-solax-modbus/blob/ee8f5bb77add60ed53261645bc13b8c5100f5aaa/custom_components/solax_modbus/sensor.py#L203-L214

with the following:

```
            if hub.plugin.plugin_name != 'Sofar':
                newdescr = copy(sensor_description)
                newdescr.name = newdescr.name
                newdescr.key = newdescr.key
                entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
            else:
                if sensor_description.value_series is not None:
                    for serie_value in range(sensor_description.value_series):
                        newdescr = copy(sensor_description)
                        newdescr.name = name_prefix + newdescr.name.replace("{}", str(serie_value+1))
                        newdescr.key = key_prefix + newdescr.key.replace("{}", str(serie_value+1))
                        newdescr.register = sensor_description.register + serie_value
                        entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
                else:
                    newdescr = copy(sensor_description)
                    newdescr.name = name_prefix + newdescr.name
                    newdescr.key = key_prefix + newdescr.key
                    entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
```

It's now working again and shouldn't break your setup.

wills106:
I can always invert the if statement in the future if other Inverters provide detailed battery info like Sofar does.

On my SolaX even though I three batteries they are treated as a single entity.

Zwer2k:
image

You do not have to make this differentiation, I have only removed hw_version and sw_version because they have not been used so far and I have provided the functions getSoftwareVersion and getHardwareVersion in the plugins (see Sofar plugin line 4053). This has two advantages: When starting, the integration does not block as long (only for determining the serial number) and the integration is informed when the software version changes (does not have to be restarted).  
With Sofar, the software version and hardware version are simply taken from existing entities, but the functions can also contain their own logic if required.

Zwer2k:
> I can always invert the if statement in the future if other Inverters provide detailed battery info like Sofar does.
> 
> On my SolaX even though I three batteries they are treated as a single entity.

If others try to make the extension, they would wonder why it doesn't work.
Can you also read out individual battery cells with Solax?


```
            if sensor_description.value_series is not None:
                for serie_value in range(sensor_description.value_series):
                    newdescr = copy(sensor_description)
                    newdescr.name = name_prefix + newdescr.name.replace("{}", str(serie_value+1))
                    newdescr.key = key_prefix + newdescr.key.replace("{}", str(serie_value+1))
                    newdescr.register = sensor_description.register + serie_value
                    entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
            else:
                newdescr = copy(sensor_description)
                try:
                   newdescr.name = name_prefix + newdescr.name
                except:
                   newdescr.name = newdescr.name
                   
                newdescr.key = key_prefix + newdescr.key
                entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
```
Maybe like this?
There should be no problems with key_prefix.

wills106:
> This has two advantages: When starting, the integration does not block as long (only for determining the serial number) and the integration is informed when the software version changes (does not have to be restarted). With Sofar, the software version and hardware version are simply taken from existing entities, but the functions can also contain their own logic if required.

I'm trying to minimise the amount of registers read.
Although I could default them to a super slow poll rate or something instead.

> getSoftwareVersion and getHardwareVersion

I gave it a quick go, but when I tried two Inverters it was reading the second inverter. But I think I know why. I'll have another look into it over the weekend.

> Can you also read out individual battery cells with Solax?

No, even though I have 3 batteries they only show as one.

There are different batteries for SolaX on the market now. They may behave different.

Zwer2k:
> I'm trying to minimise the amount of registers read.
> Although I could default them to a super slow poll rate or something instead.

Exactly, I think an extra slow reading group would be a good solution

The error with name_prefix would also occur with Sofar as long as Invertername-Prefix is not set in the configuration. Therefore I think the try-except solution would be better

wills106:
> ```
>             if sensor_description.value_series is not None:
>                 for serie_value in range(sensor_description.value_series):
>                     newdescr = copy(sensor_description)
>                     newdescr.name = name_prefix + newdescr.name.replace("{}", str(serie_value+1))
>                     newdescr.key = key_prefix + newdescr.key.replace("{}", str(serie_value+1))
>                     newdescr.register = sensor_description.register + serie_value
>                     entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
>             else:
>                 newdescr = copy(sensor_description)
>                 try:
>                    newdescr.name = name_prefix + newdescr.name
>                 except:
>                    newdescr.name = newdescr.name
>                    
>                 newdescr.key = key_prefix + newdescr.key
>                 entityToListSingle(hub, hub_name, entities, groups, newgrp, computedRegs, device_info, newdescr, readPreparation, readFollowUp)
> ```
> 
> 
>     
>   
> 
> Maybe like this?
> There should be no problems with key_prefix.

I have reverted all by changes back to your PR with the addition of the above code change.

I can't get the Hardware or Software versions into the Device info panel.
I think it's to do with:

https://github.com/wills106/homeassistant-solax-modbus/blob/6c5dc0ce1e2c1b419c3690ad00243c2f6bb616eb/custom_components/solax_modbus/plugin_solax.py#L7238-L7242

I have the `software_version` working as a standalone sensor:
https://github.com/wills106/homeassistant-solax-modbus/blob/6c5dc0ce1e2c1b419c3690ad00243c2f6bb616eb/custom_components/solax_modbus/plugin_solax.py#L5978-L5983

Built from:
https://github.com/wills106/homeassistant-solax-modbus/blob/6c5dc0ce1e2c1b419c3690ad00243c2f6bb616eb/custom_components/solax_modbus/plugin_solax.py#L293-L294

Which displays correctly in the normal Sensor list.
image


I think the issue is to do with `new_data` do I need to first pass "hardware_version" and "software_version" from the sensor list to `new_data`

I have noticed you are using `new_data` in:
https://github.com/wills106/homeassistant-solax-modbus/blob/6c5dc0ce1e2c1b419c3690ad00243c2f6bb616eb/custom_components/solax_modbus/plugin_sofar.py#L3869
and
https://github.com/wills106/homeassistant-solax-modbus/blob/6c5dc0ce1e2c1b419c3690ad00243c2f6bb616eb/custom_components/solax_modbus/plugin_sofar.py#L3900

Do I need to create empty versions of those for `getSoftwareVersion` and `getSoftwareVersion` to function correctly?

Zwer2k:
The readFollowUp function is called before value_function. In readFollowUp, getSoftwareVersion and getHardwareVersion are called in turn.  

Try replacing async_read_modbus_registers_all with the following code:
```    
async def async_read_modbus_registers_all(self, group):
        if group.readPreparation is not None:
            if not await group.readPreparation(self.data):
                _LOGGER.info(f"device group read cancel")
                return True
        else:
            _LOGGER.debug(f"device group inverter")

        data = {}
        res = True
        for block in group.holdingBlocks:
            res = res and await self.async_read_modbus_block(data, block, "holding")
        for block in group.inputBlocks:
            res = res and await self.async_read_modbus_block(data, block, "input")

        if self.localsUpdated:
            self.saveLocalData()
            self.plugin.localDataCallback(self)
        if not self.localsLoaded:
            self.loadLocalData()
        for reg in self.computedSensors:
            descr = self.computedSensors[reg]
            self.data[descr.key] = descr.value_function(0, descr, self.data)

        if group.readFollowUp is not None:
            if not await group.readFollowUp(self.data, data):
                _LOGGER.warning(f"device group check not success")
                return True

        for key, value in data.items():
            self.data[key] = value

        if (
            res and self.writequeue and self.plugin.isAwake(self.data)
        ):  # self.awakeplugin(self.data):
            # process outstanding write requests
            _LOGGER.info(
                f"inverter is now awake, processing outstanding write requests {self.writequeue}"
            )
            for addr in self.writequeue.keys():
                val = self.writequeue.get(addr)
                await self.async_write_register(self._modbus_addr, addr, val)
            self.writequeue = {}  # make sure we do not write multiple times
        self.last_ts = time()
        for (
            k,
            v,
        ) in self.data["_repeatUntil"].items():
            if self.last_ts < v:
                buttondescr = self.computedButtons[k]
                payload = buttondescr.value_function(0, buttondescr, self.data)
                _LOGGER.debug(f"ready to repeat button {k} data: {payload}")
                await self.async_write_registers_multi(
                    unit=self._modbus_addr,
                    address=buttondescr.register,
                    payload=payload,
                )
        return res
```

wills106:
Unfortunately the behaviour is the same.
I can't see `software_version` or `hardware_version` in Device info

Zwer2k:
Something else forgotten. 
Line:
```self.data[descr.key] = descr.value_function(0, descr, self.data)```

Replace with:
```data[descr.key] = descr.value_function(0, descr, data)```

**Comment by wills106** [comment]
Thanks it's working now.

**Comment by wills106** [comment]
Tested ok with Gen3 SolaX and an emulated MIC or MAX/MEGA. But when emulating a Gen4 I get the following error.

```
Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:693
integration: SolaX Inverter Modbus ([documentation](https://homeassistant-solax-modbus.readthedocs.io/), [issues](https://github.com/wills106/homsassistant-solax-modbus/issues))
First occurred: 19:57:02 (1 occurrences)
Last logged: 19:57:02

Something went wrong reading from modbus
  File "/config/custom_components/solax_modbus/__init__.py", line 693, in async_read_modbus_data
    res = await self.async_read_modbus_registers_all(group)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/solax_modbus/__init__.py", line 881, in async_read_modbus_registers_all
    data[descr.key] = descr.value_function(0, descr, data)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/solax_modbus/plugin_solax.py", line 271, in value_function_remotecontrol_autorepeat_remaining
    return autorepeat_remaining(datadict, 'remotecontrol_trigger', time())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/solax_modbus/const.py", line 234, in autorepeat_remaining
    remaining = datadict['_repeatUntil'].get(entitykey,0) - timestamp
                ~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: '_repeatUntil'
```

I'll pick this up again in a few days. Just leaving this error in case either yourself or @litinoveweedle can spot anything obvious.
I don't have enough time to finish trouble shooting tonight.

**Comment by litinoveweedle** [comment]
Hi,

it seems that you in some cases call `autorepeat_remaining()` with wrong parameters order:

```
const.py:def autorepeat_remaining(datadict, entitykey, timestamp):
plugin_solax_a1j1.py:    return autorepeat_remaining(datadict, 'remotecontrol_trigger', time())
plugin_solax.py:def value_function_remotecontrol_autorepeat_remaining(initval, descr, datadict):
plugin_solax.py:    return autorepeat_remaining(datadict, 'remotecontrol_trigger', time())
plugin_srne.py:    return autorepeat_remaining(datadict, 'remotecontrol_trigger', time())
```
so

`plugin_solax.py:def value_function_remotecontrol_autorepeat_remaining(initval, descr, datadict):`

`datadict` shall be first parameter. ;-)

Zwer2k:
@wills106 
How can I emulate Gen4?

litinoveweedle:
Hopefully not another stupid remark, but where do you init `datadict['_repeatUntil']` if it is not previously set by `autorepeat_set`?

Shouldn't you check if key `_repeatUntil` key exists in `datadict`?

wills106:
I just use ModRSsim2 on PC, you just need to have enough of a Gen4 Serial Number for the Integration to detect it as a Gen4 or what ever Inverter.

image

First two Registers is all that's required for a Gen4. Some inverters like the MIC's etc need 3 or 4 Registers.

Then I add a second hub directing it to my PC. (Following Pics show the second Hub setup as a Gen1 MIC)

image

You don't really need to populate any values to perform a basic test.

image

Just for info I didn't write any of the [modbus power control](https://homeassistant-solax-modbus.readthedocs.io/en/latest/modbus-power-control/) code. So I don't quite fully understand it myself.

litinoveweedle:
Ok, I really do not see where datadict[''_repeatUntil''] is initialized as dictionary....

Zwer2k:
Originally .data was taken directly from the hub where _repeatUntil was initialized. 
With the current change, a temporary version of "data" is created in the function async_read_modbus_registers_all in order to be able to discard the data if necessary (e.g. if there were read errors). In the temporary version, _repeatUntil is not initialized, which is why the error occurs. The question is whether _repeatUntil has to be split between the reads or not.

wills106:
@litinoveweedle If you decide to test 2024.07.0.b1 (or anyone else reading this) it's a breaking change due to a change in `config_flow`.
So you need to remove the Inverter(s) from the Integration, upgrade and add them back in a gain following a HA restart.

I'm not sure how to migrate the `config_flow`.

Zwer2k:
> @litinoveweedle If you decide to test 2024.07.0.b1 (or anyone else reading this) it's a breaking change due to a change in `config_flow`. So you need to remove the Inverter(s) from the Integration, upgrade and add them back in a gain following a HA restart.
> 
> I'm not sure how to migrate the `config_flow`.

If the values are not intended to be shared between the runs, _repeatUntil can be initialized as a dictionary in the function async_read_modbus_registers_all

Zwer2k:
> Originally .data was taken directly from the hub where _repeatUntil was initialized. With the current change, a temporary version of "data" is created in the function async_read_modbus_registers_all in order to be able to discard the data if necessary (e.g. if there were read errors). In the temporary version, _repeatUntil is not initialized, which is why the error occurs. The question is whether _repeatUntil has to be split between the reads or not.

If the values are not intended to be shared between the runs, _repeatUntil can be initialized as a dictionary in the function async_read_modbus_registers_all

litinoveweedle:
> If the values are not intended to be shared between the runs, _repeatUntil can be initialized as a dictionary in the function async_read_modbus_registers_all

I think that datadict[''_repeatUntil'']  shall be kept between runs...

Zwer2k:
> @litinoveweedle If you decide to test 2024.07.0.b1 (or anyone else reading this) it's a breaking change due to a change in `config_flow`. So you need to remove the Inverter(s) from the Integration, upgrade and add them back in a gain following a HA restart.
> 
> I'm not sure how to migrate the `config_flow`.

In my tests, the try -except change did not lead to any problems with the config flow

litinoveweedle:
Honestly I do not see reason, why `autorepeat_set, autorepeat_stop, autorepeat_remaining` should not work over `self.data` instead of `datadict`. IMHO those has nothing to do directly with the modbus register reading...

wills106:
> In my tests, the try -except change did not lead to any problems with the config flow

I'll have to retest again tomorrow on my Test HA. Perhaps some of the other issues caused the fault?

One thing I have noticed with the Name Suffix "Inverter" addition is that on the First Hub it just displays `Grid Export` or `Grid Import` on the sensor list which is ok.

image

Although when you click on the sensor it lists Inverter as part of the name.

image

But on the second Hub it's including the full name on the Sensor list

image

So the names are getting quite long. It might just be how HA handles additional Hubs. I'll go back and test on 2024.05.7 again in the next day or two.

Is the addition of "Inverter" in the name necessary?

Zwer2k:
datadict was intended for passing through self.data . "self" does not exist in the function. 

Then the solution should be
line:
```data = {}```

with the following:
```data = {"_repeatUntil": self.data["_repeatUntil"]}```

Zwer2k:
image
The problem with long names actually always occurs when the wrong main prefix is entered. But that was already the case in the past. I entered "Sofar Solar" as the prefix at the very beginning and was annoyed all the time that the names were so long, then had to rename the entities in the DB in order not to lose the statistics.


> Is the addition of "Inverter" in the name necessary?

image
No, it is not absolutely necessary, it was only intended for a nicer presentation in the list

Zwer2k:
> I just use ModRSsim2 on PC, you just need to have enough of a Gen4 Serial Number for the Integration to detect it as a Gen4 or what ever Inverter.

Very good, I didn't know about this tool. Can be quite helpful

Zwer2k:
> datadict was intended for passing through self.data . "self" does not exist in the function.
> 
> Then the solution should be line: `data = {}`
> 
> with the following: `data = {"_repeatUntil": self.data["_repeatUntil"]}`

This works, I have tested it with ModRSsim2

Zwer2k:
This is exactly the line

fxstein:
@wills106 This great PR introduced the a name suffix in the configuration flow. While that is great for those that want it, It defaults the name to "Inverter" and many of us always have to remember to delete that default. Would it be possible to keep the setting but make the default empty? 

See here #1732

Issue #920, opened 2024-06-10
HA 2024.6.0 Detected blocking call to open inside the event loop by custom integration 'solax_modbus' 


litinoveweedle:
### Describe the bug

After upgrade to HA 2024.6.0 during the startup new warning appear during solax_modbus integration init. Those are linked to the additional HA checks for identifying integration blocking calls in the HA event loop.

```
Log details (WARNING)
Logger: homeassistant.util.loop
Source: util/loop.py:84
First occurred: 5:45:10 PM (6 occurrences)
Last logged: 5:46:25 PM

Detected blocking call to open inside the event loop by custom integration 'solax_modbus' at custom_components/solax_modbus/__init__.py, line 332: fp = open(self._hass.config.path(f"{self.name}_data.json")) (offender:
```

see log bellow.

### Integration Version

 2024.05.7

### Homeassistant core version

2024.6.0

### Inverter brand

doesn't matter

### Plugin used

plugin_growatt.py

### Serial prefix

doesn't matter

### Connection Method

doesn't matter

### Detailed Error Log

2024-06-10 17:45:25.030 WARNING (MainThread) [homeassistant.util.loop] Detected blocking call to open inside the event loop by custom integration 'solax_modbus' at custom_components/solax_modbus/__init__.py, line 332: fp = open(self._hass.config.path(f"{self.name}_data.json")) (offender: /config/custom_components/solax_modbus/__init__.py, line 332: fp = open(self._hass.config.path(f"{self.name}_data.json"))), please create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues
Traceback (most recent call last):
File "<frozen runpy>", line 198, in _run_module_as_main
File "<frozen runpy>", line 88, in _run_code
File "/usr/src/homeassistant/homeassistant/__main__.py", line 223, in <module>
sys.exit(main())
File "/usr/src/homeassistant/homeassistant/__main__.py", line 209, in main
exit_code = runner.run(runtime_conf)
File "/usr/src/homeassistant/homeassistant/runner.py", line 190, in run
return loop.run_until_complete(setup_and_run_hass(runtime_config))
File "/usr/local/lib/python3.12/asyncio/base_events.py", line 672, in run_until_complete
self.run_forever()
File "/usr/local/lib/python3.12/asyncio/base_events.py", line 639, in run_forever
self._run_once()
File "/usr/local/lib/python3.12/asyncio/base_events.py", line 1988, in _run_once
handle._run()
[...]

PR #946, opened 2024-06-30
moved blocking file operations into async executor job

litinoveweedle:
Fix for #920 as discussed.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -872,10 +872,10 @@ class SolaXModbusHub:
             res = res and await self.async_read_modbus_block(data, block, "input")
 
         if self.localsUpdated:
-            self.saveLocalData()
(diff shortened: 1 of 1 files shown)


wills106:
Thanks, I was trying to turn:
https://github.com/wills106/homeassistant-solax-modbus/blob/140b0ebc2618a124cc51cfcebda8112ed0239e93/custom_components/solax_modbus/__init__.py#L357

into an async call 😓 no wonder I couldn't get it to work!

litinoveweedle:
> into an async call 😓 no wonder I couldn't get it to work!

No problem, happy to help. ;-)

File open would be always considered as sync/blocking, therefore the proper way is to call the function from HA job executor which would schedule it in async. So there is no need to convert `loadLocalData` into coroutine. It is documented [here](https://developers.home-assistant.io/docs/asyncio_working_with_async#calling-sync-functions-from-async).

PR #997, opened 2024-08-14
Max Battery charge/uncharge set to 35A

Hetzi74:
My Solax battery can work with more then 20A current on the battery. The maximum according to the data sheet is 35A. 

See data sheet: https://de.solaxpower.com/uploads/file/t-bat-sys-hv-5.8-t58-datasheet-de.pdf

I'm not sure if this battery is also compatible with Hybrid/Gen2 in the next sections.


wills106:
The correct way it to add an entry here:
https://github.com/wills106/homeassistant-solax-modbus/blob/a7e4cd7f0d5968f4e174243fa7b89d723a967701/custom_components/solax_modbus/plugin_solax.py#L297

Can you actually set the charge / discharge to 35A in the web portal?

PR #1002, opened 2024-08-19
Fix: use retries parameter instead of fixed value

lhorak:
It looks like a reties parameter was introduced which should perform the retry 6 times for `self._client.connect()` but the parameter was ignored and the retry only happened twice based on the hardcoded value


lhorak:
I'll close this, since it's not helping fix the write issue, but I have another PR which fixes the issue, will raise that shortly

Issue #998, opened 2024-08-17
[Bug]: Can not control values from HA


toko-dotcom:
### Describe the bug

Since update 2024.08.08 I am not able to change values in "SolaX Inverter Export Control User Limit" via HA. It is strange behavior, because I test it several times and in non expected period some of request change value. But after that next 20 times nothing happend and only get back error Modbus erroc connection. Client is not connected. But all values I see in overview actively chaning so modbus is connected all time. 

### Integration Version

2024.08.08

### Homeassistant core version

2024.8.2

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

H34A10

### Inverter firmware versions

Dsp  v1.41 arm 1.40

### Connection Method

Solax pocket wifi 3.0

### Dongle firmware

3.006.04

spetrie13:
I have similar issues with Solis inverter and EW11 modbus device. Was working perfectly for over 1 year with setup until the latest updates.

mobyline:
The same issues here with X3 G4. 
A few month ago it worked fine.

toko-dotcom:
I also have x3 g4 10kw. I use wifi pocket 3.0  modbus because I have EVC connected via Comm port. As i read in different issue report, there is not solution to have EVC and Modbus TCP RS485 on one Comm port. So my only one solution is to stay Wifi Pocket 3.0  modbus and EVC via Comm port. 

I can confirm that some weeks ago I was able set inverter configuration from HA via Wifi pocket 3.0 modbus.

Thank you for help in advance

Btw: what already tried:
Uninstall Solax inverter and modbus from HA and delete all related setting
Install solax modbus via hacks and updated repositories
Setup automatization with continue on error (no help, all time stops on failure which I wrote upper)
Setup loop in automatization until value will be changed ( no help, all time stops on failure which i wrote upper)

The strangest is sometimes by manual updating via HA configuration is accepted and written to inverter. But it is not possible to say that for example each 20 times it is ok or something like that.

Read entity information from Inverter to HA works fine without any issue. Still with wifi pocket 3.0 modbus.

Hetzi74:
On my inverter Solax X3-Ultra 15k i can change the value for number.solax_export_control_user_limit and sometimes i get that modbus error too. 

But the value i seen in HA are 10-times off. So a value of 500W in HA is 5000W on the inverter itself, so there is allways a faktor of 10 difference between the visible HA value and the value at the inverter.

[...]

Issue #993, opened 2024-08-09
unable to control inverter following upgrade [Bug]: 


luke3579:
### Describe the bug

after upgrading to 2024.08.** i am unable to control the inverter, selecting inverter charger use modes, adjusting the charging times etc. error comes up when adjusting the nightcharge upper SOC option for example is "failed to perform action number/set_value. Modbus error: [Connection] Client is not connected" 
HA can ping the WiFi adapter on the inverter, no dropouts etc. 
- tried unplugging and replugging the dongle.
- tried removing inverter, removing the integration, restarting HA, redownloading the integration, restarting HA, readding the inverter... still the same

### Integration Version

2024.08.8

### Homeassistant core version

2024.8.0

### Inverter brand

SolaX

### Plugin used

plugin_solax.py

### Serial prefix

H4372A

### Inverter firmware versions

DSP v1.36 ARM v1.34

### Connection Method

Pocket WiFi 3.0

### Dongle firmware

3.015.02

### Detailed Error Log

2024-08-09 09:40:09.321 ERROR (MainThread) [pymodbus.logging] Cancel send, because not connected!
2024-08-09 09:40:09.321 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [139676680027392] Unexpected exception
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 241, in handle_call_service
response = await hass.services.async_call(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/core.py", line 2763, in async_call
response_data = await coro
^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/core.py", line 2806, in _execute_service
return await target(service_call)
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 999, in entity_service_call
single_response = await _handle_entity_call(
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1071, in _handle_entity_call
result = await task
^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/components/number/__init__.py", line 120, in async_set_value
await entity.async_set_native_value(native_value)
File "/config/custom_components/solax_modbus/number.py", line 151, in async_set_native_value
await self._hub.async_write_register(
[...]

Issue #992, opened 2024-08-09
[Bug]: upgraded to 2024.08.8 Now unable to control inverter through modbus


UK-JustinSmall:
### Describe the bug

I have problem with inverter not saving values when they are updated through HA

Values change but inverter does not take action

E.g. I change 
select.solax_charger_use_mode to "Manual Mode" and

select.solax_manual_mode_select to "Force Discharge"

I expect the battery to force discharge, but nothing happens, the values change in the UI, but battery does not force discharge!!!

I have waveshare Modbus connection to Solax Inverter and can see the values of the entities OK.

### Integration Version

2024.08.8

### Homeassistant core version

2024.8.0

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

H4752A

### Inverter firmware versions

DSP v1.42 ARM v1.39

### Connection Method

Waveshare Adapter / Wired Ethernet. 

### Dongle firmware

3.015.02

### Additional context

I have attempted a fresh install by removing the previous intigration custom_components/solax_modbus folder on server, restarting HA, and doing fresh install from HACS this fixed a issue I was having with it giving a error about "Failed to call service select/select option Modbus connection client not connected"

But as I say it's no longer controlling the inverter, but values now change and don't get the above modbus connection error.

wills106:
Just released [2024.08.9](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.08.9) reverting the `client.connect` change from 2024.08.1

wills106:
@UK-JustinSmall or @fastsonic any feedback? Can this be closed now?

UK-JustinSmall:
From my side can be closed thank you :-)Sent from my iPadOn 7 Sep 2024, at 19:50, wills106 ***@***.***> wrote:﻿
@UK-JustinSmall or @fastsonic any feedback? Can this be closed now?

PR #1003, opened 2024-08-19
Fix: Modubus Client not connected error

lhorak:
Since I wasn't able to write any values through the modbus connection, this fixed the issue for me and I can write all the values just fine now. 

According to the documentation of pymodbus, the client should be trying to reconnect automatically within the given timout:

The line await client.connect() connects to the device (or comm port), if this cannot connect successfully within the timeout it throws an exception. If connected successfully reconnecting later is handled automatically

This might be a quick fix, maybe there should be the asyncio lock applied in the for loop to keep the retry functionality?

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -508,7 +508,7 @@ class SolaXModbusHub:
 
         return self._client.connected
 
-    async def async_connect(self, retries=6):
+    async def async_connect(self):
         result = False
 
         _LOGGER.debug(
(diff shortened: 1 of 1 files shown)


lhorak:
Potentially resolves #998 #993 #992 - will need more testing from others

lhorak:
@wills106 can I get your review on this please, I believe this could close a lot of the reported issues mentioned above

wills106:
Thanks, I was going to look at reverting that change like you have done. But I have been too busy.

I find it strange that my Inverter always seems to just work when others are having issues. So it's hard to troubleshoot at times.

lhorak:
I understand, no worries, feel free to ping me if you need help testing, I have Solax X3 Hybrid Gen4, so I can help at least wit that one. 

Should I also create a PR with bumping the version so it gets to HACS or can you do that?

wills106:
I'm just doing some updates myself for another model of the Gen5 Hybrids. Then I'll push a release out.

Issue #976, opened 2024-08-04
[Bug]: 'SolaX' - 'Gen3' crash on add integration "KeyError: 'support-battery'"


MrRickwi:
### Describe the bug

Same like Growatt bug. Reported here to have different label on it. Additionally, some entities are missing after getting the integration up and running with a hack. See also other reported bug.

### Integration Version

2024.08.02

### Homeassistant core version

2024.7.4

### Inverter brand

SolaX Gen3

### Plugin used

plugin_solax.py

### Serial prefix

H34C06

### Inverter firmware versions

DSP v1.29 ARM v1.31

### Connection Method

USB Modbus Adapter

### Dongle firmware

None

MrRickwi:
Latest version is working for me. Missing entities have been found, are disabled by default.

wills106:
I'll close this down then.

ruimmsimoes:
I’m having the same issue when adding a SolaX inverter after upgrading to 2024.08.08

wills106:
Just released [2024.09.3](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.09.3) which hopefully fixes this error.

Issue #968, opened 2024-07-25
[Bug]: 'Growatt' - 'SPF5000EU' crash on add integration "KeyError: 'support-battery'"


Evka2k:
### Describe the bug

Add integration -> Select Interface:Serial, Inverter type: growatt or any other -> Serial Interface Parameters: Submit

![crash](https://github.com/user-attachments/assets/01269603-a7cb-4567-b7a5-a9b55ec0c8d9)

<pre>
home-assistant  |     next_step_id_or_end_flow = await form_step.next_step(self._options)
home-assistant  |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
home-assistant  |   File "/config/custom_components/solax_modbus/config_flow.py", line 193, in _next_step_battery
home-assistant  |     if user_input["support-battery"]:
home-assistant  |        ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
home-assistant  | KeyError: 'support-battery'
</pre>

### Integration Version

2024.07.0b3

### Homeassistant core version

2024.7.3

### Inverter brand

growatt

### Plugin used

plugin_growatt.py

### Serial prefix

113.02

### Connection Method

RS485-USB

### Detailed Error Log

[crash.log](https://github.com/user-attachments/files/16382329/crash.log)

### Additional context

This fix prevents from crash.

<pre>
--- custom_components/solax_modbus/config_flow.py
+++ custom_components/solax_modbus/config_flow.py
@@ -190,6 +190,8 @@ async def _next_step_modbus(user_input: Any) -> str:
 
 async def _next_step_battery(user_input: Any) -> str:
     _LOGGER.debug(f"_next_step_battery: returning data: {user_input}")
+    if "support-battery" not in user_input.keys():
+        return None
     if user_input["support-battery"]:
         return "battery"
     return None
</pre>

Beunhaaz:
I'm having the same problem on my Growatt SPH8000 but unfortunately your fix didn't work for me.

Now I get:
`  File "/config/custom_components/solax_modbus/config_flow.py", line 193, in _next_step_battery
    if "support-battery" not in user_input.keys():
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'support-battery'`

It seems to trip over the first support-battery it encounters.

MrRickwi:
[...]

PR #1054, opened 2024-09-15
Fix support-battery error for serial inverters #976

noessi5:
Adding serial inverters was not possible for me without this change


noessi5:
Same issue as #968 - I only took over the suggested fix from @siebeclaes

PR #1068, opened 2024-09-25
Growatt MOD TL3-XH (GEN3): Add Time Slot 1 settings for priority mode change

TonyMathiesen:
Growatt MOD TL3-XH: 

Add registers for Time Slot 1 settings in order to set priority mode. Will only expose registers and needs to be set in decimal. 
Register Time 1 Start must be set first and Time 1 End within seconds after, otherwise it will be refused. Also it cannot overlap with other time registers. 

Register 3038 values:
    Minutes bit 0-7 (+mins)
    Hours bit 8-12 (+hours*256)
    Priority bit 13-14: load first = +0, battery first = +8192, grid first = +16384
    Enable bit 15: disabled = +0, enabled = +32768
Register 3039 values:
    Minutes bit 0-7 (+mins)
    Hours bit 8-12 (+hours*256)
    Reserved bits 13-15: 000 

Currently I have only added Time Slot 1, as it requires some Home Assistant helpers and an automation to make user friendly - however it at least give the opportunity to control through HA.


wills106:
> Register Time 1 Start must be set first and Time 1 End within seconds after, otherwise it will be refused.

Sounds like you need to use write_multi

Have a look at

https://github.com/wills106/homeassistant-solax-modbus/blob/e132ec0a1282a2b20924c638a1088bb374857d5d/custom_components/solax_modbus/plugin_solis.py#L157-L165

https://github.com/wills106/homeassistant-solax-modbus/blob/e132ec0a1282a2b20924c638a1088bb374857d5d/custom_components/solax_modbus/plugin_solis.py#L223-L254

Coupled with your bit handling.

TonyMathiesen:
Looks really interesting, might be what I need here - thanks😊 .. Will see if I can improve it 😉

TonyMathiesen:
Changed this pull request to utilize the write_multi and value_function. Really nice functionality in the integration 👏
image

Selects for setting the values:
image
Similar to UI from Growatt: 
image

I'm struggling to get select updated with the sensor, so currently I had to add "read" version to be able to see if the changes have been made: 
image
I have spend days trying adding the prevent_update from number to select, and that I can get to work with tmpdata - also on shown sensor values by modifying select.py and sensor.py. 
However I cannot get button function which pulls the values using "datadict.get('time_1_begin', '00:00')" to use the selected value. If the sensors (also internal) is read meanwhile it will always send the modbus value it has read, not what is selected. 
I could get it to work on "Time 1 End" as this is the only one on register 3039, but for Time 1 Begin/Mode/Enabled which derived from the same register 3038 I could not.

Does the current structure in the integration support having multiple entities derived from a single register without having the values overwritten every time the integration reads the modbus addresses? 
Regardless, I still believe others can use it as it is now, as it gives the possibilities to control the priority mode using HA 😊

If we proceed with the current read sensors, I will add Time Slot 2 and 3—then it should cover everyone's needs.

wills106:
> I have spend days trying adding the prevent_update from number to select, and that I can get to work with tmpdata - also on shown sensor values by modifying select.py and sensor.py.
> However I cannot get button function which pulls the values using "datadict.get('time_1_begin', '00:00')" to use the selected value. If the sensors (also internal) is read meanwhile it will always send the modbus value it has read, not what is selected.

Is your PR the final version as I can't see a modified select.py

You need `prevent_update = True,` on each of your selects to prevent it from updating on each poll cycle.

TonyMathiesen:
Thanks for the inputs. Had tried with the prevent_update and changes to select.py, and a lot of reboots.
However, tried again after your inputs and now it appears to be working 😊👏.. 

Will do some more testing, before adding Time Slot 2 & 3.

wills106:
Glad to hear your making progress on it!

TonyMathiesen:
Have marked it as draft again, as there are some situations where it still seems to write the original value and not the selected one, depending on some timing. Then I have to await until the TMPDATA_EXPIRY has cleared everything, before I can try again. 
Will have to look into it further.. But will stop for today

TonyMathiesen:
Hi @wills106  
I have tested for some days now, and have also done some debugging. With the 3 entities derived from one registers, there are some situations were the prevent_update causes issues. The Select itself works, but when calling the values with the button function to write it then uses the read value from modbus if a read cycle has been done. One cannot see that it is writing the old value and you have to wait 2 minutes before the values revert to original to see what the values actually are, or before one can try again if not written almost instantly. 

I would recommend that we keep the specific read sensors for the select values in this case. Because then the user can directly see what the values are on next read cycle. I know the user experience would be a bit better with the select entities being updated with the register value - but in this case, I believe it would create more frustration not knowing that the value has been changed to and having the 2 minutes waiting time. I have had several times where I though I had updated the timeslots or mode, only to find out later that it was not changed in the inverter. 

If you agree, then I will clean up this PR and have it prepared for merging?

TonyMathiesen:
I made a mess in this PR, have created a new #1097 and will close this one.

wills106:
I probably didn't help by shifting GEN, GEN2 & GEN3 back to 2, 3 & 4. While you have been working on this trying to support the TL3-SL.

I'll try and catch up with all of your PR's over the next few days / weekend.
Thanks for all of the PR's though!

TonyMathiesen:
No worries, and no rush 🙂.. Have a great day

PR #1081, opened 2024-10-06
enhanced SolaXModbusHub to utilize core ModbusHub connection

hernot:
Adds the possibility to attach to the homeassistant core ModbusHub as a third type of connection. This requires that the core Modbus module is properly configured and registers can be read from the SolaX Inverter via the core Modbus component. If selected only the name of the Modbus Hub has to be specified. At the moment the name has to be typed into the field. 

This pull request reflects the needs for  my personal setup. I do appriciate any feedback, discussion, suggestions, recommendations etc.  I will update given sufficient time and resources.


wills106:
Thanks I'll take a look at the PR.

PR #1082, opened 2024-10-06
Solax Gen4 meter direction and core meter direction fix

fxstein:
This PR contains:

- Adds the missing meter direction register definitions for GEN4 inverters 
- Adds meter direction logic to measured power to avoid incorrect calculations for meters configured with positive meter flow. This fixes a major issue with incorrect house load calculations when a meter is configured with positive meter flow

To accomplish this the `measured_power` register is imported a as `measured_power_raw`. A new computed function has been added as `value_function_measured_power` that applies the direction to the raw data. That way all computations that leverage measured power inherit the fix without additional code changes.

I have been able to test this fix on Solax X3 Hybrid G4 inverters.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -193,6 +193,19 @@ def value_function_hardware_version_g4(initval, descr, datadict):
 def value_function_hardware_version_g5(initval, descr, datadict):
     return  f"Gen5"
 
+# Function to apply meter direction to raw power measurements
+def value_function_measured_power(initval, descr, datadict):
+    meter_1_direction = datadict.get('meter_1_direction', 'Negativ')
+    _LOGGER.debug(f"meter_1_direction: {meter_1_direction}")
+    
+    # If meter is set to Positiv direction it reports imported energy as negative values
+    # If meter is set to Negative direction it reports imported energy as positive values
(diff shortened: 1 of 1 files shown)


fxstein:
I only added the logic fix to `measured_power` but not to the individual phases. Maybe something to consider.

fxstein:
Also to consider: This PR only adds the missing meter direction registers to GEN4 devices. I don't have access to GEN5 inverters but changes are the same fix wold apply to them.

wills106:
I'll check the Gen5 docs.

The majority of the registers are the same. But there are a few sections where they drift apart, quite strange really.

fxstein:
Since this PR got reverted can we reopen it and provide specifics on what goes wrong for Gen3 inverters? Without this PR Gen4 inverters that use an external meter set to positive flow create totally incorrect calculations and lead to vastly out of bounds grid imports.

wills106:
On my Gen3 both Meters are set as Positive, although I don't use CT Clamp just one Modbus Power Meter.
When trying this code all of my Export is showing as Imported electricity usage.

It might be worth having it as an optional setting, you don't want to force all users to use the new method if there calculations are correct.

fxstein:
I will update the logic to only apply to Gen4 inverters. That way the Gen3 logic stands as is.

PR #1083, opened 2024-10-07
Limit No Discharge/Battery Hold to 98% Soc to avoid PV from shutting …

fxstein:
The No Discharge option has an unintended side effect: When the SoC reaches 100% (and even when it later drops to 99%) an active No Discharge will make the inverter(s) stop producing solar power. Even on the sunniest day, PV will drop to zero and all load will come from the grid, even if PV would produce more than the houseload.
In order to avoid this situation this PR limits the No Discharge option to below 98%. That way, even when the battery reached full due to excess PV production, the system will continue to produce full PV power.


wills106:
@infradom Have you noticed this situation as well?

fxstein:
The issue has been discussed here: https://github.com/wills106/homeassistant-solax-modbus/discussions/833

This is very noticeable in parallel mode when individual inverters go to zero PV as they reach 100 SoC on a Sunny day. 

I have a working prototype for the improved parallel logic that I will submit once these underlying PRs of mine are merged. (Trying to avoid too many branches on my end)

PR #1089, opened 2024-10-12
Allow RTC sync offsets large enough for DST change

isorin:
With the upcoming DST change, increase the RTC sync offsets so they would allow to set the Solis RTC clock to the new time (-1h/+1h) in advance.

It is not possible to wait for the time change to sync the RTC clock as the inverter would start charging / discharging on an existing schedule ahead of time.


wills106:
Wont you have to sync it again once DST actually changes otherwise the Inverter will remain an hour out?

isorin:
The idea is to sync to the new time with a few hours in advance and then, after the DST change the periodic time checks / syncs will revert the offset to zero (or whatever previous value we were using) before syncing the clock again.On 13 Oct 2024 11:39, wills106 ***@***.***> wrote:
Wont you have to sync it again once DST actually changes otherwise the Inverter will remain an hour out?

PR #1091, opened 2024-10-13
Revert "Solax Gen4 meter direction and core meter direction fix"

wills106:
Reverts wills106/homeassistant-solax-modbus#1082

Messes up Gen3


fxstein:
@wills106 Can you share any specifics on what got messed up for Gen3 inverters?

I can either fix the issue for Gen3 or exclude Gen3 and below from the fix logic.

PR #1096, opened 2024-10-15
Growatt: Fix firmware control version format

TonyMathiesen:
This corrects the format to display the firmware control version: 
image

PR #1098, opened 2024-10-15
Growatt TL3-XH: Fix missing scaling in "Total Grid Power" and change to VA

TonyMathiesen:
Only for MOD TL3-XH (GEN4):
This fixes a missing scaling on "Total Grid Power" and changes unit to S32 as it can be negative on export.


TonyMathiesen:
Changed this to also set "total_grid_power" and "grid_power" as hidden by default, as it uses register 3023 which is output power. It has the exact same value. Did this to not do any name breaking, in case some are using it. 
Added a "total_grid_power_va" which is a value_function of the grid_power_l1+l2+l3.

wills106:
Same issue as #1096

PR #1099, opened 2024-10-15
Growatt TL3-XH: Scale battery voltage depending on battery system

TonyMathiesen:
For Growatt TL3-XH (GEN4): 
Depending on battery system, the scaling on battery voltage can be different. Default is set to 0.01 as per Growatt Modbus documentation. 
Scaling is determined by value_function_battery_voltage which uses BMS Monitoring Version (register 3096) to select if it should be 0.1 scaling. 
The APX HX system uses "ZECA" and we can add more when they are known.


wills106:
Same issue as #1096

PR #1118, opened 2024-10-29
Solax X3-MIC: Blacklist invalid registers

lcintrat:
New holding registers prevented the integration to update values.

Blocking registers are disabled using the blacklisting mechanism, however they may need to be disabled for all X3-MICs.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -6094,6 +6094,7 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
         scale = value_function_rtc,
         entity_registry_enabled_default = False,
         allowedtypes = MIC,
+        blacklist=('MC210T',),
         entity_category = EntityCategory.DIAGNOSTIC,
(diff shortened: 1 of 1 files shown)


wills106:
Are you sure you need to Blacklist all of them including RTC?

You are probably missing one of the newer registers, either because your specific model doesn't support one of the registers or your on an older firmware version.

https://github.com/wills106/homeassistant-solax-modbus/commit/cf216d70ac78d92a4f81a09fc06fb392d84d71b9

lcintrat:
The inverter firmware are ARM 1.07 and DSP 1.09.
I will update and give you a feedback.

Thank you!

lcintrat:
Updating firmware to ARM 1.17 and DSP 1.19 fixed the issue.

wills106:
Thanks for the update, looks like there is actually even newer Firmware's available than you upgraded to.
Or did SolaX recommend to use ARM 1.17 and DSP 1.19?

lcintrat:
You are right, there are newer firmware available. However, they come with a third one named "ARC". I did not want to try them for the moment.

Issue #1137, opened 2024-11-18
[Bug]: wrong parameter at state class


steps56:
### Describe the bug

Logger: homeassistant.components.sensor
Quelle: components/sensor/__init__.py:571
Integration: Sensor (Dokumentation, Probleme)
Erstmals aufgetreten: 11:33:44 (1 Vorkommnisse)
Zuletzt protokolliert: 11:33:44

Entity sensor.solax_remaining_battery_capacity (<class 'custom_components.solax_modbus.sensor.SolaXModbusSensor'>) is using state class 'measurement' which is impossible considering device class ('energy') it is using; expected None or one of 'total', 'total_increasing'; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues

### Integration Version

newest

### Homeassistant core version

newest

### Inverter brand

SolaX G3 Hyb

### Plugin used

plugin_solax.py

### Serial prefix

unnecessary here

### Inverter firmware versions

35 9 36

### Connection Method

SolaX Lan/Wlan Dongle

### Dongle firmware

3.0

### Detailed Error Log

__Logger: homeassistant.components.sensor
Quelle: components/sensor/__init__.py:571
Integration: Sensor (Dokumentation, Probleme)
Erstmals aufgetreten: 11:33:44 (1 Vorkommnisse)
Zuletzt protokolliert: 11:33:44

Entity sensor.solax_remaining_battery_capacity (<class 'custom_components.solax_modbus.sensor.SolaXModbusSensor'>) is using state class 'measurement' which is impossible considering device class ('energy') it is using; expected None or one of 'total', 'total_increasing'; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues

### Additional context

Hello,
got Error message shown above.
Workaround I used, was to set a # in front of the state class line at line 5247 at solax.py

The opposit entity chargable battery capacity hasn't any state class... so ... 🤷🏼‍♂️

May be, its an unnecessary line, that state class at remaining capacity entitiy 🤔

steps56:
Sorry, its at line 5242
state class again enabled for statistics, but changed:
ENERGY
to:
TOTAL_INCREASING

PR #1139, opened 2024-11-19
Remove Solax Remaining Battery Capacity state class

prehor:
The state class for the Remaining Battery Capacity should be removed, consistent with other battery capacity sensors using the ENERGY device class. The ENERGY device class cannot use the MEASUREMENT state class.

Fixies #1137


prehor:
May be better solution is change device class to ENERGY_STORAGE which could be used with MEASUREMENT state class.

steps56:
Hias I already told at the initial issue post:
The clean solution is:
state class change
from
ENERGY
to:
TOTAL

steps56:
> The state class for the Remaining Battery Capacity should be removed, consistent with other battery capacity sensors using the ENERGY device class. The ENERGY device class cannot use the MEASUREMENT state class.
> 
> Fixies #1137

Isn't any solution, as still throwing error/warning at system logs.
Clean errorfree solution please see my last post here at the end.

PR #1140, opened 2024-11-20
Add entity Main Breaker Current Limit

mikael-andersson:
NB Tested only on G4, can someone please look into G5?

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -804,6 +804,18 @@ NUMBER_TYPES = [
         icon = "mdi:battery-charging-high",
     ),
     SolaxModbusNumberEntityDescription(
+        name = "Main Breaker Current Limit",
+        key = "main_breaker_current_limit",
+        register = 0x71,
+        fmt = "i",
+        native_min_value = 10,
+        native_max_value = 250,
+        native_step = 1,
+        native_unit_of_measurement = UnitOfElectricCurrent.AMPERE,
+        device_class = NumberDeviceClass.CURRENT,
+        allowedtypes = AC | HYBRID | GEN4 | GEN5,
+    ),
+    SolaxModbusNumberEntityDescription(
         name = "Feedin On Power",
         key = "feedin_on_power",
         register = 0xB7,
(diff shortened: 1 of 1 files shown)


wills106:
I can't test on a G5, but according to the docs it's the same as the G4.

custom_components/solax_modbus/plugin_solax.py
```diff
@@ -803,6 +803,18 @@ def value_function_battery_voltage_cell_difference(initval, descr, datadict):
         allowedtypes = AC | HYBRID | GEN4 | GEN5,
         icon = "mdi:battery-charging-high",
     ),
+    SolaxModbusNumberEntityDescription(
+        name = "Main Breaker Current Limit",
+        key = "main_breaker_current_limit",
+        register = 0x71,
+        fmt = "i",
+        native_min_value = 10,
+        native_max_value = 250,
```
wills106, line 812: Does the Inverter allow you to set 250A? As the docs only mention 100A

custom_components/solax_modbus/plugin_solax.py
```diff
@@ -2842,6 +2854,13 @@ def value_function_battery_voltage_cell_difference(initval, descr, datadict):
         allowedtypes = HYBRID | GEN4 | GEN5,
         internal = True,
     ),
+    SolaXModbusSensorEntityDescription(
+        name = "Main Breaker Current Limit",
```
wills106, line 2858: name = 

Isn't actually required on Internal entities, only the key is required.

custom_components/solax_modbus/plugin_solax.py
mikael-andersson: My pull request was intended for X3, but does my pull request apply to X1?

From Solax Modbus spec (V3.24):

image

custom_components/solax_modbus/plugin_solax.py
mikael-andersson: Do you want me to remove it then so that it is consistent with other Internal entities?

custom_components/solax_modbus/plugin_solax.py
wills106: I was actually looking at 0xD7, where it only mentions 32-100A. It's common for there to be a difference within the docs...
I would keep it at 10-250, people can then set it to what suits.

Or we could have an X1 0x71 with 10-100 and an X3 0x71 with 10-250?

custom_components/solax_modbus/plugin_solax.py
mikael-andersson: OK, will do. :)

Should I also add X1 and X3 in the list of allowedtypes?

Issue #1142, opened 2024-11-20
Solis Firmware 4B00


fboundy:
### Describe the bug

Solis firmware 4B00 introduces a completely new set of registers for time based control. It also adds a target SOC (and a voltage though it’s unclear what this is used for) to each slot.

The old registers are no longer read so the registers used need to depend on the FW version.

I’ve polled the API and Modbus and come up with this list: https://github.com/fboundy/pv_opt/issues/305#issuecomment-2489570654

### Integration Version

2024.11.2

### Homeassistant core version

2024.11.2

### Inverter brand

Solis

### Plugin used

plugin_solis.py

### Serial prefix

NA

### Connection Method

Waveshare

alienatedsec:
@wills106 Is it possible to introduce a switch that uses a new set of registers if, for example, another register represents a specific value? 

The input register hex value would have part of the firmware version.

The below `33002` is the HMI version - `4A` in my case (currently unaffected) 
![image](https://github.com/user-attachments/assets/46392c82-c65e-4b5a-b11d-0df314f998ef)

This is usually followed with `00`, and combines with the register `33001` the DSP version - `4C` in my case - and `33069` both unrelated at this moment - or could be if DSP has any specific function defined.
![image](https://github.com/user-attachments/assets/2152a162-a24b-46d3-a636-ed8cc6c2a667)

FW version `33002` + `00` + `33001` + HMI subversion `33069`
![image](https://github.com/user-attachments/assets/540e64da-4572-45b9-b595-66c4d4378675)

So, the final logic should be that if the register `33002` is `0x004b` or higher, than the new set of registers should be used.

@fboundy @infradom @jmccrohan FYI

fboundy:
@alienatedsec - I think it might be easier just to use a separate plugin then all the logic will be very similar but a separate set of registers etc

fboundy:
Should be addressed by #1143

alienatedsec:
@fboundy 
I got the new firmware files from Solis, but I am delaying to load them. I also believe it's a newer version than 4B00, so I guess I will have to decide at some point.

fboundy:
I was told yesterday that 4B00 is still the latest

wills106:
If the Firmware was read at the same time as the Serial Numer then it should be possible to use the unused GEN block to split it up?

If lower than y = GEN if above x = GEN2?
Depends how the version numbering works as a whole really on Solis.

fboundy:
[...]

Issue #948, opened 2024-07-03
[Bug]: HA 2024.7.x Detected code that calls async_forward_entry_setup


wills106:
### Describe the bug

Detected code that calls async_forward_entry_setup for integration solax_modbus with title: SolaX and entry_id: xyz123, during setup without awaiting async_forward_entry_setup, which can cause the setup lock to be released before the setup is done. This will stop working in Home Assistant 2025.1. Please report this issue.

### Integration Version

2024.07.0b1

### Homeassistant core version

2024.7.0b10

### Inverter brand

SolaX

### Plugin used

plugin_solax.py

### Serial prefix

H1E5

### Inverter firmware versions

ARM 3.28, DSP 3.32

### Connection Method

Gen3 Ethernet

### Dongle firmware

N/A

### Detailed Error Log

`2024-07-03 09:43:38.436 WARNING (MainThread) [homeassistant.helpers.frame] Detected that custom integration 'solax_modbus' calls async_forward_entry_setup for integration, solax_modbus with title: SolaX and entry_id: xyz123, which is deprecated and will stop working in Home Assistant 2025.6, await async_forward_entry_setups instead at custom_components/solax_modbus/__init__.py, line 339: self._hass.async_create_task(, please create a bug report at https://github.com/wills106/homsassistant-solax-modbus/issues`

wills106:
There is no need for additional reports of this, or for people to also confirm they get this. Everyone on HA 2024.7 will get this.
It's the same with #920 on HA 2024.6, everyone will get the error until it's fixed.

It's just a single warning at the start of HA per Inverter configured.

wills106:
Just released [2024.07.0b2](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.07.0b2)

wills106:
Reopened as not fully resolved.

litinoveweedle:
> It's the same with #920 on HA 2024.6, everyone will get the error until it's fixed.

Hi, not it is not the same, this one is a new warning, but for similar topic. I can prepare fix for this one as well.

wills106:
> Hi, not it is not the same, this one is a new warning

I know it's different, I just meant everyone will get this error like everyone got the last error. So no need for people to keep reporting it.

PatrikTrestik:
> > It's the same with #920 on HA 2024.6, everyone will get the error until it's fixed.
> 
[...]

PR #1143, opened 2024-11-21
Solis HMI Firmware FB00

fboundy:
Add support for Solis HMI firmware which:

- Simplifies the Energy Code Switch options 
- Has 6 charge and discharge slots rather than 3
- Has separate Current, Target SOC and Voltage for each Charge/Discharge slot

Note that on the inverter and in the SolisCloud API interface each slot also has an On/Off switch. I have not found a ModBus register for this so this will need to be set "ON" once on the inverter or through SolisCloud. It is also not clear what the Voltage target on each slot does.


ajan1:
I have charge/discharge slot on/off switches in address 43707. Bits 0-5 charge 1-6 and bits 6-11 for discharge 1-6

I tried this pull. I didn't get how to enable/disable my time slots?

fboundy:
I hadn’t managed to find those registers. I had to enable manually on the inverter. I’ll update and push a commit.

fboundy:
No updated to add switches. This is quite a big PR as the SWITCHES platform wasn't previously used. I have added SWITCH_TYPES to the solis_fb00 plugin and empty lists to all the other plugins.

wills106:
Need to relook into https://github.com/wills106/homeassistant-solax-modbus/issues/948#issuecomment-2217068383 to prevent all of these huge file diff's due to people having different formatting settings. It's making it impossible to know what's actually changed. Same with #1173 

@litinoveweedle @fboundy @PatrikTrestik any idea on the best approach / guide on how to set something up?
Last time I looked at VS Code plugin "Black Formatter" it made a right mess locally.

fboundy:
I use Black in VS Code with --linelength=119

fboundy:
On a separate note, I'm note sure whether it is best to keep this as a separate plugin or to try to merge it with `plugin_solis.py`. Register 33002 gives the relevant firmware version. If it is >= `0x4B` (75) then this new plugin should be used, otherwise the old one is correct. Just not quite sure how to implement this switch.

fboundy:
> Need to relook into [#948 (comment)](https://github.com/wills106/homeassistant-solax-modbus/issues/948#issuecomment-2217068383) to prevent all of these huge file diff's due to people having different formatting settings. It's making it impossible to know what's actually changed. Same with #1173
> 
> @litinoveweedle @fboundy @PatrikTrestik any idea on the best approach / guide on how to set something up? Last time I looked at VS Code plugin "Black Formatter" it made a right mess locally.

ChatGPT says it's easy. I will have a try on one of my repos.

To deploy a GitHub CI workflow that runs the Black formatter on all commits, you can use GitHub Actions. Here's how to set it up:
Steps:

Create a .github/workflows directory in your repository if it doesn’t already exist.

Add a new YAML file for the workflow:
  Name it something like black-check.yml.

 Write the workflow:
   Below is an example workflow that uses Black to check Python files in your repository:

```
name: Black Formatter Check

on:
  push:
    branches:
      - main  # Specify the branches you want to check (e.g., main)
  pull_request:
    branches:
      - main  # Run checks for pull requests targeting main

jobs:
  black:
    name: Run Black Formatter
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: '3.x'

      - name: Install Black
        run: |
          python -m pip install --upgrade pip
          pip install black

      - name: Run Black Formatter
        run: black --check --diff .


```

wills106:
I was thinking of adding GEN to older Firmware's and then adding in GEN2 for the newer.
https://github.com/wills106/homeassistant-solax-modbus/issues/1142#issuecomment-2506606021
Then have the relevant tags where they spilt. The GEN, GEN2, GEN3 etc used used quite a bit on SolaX as they share base registers and then split. But it's defined by the serial and not an extra Register ie 33002

I have just been too busy over the past few weeks to look at the various PR's on here / do any development.

litinoveweedle:
Hello,

I am using for my project automatic formatters hooked to the CI (GitHub actions). It works as supposed -  runs on PR merge/push so code is always reformatted before merging - therefore you avoid this commits mess. I can try to help, although I am not an expert and my actions are mostly copy paste of existing examples. :-)

wills106:
I have added in pyproject.toml with black and line-length 119
Is it possible to force VS Code to have black as a dependency in the pyproject.toml file? I'm quite new to this.

Also added in workflow for black as well with the same line-length

fboundy:
I’ve got it set up now on my pvopt repo if you have a look in .github/workflows/black.yaml

PR #1154, opened 2024-12-02
Fix inverter and measured power for older SolaX MIC inverters

Kars-de-Jong:
**Correct inverter and measurement entities for older SolaX MIC inverters**

- The Measured Power entity for Gen1 and Gen2 MIC inverters is actually the Inverter Power.
- The Inverter Power entity for Gen1 and Gen2 MIC inverters is actually the Measured Power
- The Measured Power Lx entity for Gen1 X3-MIC inverters is actually the Lx Inverter Power.
 
**Enable house load entity for MIC inverters**

House load can be calculated for MIC inverters that have a meter connected.

**Enable Inverter Power Lx entities for X3 Gen2**

---

Tested with a Gen1 X3-MIC with an emulated SDM230 meter connected.

Please note that this change breaks the history of the existing entities. Not sure if an upgrade script is possible / needed.


mrichtarsky:
Thanks @Kars-de-Jong for the fix, I will test this locally

mrichtarsky:
@Kars-de-Jong Looks good to me. Except for Measured Power L1, which still does not make sense to me (i.e. it stays at 0 even when charging with 11kW). But that might be a problem of the inverter.

Kars-de-Jong:
@mrichtarsky I don't know the Modbus address of `Measured Power L1` (which would be the power measured by an external kWh meter). After my change, the `Measured Power Lx` entities no longer exist, you should have `Inverter Power Lx` entities instead.

Kars-de-Jong:
@mrichtarsky Done. Could you please test?

The fact that "Measured Power L1" is 0 in your case might indicate a problem with your kWh meter. Maybe the L1 CT is not connected properly?

wills106:
Thanks for working on this. Once mrichtarsky has confirmed the latest changes are ok I'll merge it.

mrichtarsky:
Thanks @Kars-de-Jong! 
Currently it looks like this

Screenshot_20241211_084343_Home Assistant

Inverter Power L1-3 are zero although current and voltage are not, perhaps the value must be calculated and is not available directly?

Kars-de-Jong:
@mrichtarsky Thanks for testing again.

I see now in your original issue report that you already tried reading 0x410, 0x411 and 0x412 and that they always return 0 in your case.
That's quite odd, those addresses match with the documentation of X3-MIC Pro-G2 V2.7.
In that case I'm at a loss.

All I can suggest is for you to try and read all input registers using pymodbus or for instance `mbpoll` and see if we can find values on addresses that match with the individual phase powers.

PR #1207, opened 2025-01-08
Fix storage modes (and a register) for Solis-FB00

isorin:
I've upgraded my Solis inverter to the latest firmware and started using FB00 implementation.

The storage control switch is the 1st thing I noticed.
1. Incorrect settings for some modes (so trying to make it more consistent with the pre FB00 implementation)
2. The two sets need to match, otherwise setting a mode will result in an empty value.

Also fixed register for "timed_charge_start_hours_6".

One other thing which needs to be fixed (I will send another PR if I find where it is) is the missing prefix for switch entities
e.g. "Timed Charge Slot X Enable" is not prefixed (just the name, the entity_ids are correct)
it should be "Solis Timed Charge Slot 1 Enable" (or whatever the user chooses as prefix in the config)


wills106:
@fboundy are you able to check this change as you did the initial support for the newer Firmware version.

Just want to check that different Inverter's aren't using a different register for slot 6.

fboundy (approved):
This looks correct. Must have been a typo in the original.

fboundy (approved):
Yes, this all look OK. Personally I only use 33 and 49 anyway.

PR #1210, opened 2025-01-12
Growatt serial-no buffer overrun + manifest pymodbus

PSMGoossens:
- Solved issue where the serial number read count is to large for the actual registers in function `async_read_serialnr`.
- Updated pymodbus to align with home assistants 25.1 version (thanks @Luflosi for the comment on the PR)
- Updated version so version could be released if needed.
- Updated functions `value_function_rtc_ymd` and `value_function_rtc` to handle years with 4 digits as well

### DIFF
--- a/custom_components/solax_modbus/const.py
+++ b/custom_components/solax_modbus/const.py
@@ -407,7 +407,7 @@ def value_function_rtc(initval, descr, datadict):
             rtc_months,
             rtc_years,
         ) = initval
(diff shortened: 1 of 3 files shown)


Luflosi:
My comment can be seen here: https://github.com/wills106/homeassistant-solax-modbus/commit/00a319cd3daaea7aae6078076172bfaf8805e05f.

Issue #1285, opened 2025-03-15
[FR]: Solax Gen2 EVC X1|X3-HAC-* support


jefft4:
### Description

Add support for the Solax Gen2 EVC range.

### Suggested Solution

I have a Gen2 X1-HAC-7 and I've got the Solax modbus documentation for the Gen2. Looks mostly the same sa Gen1, some extra values and possibly a couple of register changes. 

I'll clone the repo and generate a pull request for the changes I can identify.

Will need someone with a X3-HAC-* to identify the serial number range for those - add later.

jefft4:
https://github.com/wills106/homeassistant-solax-modbus/pull/1286

empx92:
Have a X3-HAC-11 and could help, but at the moment it is hard to find time...until end of march

PR #1286, opened 2025-03-15
Update plugin_solax_ev_charger.py for Gen2 EVC support (#1285)

jefft4:
Added serial number recognition for Gen2 EVC and additional registers/values.
- 0x610 additional options for GEN2
- 0x61C EVSE Scene GEN2
- 0x668 max charge current GEN2
- 0x669 EVSE mode GEN2
- 0x10 is total charge energy (cumulative) GEN2; made the other register GEN1 only
- 0x1D additional options for GEN2
- 0x25 firmware version GEN2
- 0x26 network state GEN2
- 0x27 wi-fi RSSI GEN2
- 0x2B charging duration GEN2
- 0x2D lock status GEN2
- 0x2E main breaker state GEN2
- 0x2F delay state GEN2
- 0x30 charge ban state GEN2

I've not yet added the new time delay/charge schedule registers as I'm not sure how best to implement time types (Solax has separate registers for h,m,s which would be messy):
- RTC setting registers 0x61D-623
- Time delay registers 0x634-639
- Smart boost energy 0x63A (goes with 634-639)
- Charge start registers 0x31-36
- Adjust current 0x628, as I don't know what it does!

Made charge phase register X3 only, doesn't apply to X1 single-phase.


wills106:
> I've not yet added the new time delay/charge schedule registers as I'm not sure how best to implement time types (Solax has separate registers for h,m,s which would be messy):

So far I have used selects for most of the times and calculated them internally.

I would love to implement proper time selections. But as you have found the implementation on the SolaX side is quite messy.
I think with the different SolaX setups we would need 3-4 different methods to fully implement time selection.

So use high and low 8bit's of a 16bit register for time.
Some use individual 16bit registers for hours and mins.
Some Inverters reverse that and swap hours and mins around...

jefft4:
@wills106 can you help here?

I'm getting two problems logged in testing this, that I can't see the cause of:
```
evc: read failed at 0x30: ban_state
  File "/config/custom_components/solax_modbus/__init__.py", line 706, in treat_address
    val = decoder.decode_16bit_uint()
  File "/config/custom_components/solax_modbus/payload.py", line 370, in decode_16bit_uint
    return unpack(fstring, handle)[0]
           ~~~~~~^^^^^^^^^^^^^^^^^
struct.error: unpack requires a buffer of 2 bytes
```
That register is a uint16, can't see any difference between it and others.

and 
```
Source: custom_components/solax_modbus/sensor.py:263
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 17:37:31 (3 occurrences)
Last logged: 17:51:16

input register already declared: 0x1d run_mode
```
Reverting the code to the original, all-generations run_mode definition doesn't get rid of the error.

Any ideas? Have I missed something in the way I've defined those?  Thanks!

jefft4:
I'm not sure "allowedtypes" is working in the SelectEntity descriptions. The values I see for start charge mode are from the first of two definitions, not the second (GEN2) definition. That might explain the complaint in the log about run_mode too.

jefft4:
> I'm not sure "allowedtypes" is working in the SelectEntity descriptions. The values I see for start charge mode are from the first of two definitions, not the second (GEN2) definition. That might explain the complaint in the log about run_mode too.

Found it :-) matchInverterWithMask in the evc plugin wasn't testing GEN and also, POW and GEN were using the same hex digit, so POW7 == GEN1, making any POW7 device also a GEN1 device. Moved POW up a digit to be 0x0010 etc and added the genmatch test. Now getting the right definitions.

Code pushed.

empx92:
Does the HAC need to be unbound from Solax Account to work via  Modbus TCP?

dskvml:
i am not sure how this would work, sorry if this is the wrong place. I have a 5070 Serial number and the device is not recognized by the adapter automatically. Do i have to connect it manually (which port to use?) Or is my serial number to be added?

PR #1288, opened 2025-03-16
Change pymodbus requirement to soft pin

davidrapan:
Hi, please, hard pinning of python requirements is really not healthy for the ecosystem (see https://github.com/davidrapan/ha-solarman/issues/456).


wills106:
It's sometimes required. Especially when the upstream Library introduces a breaking change. Ie 3.7.x to 3.8.x

davidrapan:
Did you read the threads? Hard pinning in custom component can cause breaking of other components and it took us quite a while to figure out that yours was causing the issue...

Therefore, I would like to suggest checking the package version programmatically.

PR #1292, opened 2025-03-17
Sofar: Merge duplicate PV Power Total

cschlipf:
PV Power Total was added twice, just with different allowedtypes definitions. Merged into one entity.


wills106:
I don't think this is a duplicate. One is for PV only and the other for Hybrid. They have different scaling parameters.

cschlipf:
Ouch! You are right. Missed the scaling factor. Thank you very much for watching!

PR #1296, opened 2025-03-19
Sofar: Remove LSE-3 recommendation from FAQ

cschlipf:
We can no longer recommend the LSE-3, but the FAQ still contained a recommendation to do so. Removed the recommendation and suggest using a RS485 TCP adaptor instead.


robertugo2:
Agree. I bought LSE-3 especially for use with this integration and it sucks :(

cschlipf:
I am very sorry for this previous recommendation. It worked fairly well in the past, however quite some changes in pymodus and from SolarMan changed that.

robertugo2:
> I am very sorry for this previous recommendation. It worked fairly well in the past, however quite some changes in pymodus and from SolarMan changed that.

You don't need to sorry! It is not your fault :) We are all in open source community and we are doing everything for community and things like that, especially when we have no power to prevent such things, will happen. It's normal.

It would be just good to remove this recommendation :)

PR #1297, opened 2025-03-19
Update for Sofar FAQ about LSW-3 Wifi Stick Logger

davidrapan:
Hi, I was bold enough to include link to my repo but feel free to remove it if it's a problem. 😉 I'm more concerned about the factual information about this logger and it's 8899 port because some users of my integration used this docs (and its not really correct info) as a proof of why something doesn't work, etc.

Have fun, thanks.


cschlipf:
Hm, ideally you should also list pro and cons then and not market it as an equivalent alternative. Nothing against your integration. If reading values is good enough it's a fair alternative, but when it comes to actively controlling the inverter it lacks quite some features.

davidrapan:
> If reading values is good enough it's a fair alternative, but when it comes to actively controlling the inverter it lacks quite some features.

Mentioned integration supports full control of the inverters so there are no pros and cons.

> And I think the doc is not wrong: ModBus TCP connection is indeed timing out on port 8899

I'm not saying it's wrong, but rather adding more specific information about the port.

> The question would rather be why are users of your integration using the documentation of this integration?

Google. Not enough information (in general) results in users connecting dots which are not really relevant.

davidrapan:
Fixed conflicts.

wills106:
Closing down as not relevant to this Integration.

PR #1329, opened 2025-04-19
Update compatible-inverters.md

JohnMacrae:
After testing with X1 Mini G4


JohnMacrae:
Thanks for your work :)

PR #1333, opened 2025-04-21
Ignore faulty Modbus reading

PatrikTrestik:
This is minimal commit to skip faulty Modbus readings causing spikes in data.

## Description of problem
There is bug in PyModbus (<3.8.6).
Modbus has no option to physically cancel operation. When Read request times out (logic cancel in code), but physical device responds later, there is "orphan" packet which can in some situation be falsely read by next transaction. 
Pymodbus<3.8.6 does not check TransactionID so is vulnerable to this error.

PyModbus is core HA library so it is not possible to upgrade it only for one integration. We have to wait till HA upgrades.

## Solution
I picked up exact solution used in PyModbus 3.8.6. On data receive it should check Request.TransactionId==Response.TransactionId.
Have to use more low lever reading function to have access to response properties.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -14,6 +14,7 @@ from weakref import ref as WeakRef
 
 from pymodbus.client import AsyncModbusSerialClient, AsyncModbusTcpClient
 from pymodbus.exceptions import ModbusException
+from pymodbus.pdu import register_message
 
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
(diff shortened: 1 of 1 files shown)


PatrikTrestik:
@wills106 please consider merging this.
Running already five days in row without a spike on my system.

PatrikTrestik:
@wills106 you put wrong credit in release description 😢
No harm done 😁

wills106:
Sorry, should be corrected now.
Trying to release it in a rush before going out 🤦‍♂️

PatrikTrestik:
No problem 😁
On first I've think that it is different fix and wanted to check how it was done. 
Thanks for correction. 

I've done bunch of improvements to other parts of code. I'll post other PR when storm with spikes calms down. I want to test it well and not cause confusion.

Also I have SQL to fix spikes in history. I'll write wiki page and let you know.

wills106:
@PatrikTrestik How are you connecting to your Inverter? PocketWiFi or Waveshare RS485 - Ethernet connection?
Just trying to understand why some people are having issues with the latest version.

I know some Inverters are using scan_group= so could be polling at different Intervals and causing issues with different transaction id's
But on SolaX Inverters we aren't using that and I have noticed some SolaX users are getting `Modbus: ERROR: expected transaction 126 but got 0, IGNORING` but that perhaps isn't preventing the Serial number from being read.

But at the same time it seems other users aren't getting these errors. My install is fine although I'm using direct Modbus over TCP and not an adaptor.

TonyMathiesen:
Hi, just if it helps narrowing down the issue. 
I use Growatt plugin with Waveshare RS485 Ethernet connection. 
I get the same errors and the values cannot be read with version 2025.04.2:
[custom_components.solax_modbus] Modbus: ERROR: expected transaction 1 but got 0, IGNORING.
[custom_components.solax_modbus.plugin_growatt] Growatt: attempt to read firmware failed at 0x9

PatrikTrestik:
@wills106 I see that problem is when TransactionID=0. Other cases are real bug we want to filter.
It can be caused by:
- serial connection: There is no TransactionID in serial protocol. This is bug in my code and I'll fix it today or tomorrow.
- using Modbus RTU over TCP or other form of serial tunel: I think there is no reason for using this setup. I'll try to cover this in my fix.
- incorrect setup of RTU<>TCP converter (Waveshare atc.) or bad Modbus TCP implementation (cheap converters). I have no idea how to fix this. When pymodbus is upgraded these setups will stop working. :(

I'll go through TCP and RTU Modbus specification and I'll try my best to produce fix which reacts only in case of realy invalid transaction.

@TonyMathiasen thanks for posting. It shows that "0" transaction ID is incorrectly ignored.
But correctly setup Waveshare RS485 converter should not produce TransactionID=0.
Are you using some son standard setup? Can you please share.

TonyMathiesen:
Hi @PatrikTrestik 
What do you need of information? I believe I run a standard setup, it is configured as "Modbus RTU over TCP".. Is this the reason?

PatrikTrestik:
> Hi @PatrikTrestik What do you need of information? I believe I run a standard setup, it is configured as "Modbus RTU over TCP".. Is this the reason? Edit: It's a Waveshare RS232/485/422 TO POE ETH (B)

Thanks for info. It is reason for incorrect behaviour.
1) Clearly shows that I have bug in code. Now I know how to fix.
2) You should run it in Modbus TCP mode. It is protocol for TCP network. 
RTU over TCP should be used only in rare cases where there is no other option.

@wills106 Not in relation to my fix, but do we have any reason for supporting RTU over TCP? Even cheap RS485 converters support Modbus TCP. RTU should be only for pure serial.

wills106:
One of the adaptors in the Wiki uses RTU over TCP, will check later. Someone else added in the support for their adaptor.

But there are chances people have their adaptors configured wrong.

TonyMathiesen:
@PatrikTrestik Thanks for clarifying. I will get the config changed later today 😉

TonyMathiesen:
FYI it worked changing to Modbus TCP and running 2025.04.2

PR #1347, opened 2025-04-29
Fix Modbus RTU does not have transactionID

PatrikTrestik:
Fixing bug in previous PR.
Only Modbus TCP uses TransactionID. It should not be used for RTU or Ascii.

Actually bug where overlapping transactions results in incorrect data is only possible on Modbus TCP.

For non-TCP framer, TransactionID in response is always 0. Added condition to filter RTU out.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -576,10 +576,10 @@ class SolaXModbusHub:
             try:
                 pdu_request = register_message.ReadInputRegistersRequest(address=address, count=count, dev_id=unit)
                 resp = await self._client.execute(False, pdu_request)
-                if resp.dev_id != pdu_request.dev_id:
+                if resp.transaction_id!=0 and resp.dev_id != pdu_request.dev_id:
                     _LOGGER.warning("Modbus: ERROR: expected id %s but got %s, IGNORING.", pdu_request.dev_id, resp.dev_id)
                     return None
-                if pdu_request.transaction_id != resp.transaction_id:
+                if resp.transaction_id!=0 and pdu_request.transaction_id != resp.transaction_id:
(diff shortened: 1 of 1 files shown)


PatrikTrestik:
@wills106 sorry for complication your work with this.
I was too focused on solving spikes that I forget that there are also RTU connections.

TonyMathiesen:
@PatrikTrestik Is this fix also needed in, for serial connections:
https://github.com/wills106/homeassistant-solax-modbus/blob/0ffe6cb36147a0d099117d8dec46e1941dc0d534/custom_components/solax_modbus/__init__.py#L550-L569

I do not have an issue with my Modbus TCP connection, but I can see there are issues raised.

PatrikTrestik:
> @PatrikTrestik Is this fix also needed in, for serial connections:
> 
> https://github.com/wills106/homeassistant-solax-modbus/blob/0ffe6cb36147a0d099117d8dec46e1941dc0d534/custom_components/solax_modbus/__init__.py#L550-L569
> 
> I do not have an issue with my Modbus TCP connection, but I can see there are issues raised.

It was my mistake to include serial connections. There is no TransactionID for serial. Fixed in my last PR #1347 which was merged today.

TonyMathiesen:
@PatrikTrestik 
In this PR you have added some transaction id logic to "async_read_input_registers", my question was if it also needed to be added to "async_read_holding_registers"? It seems it might be missing the changed logic. 😊

PatrikTrestik:
> @PatrikTrestik In this PR you have added some transaction id logic to "async_read_input_registers", my question was if it also needed to be added to "async_read_holding_registers"? It seems it might be missing the changed logic. 😊

😭 You are right. I was too fast with PR. I hate if someone does this at my workplace and than I do same in hobby project. 👎 

@wills106 please wait with version. I have to fix it.
I edited only part of code in this PR. Thanks to @TonyMathiesen  that he spotted it.

PR #1351, opened 2025-04-30
Fix No transactionID for holding registers

PatrikTrestik:
In last fix I forget to do it for holding registers


PatrikTrestik:
@wills106 this should help all serial errors. Not checking transaction ID for Input and Holding register read.

PR #1356, opened 2025-05-03
Added special settings for solis

nielskool:
I have a dynamic contract and for that I want to be able to "disable" the PV power generation of the inverter when the grid price is negative.

By adding the the special settings register from the Solis I am able to do so.

Register 43249:
Bit0, MPPT parallel function
Bit1, IgFollow 
Bit2, Relay protection 
Bit3, I-leak protection 
Bit4, PV iso Protection 
Bit5, Grid-Interference protection 
Bit6, The DC component of the grid current protection switch 
Bit7, Const Voltage Mode Enable
Bit8-11 Grid-filter level 0-7 (not added)
Bit12-15 Afci sensitivity 0-7 (not added)

By default the const voltage is at 600V, for me that is not optimal for my MPPTs, essentially turning off the PV generation.
This is the only setting I could find to disable PV generation and keep the battery&grid working.

### DIFF
--- a/custom_components/solax_modbus/plugin_solis.py
+++ b/custom_components/solax_modbus/plugin_solis.py
@@ -108,6 +108,13 @@ class SolisModbusSensorEntityDescription(BaseModbusSensorEntityDescription):
 
 # ====================================== Computed value functions  =================================================
 
+# This value function converts the bits to the number
+def mutate_bit_in_register(bit: int, state: int, descr: str, datadict: dict):
+    value = datadict.get(descr, 0)
+    _LOGGER.debug(f">>> Old value of {descr}: {value}")
+    new_value = (value & ~(1 << bit)) | (state << bit)
+    return new_value
+
 
 def value_function_timingmode(initval, descr, datadict):
     return [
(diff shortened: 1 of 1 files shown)


janusr:
Hi Nielskool
I have same issue, personally I use the "Grid Feed in Limit" setting - not sure if that is an option for your Solis Inverter? I have a Solis S6-EH3P10K-H.
/Janus

nielskool:
Hi Janus,

if you refer to the backflow limit, then that is indeed something I also turn on. But when the price is negative, I don't want to only not export to grid, but i also want to consume from grid and turn of the pv. By enabling the constant voltage I am able to semi disable the pv generation and force the consumption from the grid. Or am I missing something from your story? To me it seems that the solution you give only limits the export to the grid. But if there is a better solution than I have now, I would like to know :)

/Niels

janusr:
My purchase price is never (almost never) negative, as distribution, taxes and VAT is added to the price when purchasing. I will always use as much of my own PV production as I can, but only sell when profitable.

nielskool:
Yea okay makes sense. It is indeed for me specifically when the prices are negative(including taxes). Last month happened a few times.

PR #1359, opened 2025-05-06
When pv_power values are none, do not calculate the value

shaarkys:
This prevents drop-outs in PV Total, tested for 5 days


infradom:
I understand this makes the function more general, but it causes some problems as sometimes pv_power_total is added again, so it doubles every polling cycle(after a recent framework change)

Furthermore your function scans the entire datadict every polling cycle (hundreds of entities). 
For these reasons, I propose to post a PR with a different solution like :

``` 
MAX_PVSTRINGS = 10
def value_function_pv_power_total(initval, descr, datadict): 
    # changed: for performance reasons, we should not iterate over the entire datadict every polling cyle (contains hundreds ...)
    total = 0
    i = 1
    while i <= MAX_PVSTRINGS:
        v = datadict.get(f"pv_power_{i}", None)
        if v is None:
            break
        else: total += v 
        i += 1
    return total

```

PR #1373, opened 2025-05-13
added Solax Remote Control registers for direct writing and Registers A0-A7 for extended remote control

ggg1psyyy:
I have added and tested the registers A0-A7 from this documentation: https://kb.solaxpower.com/solution/detail/2c9fa4148ecd09eb018edf67a87b01d2
Mode 8 & 9 are very interesting for negative import prices to disable PV while using maximum grid import to fill the battery.
As they write there "the rest of the irrelevant registers need to be filled in as “0” as well.", I also added all existing Remote Control registers (7C-8A) for direct access so I can set them to 0 via automation.

I only tested these registers (A0-A7) on my Gen4 X3 Hybrid with Firmware: DSP v1.50 ARM v1.47, so I cannot say if they are available in other Gens or firmware versions, as they are not included in the other Modbus documentation I have.

If you don't want to create confusion with the multiple Remote Control Settings, I understand, but to disable PV input is very important for flexible tariffs, so perhaps it's enough to only keep the new registers.

Also there is register A6, which is used for two different purposes depending on the selected mode, I don't know if there is a better solution than mine to just add it two times.


johny-mnemonic:
@ggg1psyyy would you have some example how to use the Target SOC mode (9)?
I am trying to either charge battery from grid or discharge it to grid, stopping at target SOC, but seems like it does something strange for short time and than returns to default behavior.
I am most probably doing something wrong.
For example I have no clue which registers should be set to "0" and whether it is right to use the "remotecontrol_trigger" button, to trigger these modes the same way as the ordinary "Remotecontrol Power Control" modes.

Thanks for adding these new modes. They really look promising.

ggg1psyyy:
I'm using mode 9 in my NetDeamon app to automate my PV, you can see the example here: https://github.com/ggg1psyyy/NetDeamonApps/blob/master/HA/Mode.yaml
For simply (dis)charging the battery you don't need the remote commands, simply switch the inverter to the corresponding mode with an automation. It's only necessary if you want to disable or limit the PV input. You can see examples for this also in the linked yaml.
Concerning the setting to "0", I'm not 100% sure if it is necessary for modes other than 1-3 but I do it in any case. 
The trigger button you only need if you're *not* using the "direct" registers, as this is a built-in function of the addon to "circumvent" the timeout of the remote control in the inverter. If you use mode 8 or 9 you can't and mustn't use the trigger!

johny-mnemonic:
@ggg1psyyy Thanks!

Seems like my mistake was setting `select.solax_remote_control_power_control_mode_direct` instead of `select.solax_modbus_power_control_direct`.

But even when I use the correct one, it still does weird stuff. Will need to play with it more.

For example I now tried this:
```
number.solax_remotecontrol_target_soc_9_direct => 40
number.solax_remotecontrol_pv_power_limit_direct => 15000
number.solax_remotecontrol_push_mode_power_8_9_direct => 2500
select.solax_remote_control_power_control_mode_direct => Individual Setting – Target SOC Mode
```
I had PV production of 8kW and SOC 33% and all it had done was to completely stop PV production and battery charging/discharging and took house load from the grid😕
I expected it to limit the charging of the battery to 2500W and then stop battery charging at 40% SOC while using PV to charge the battery and satisfy home load🤷‍♂️

infradom:
I was the original author of the remote_control_recompute loop mechanism, that recomputes the power needs every x seconds. I have been inactive last year, and am glad to see Solax has added new parameters like PVPowerLimit. Thanks for implementing this PR.
I especially want to use this in negative price scenarios. In my case, there are 2 different negative price scenarios:

- injection prices are negative
- injection and consumption prices are negative

The latter case is simple, since we can just limit the PV Power to zero. In the first case (only injection prices are negative), we may need a new recompute loop that computes the needed PV power from the house load and desired battery charge speed. I intend to implement this in the coming days/weeks but would like to hear other opinions first.

ggg1psyyy:
@infradom if you want to incorporate the new registers into the recompute loop this would be nice as the current split between direct register access and your mechanism seems to be confusing, but afaics at least for mode 8&9 it's not necessary as these have defined stopping points. I was thinking about incorporating the new modes into the loop, just to set all other registers to 0, but eventually opted against it, as they are in a different range of registers and I didn't want to change your logic.

For your 2 price scenarios:
1. I just set the export limit to 0W
2. I use mode 8/9 to limit the PV power to 0

johny-mnemonic:
> 2. I use mode 8/9 to limit the PV power to 0

That might explain why it seems to be working for you as disabling both battery and PV (to get the power from the grid) is the only thing I managed to achieve using these modes😞

I am searching for a way to stop charging or discharging at specified SOC level when either charging from the grid/PV or when selling the surplus from battery to the grid.
Would you recommend what I might be missing? Did you try any such scenario?

I have checked the documentation again and I noticed two really weird conditions for the Mode 9:
```
No limit in Export Control / Power Limit / Main Breaker Limit
A forced charging period has been set, during which the inverter is allowed to charge from the grid.
```
Maybe these are the reason why it doesn't do anything for me as I definitely have the `export_control_user_limit` set and I have no clue what the `forced charging period` is good for when you use the modbus powercontrol to control the inverter😕

infradom:
@ggg1psyyy Thanks for your feedback and your work on this, I was thinking more or less to introduce a separate recompute loop for these modes (8 and 9) . Maybe I am missing something: 

- you say mode 8 and 9 dont need a loop because they have defined stopping points

In my opinion, we need to recompute the targets (PV or battery charge) every second in order to match the variations in the house load. So for me, automatic recomputation within the loop makes sense, unless I misunderstand these new modes. I need to do more reading first.
I usually end the loop through an automation in the current active_power approach.

The varying PV is important in the scenario where export prices are negative, but import prices are still positive (tax difference in belgium)

infradom:
> I have checked the documentation again and I noticed two really weird conditions for the Mode 9:
> 
> ```
> No limit in Export Control / Power Limit / Main Breaker Limit
> A forced charging period has been set, during which the inverter is allowed to charge from the grid.
> ```
> 
> Maybe these are the reason why it doesn't do anything for me as I definitely have the `export_control_user_limit` set and I have no clue what the `forced charging period` is good for when you use the modbus powercontrol to control the inverter😕

Thanks, I dont understand this either. To be continued...

ggg1psyyy:
> > I have checked the documentation again and I noticed two really weird conditions for the Mode 9:
> > ```
> > No limit in Export Control / Power Limit / Main Breaker Limit
> > A forced charging period has been set, during which the inverter is allowed to charge from the grid.
> > ```

> > Maybe these are the reason why it doesn't do anything for me as I definitely have the `export_control_user_limit` set and I have no clue what the `forced charging period` is good for when you use the modbus powercontrol to control the inverter😕
> 
> Thanks, I dont understand this either. To be continued...

As far as I understood this, this is not a prerequisite but just the example used in this scenario ... but I absolutely might be wrong here

ggg1psyyy:
> @ggg1psyyy Thanks for your feedback and your work on this, I was thinking more or less to introduce a separate recompute loop for these modes (8 and 9) . Maybe I am missing something:
> 
> * you say mode 8 and 9 dont need a loop because they have defined stopping points
> 
> In my opinion, we need to recompute the targets (PV or battery charge) every second in order to match the variations in the house load. So for me, automatic recomputation within the loop makes sense, unless I misunderstand these new modes. I need to do more reading first. I usually end the loop through an automation in the current active_power approach.
> 
I think our use cases are somehow different. I use my own NetDeamon app to calculate the necessary state of the inverter (every 15s), with this app I have the following conditions concerning import/export price:
1. both prices are positive -> default "Self-Use-Mode"
2. export price negative, import price positive -> Self-Use-Mode, but export_limit = 0W (or 10W to make it easier for the inverter to adjust to the load) , if there is a negative import price window coming up I disable battery charging (by setting the charging limit to 0), so I can use as much grid energy as possible in this window
3. both prices are negative -> Mode 9 until 95% SoC (PV-Limit: 0W, battery-charge: 10kW) if the SoC reaches 95% I simply switch the inverter off until again case 1 or 2 

> The varying PV is important in the scenario where export prices are negative, but import prices are still positive (tax difference in belgium)

In this case I just use point 2 -> the inverter automatically adjust the available PV power according to the house load (even if it needs some seconds to adjust sometimes)
image
image

ggg1psyyy:
as the import price is not too expensive at the moment, I tested it manually with the following settings:
image
other than the charging current being too low (no idea why) it seems to be working fine:
image
but it seems to not disabling the remote mode if the target SoC is reached:
image
I'd propose to check the "target finished" flag and disable remote mode if it switches to "finished" as the inverter has set this flag on my test just now.
After setting the power control mode to "Disabled", everything goes back to normal:
image

ggg1psyyy:
> > I have checked the documentation again and I noticed two really weird conditions for the Mode 9:
> > ```
> > No limit in Export Control / Power Limit / Main Breaker Limit
> > A forced charging period has been set, during which the inverter is allowed to charge from the grid.
> > ```
> > 
> > Maybe these are the reason why it doesn't do anything for me as I definitely have the `export_control_user_limit` set and I have no clue what the `forced charging period` is good for when you use the modbus powercontrol to control the inverter😕
> 
> Thanks, I dont understand this either. To be continued...

as you can see in my previous example I also have the export limit set, while the remote mode was active, so I really believe it's just in this example.

infradom:
> 2\. export price negative, import price positive -> Self-Use-Mode, but export_limit = 0W (or 10W to make it easier for the inverter to adjust to the load) , if there is a negative import price window coming up I disable battery charging (by setting the charging limit to 0), so I can use as much grid energy as possible in this window

Thanks for your explanation; I currently also work with the export_limit when injection prices are negative, but this parameter is written to Eeprom, so we should not write it too often (I assume twice a day wont kill the inverter the first 10 years, but you never know). This is why I am working on this new recompute loop. I already extended the source code in my infradom development branch (https://github.com/infradom/homeassistant-solax-modbus), but testing will have to be done in the coming days.

I will look into your approach in more detail also; Thanks for the feedback !!

infradom:
@ggg1psyyy In the new loop mode I am working on, the mode 9 does not make much sense, since the mode8 loop can be terminated if a target_SOC is reached. So I believe I will implement 3 variations of Mode 8:

- Mode 8 raw, as documented in the solax docs, but with additional repeat loop without reevaluation
- Mode 8(1) - negative injection price - evaluation loop
- Mode 8(2) - negative injection and consumption price - evaluation loop

I still need to do some sanity checks before I start testing this on my system. 
The mode transistions can be done with simple automations.
I still need to make sure that both remotecontrol loops cannot interfere with each other, so people should not use my fork and wait until it has been tested properly and merged into the @wills106 main line.

infradom:
@ggg1psyyy @wills106 
In my development fork, the negative price autorepeat loop is starting to work as expected. I still need to do more testing, and more important work on the documentation. I know that my autorepeat work might confuse users, since there are now multiple ways of controlling the remotecontrol entities. It is good that you named your versions _direct; the duplicates I introduced might confuse users, so I wonder if I should not give them an extension _loop
Backward compatibility could be an issue if I rename the older entities remotecontrol_xxxxx_loop

I will give it some more thought before I issue a PR

infradom:
During my testing for the Mode8 autorepeat loop, I noticed some strange dashes (longer than a normal dash) in the strings for  modbus_power_control and related entities. In my fork, I replaced the slightly longer dashes by normal dash signs, and added some spaces. For automations, the formatting is very important.  Normally, this correction will be part of the pull request for the Mode8 autorepeat loop.
```
        name="Modbus Power Control (direct)",
        key="modbus_power_control_direct",
        register=0x7C,
        write_method=WRITE_MULTISINGLE_MODBUS,
        option_dict={
            0: "Disabled",
            1: "Enable Power Control Mode",
            2: "Enable Electric Quantity Target Control Mode",
            3: "Enable SOC Target Control Mode",
            4: "Push Power-Positive/Negative Mode",
            5: "Push Power – Zero Mode",
            6: "Self-Consume - Charge/Discharge Mode",
            7: "Self-Consume - Charge Only Mode",
            8: "Individual Setting - Duration Mode",
            9: "Individual Setting - Target SOC Mode",
        },

        name="RemoteControl Power Control Mode (direct)",
        key="remote_control_power_control_mode_direct",
        register=0xA0,
        write_method=WRITE_MULTISINGLE_MODBUS,
        option_dict={
            0: "Disabled",
            8: "Individual Setting - Duration Mode",
            9: "Individual Setting - Target SOC Mode",
        },

```

infradom:
@ggg1psyyy:  I have created ReadTheDocs pages for the Mode 1 and Mode 8 autorepeat mechanisms. It would be nice if you could create a small page that describes how to use the _direct approach.
See [Mode 1 doc](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/) and [Mode 8 doc](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/) as well as [the developer guide - autorepeat section](https://homeassistant-solax-modbus.readthedocs.io/en/latest/developer_guide/)

wills106:
Just released [2025.07.2](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2025.07.2) with this change.

ggg1psyyy:
> @ggg1psyyy: I have created ReadTheDocs pages for the Mode 1 and Mode 8 autorepeat mechanisms. It would be nice if you could create a small page that describes how to use the _direct approach. See [Mode 1 doc](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/) and [Mode 8 doc](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/) as well as [the developer guide - autorepeat section](https://homeassistant-solax-modbus.readthedocs.io/en/latest/developer_guide/)

I'll take a look over it over the weekend

infradom:
> I'll take a look over it over the weekend

thanks, no hurry

johny-mnemonic:
@infradom thanks for adding the Mode 8. I just tried the duration mode and it works as described 🎉
I will try those modes for negative prices when I find the time.

Would be nice to have also the target SOC mode🤞
That would be useful for the situation when you calculate how much energy you need to keep in the battery to survive till the next day and you want to sell the rest. Sometimes unpredictable load comes during the duration based discharge and discharges the battery more than desired, while with the target SOC you would be able to let the inverter stop the discharge once the target SOC is reached.
Not sure if it would be also usable for limiting the maximum charge SOC. If yes than that would be perfect as so far I am using the `battery_charge_upper_soc`  for it, which writes to EEPROM which is definitely less than ideal.

infradom:
I will look into your request, when I implemented mode 8, I intended to implement also mode 9 (target SOC), but at that time, I did not see much use as the SolaX internal system does not do what I want when the target SOC would be reached. 
In my case, it is the automation that decides when to go in which scenario. So my automations switch between 

- negative price (RC)
- normal selfuse, 
- feedin priority (RC)
- pushmode discharge (RC) in summer (when prices are higher in the morning or evening)
- pushmode charge (RC) in winter (when price is lower at night)

These automations do that on the base of price, battery level, solar forecast and house_load forecast
I once started to develop a mathematical ems model for this, but lack of time stopped that effort.

I believe you should rethink what you can achieve with automations and the existing Mode8 submodes. 
While the autorepeat loop adjusts within seconds, the automations will change the mode only once every few hours.

So target SOC can be achieved easily with mode8 and a simple automation. The advantage is that you can then determine which mode to choose when target SOC is reached.

To be continued .. these were just some quick thoughts

johny-mnemonic:
I see your point and I also have an automation which decides what the inverter should do, but I switch between these states:
- normal backup mode
- feedin priority (RC)
- battery control (RC)
- limit/disable export (now done by `export_control_user_limit` but I will change it to mode 8)
- disable battery charge (done by `battery_charge_upper_soc`)

I am using the battery control (RC) to enforce export or to charge battery from cheap grid. 
What is the advantage of using  pushmode instead? Seems equivalent to me🧐
I might be wrong, but I think both situations would be better off served by "target SOC" mode as now I just calculate charge/discharge speed from the target SOC and let it run for the whole hour using the autorepeat loop, hoping it will really reach the target SOC. Would be simpler if I would be able to tell the inverter the target SOC as I generally don't care how fast it will charge or discharge if it manages to reach the target SOC in that hour (I have hourly prices). My experience is that the inverter itself is much better than any HA automation when computing what to do.

Thanks to this new mode 8 I will be able to stop limiting/stopping export by writing to EEPROM, which is great, but I don't see how to combine it with stopping the battery charge.
It happens a lot that the prices are low/negative and battery is charging way too fast way too early, so I want to basically just let the home run from the PV and stop both charging and exporting as there is no point to wear out the battery by letting it sit at 100% SOC half of the day. 
I have the export limited by the grid operator. My PV generates up to 2x more than I am allowed to export, so even when I can export (price is good) it can still happen that I need to stop charging the battery before noon.

infradom:
> What is the advantage of using pushmode instead? Seems equivalent to me🧐

yes, it is another way of reaching the same. Personally, I like mode 8 better.
I had problems with target soc, the inverter moved to an unwanted mode once target was reached (if I remember well)
I let my automations detect when the target SOC is reached, so I didnt need it . You are right that is is better to let the inverter determine what is best to reach the target, so if I find time, I will certainly implement this. Maybe we should create a feature request for the follow up on this

Thanks for the feedback !

PR #1396, opened 2025-05-30
Add modbus values verification

julienbourgainpro:
On Sofar inverters, Modbus can shift the values, resulting in clearly false values. Lowering the read frequency results in fewer errors, but they persist.

The goal of this fix is ​​to reject any message with at least one error.

As a reminder, in the event of a ModbusIOException, no value is returned, which considers the inverter to be asleep and slows the number of requests by a factor of 10.

With this fix, I can increase the Modbus refresh rate to 5 seconds instead of 30 seconds. If an error occurs, there is a 50-second slowdown. In my case, I can have about 10 read errors per day, which overall gives me greater responsiveness.

Add value verification
- **value should exist in Dict** i have tested that on sofar HYD and solax X3 G4 inverter without problem but i don't have others brands
- Number can be in a range of values
- Add values validation in plugin_sofar.py


julienbourgainpro:
To add some context, for future development, in this pull request, the max values are for 100kw inverter. 
I think in the futur we can up to more power but i haven't recheck the sofar solar product list.

SzczepanLeon:
This fix introduce problem with Solax.


It breaks whole communication with inverter.

Issue #1408, opened 2025-06-02
[Bug]: Modbus value 6 not found


leoV:
### Describe the bug

All status are 'unknown', modbus not connected, however model and serial are read. 
ERROR (MainThread) [custom_components.solax_modbus] ModbusIOError: Modbus Error: [Input/Output] Modbus value 6 not found in scale mapping for meter_2_direction

### Integration Version

2025.05.3

### Homeassistant core version

Home Assistant OS Core 2025.5.3 Operating System 15.2

### Inverter brand

Solax	X3-Hybrid-G1/G2

### Plugin used

plugin_solax.py

### Serial prefix

H3UE06

### Inverter firmware versions

 ARM 4.45-1.05 Main DSP 4.63 Slave DSP 0.00 Module version 2.034.03

### Connection Method

TCPIP cable

### Dongle firmware

none

### Detailed Error Log

2025-06-02 12:22:36.896 INFO (MainThread) [custom_components.solax_modbus] Inverter is not connected, trying to connect
2025-06-02 12:22:37.043 INFO (MainThread) [custom_components.solax_modbus] Inverter connected at XXX:502
2025-06-02 12:22:37.044 INFO (MainThread) [custom_components.solax_modbus.plugin_solax] SolaX: trying to determine inverter type
2025-06-02 12:22:37.048 INFO (MainThread) [custom_components.solax_modbus.plugin_solax] Read SolaX 0x0 serial number: H3UE06XXX
2025-06-02 12:22:37.072 INFO (MainThread) [homeassistant.components.button] Setting up solax_modbus.button
2025-06-02 12:22:37.074 INFO (MainThread) [custom_components.solax_modbus.button] hub.wakeuButton: SolaxModbusButtonEntityDescription(key='battery_awaken', device_class=None, entity_category=None, entity_registry_enabled_default=False, entity_registry_visible_default=True, force_update=False, icon='mdi:battery-alert-variant', has_entity_name=False, name='Inverter Battery Awaken', translation_key=None, translation_placeholders=None, unit_of_measurement=None, allowedtypes=6144, register=86, command=1, blacklist=None, write_method=1, value_function=None, autorepeat=None)
2025-06-02 12:22:37.074 INFO (MainThread) [homeassistant.components.number] Setting up solax_modbus.number
2025-06-02 12:22:37.078 INFO (MainThread) [homeassistant.components.select] Setting up solax_modbus.select
[...]

Issue #1407, opened 2025-06-02
[Bug]: 'Inverter Brand' - 'Inverter Model' Description of Error


JanSekanina:
### Describe the bug

Integration not working after update to version 2025.05.3

This error originated from a custom integration.

Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:725
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 09:04:30 (1 occurrence)
Last logged: 09:04:30

ModbusIOError: Modbus Error: [Input/Output] Modbus value 16 not found in scale mapping for grid_first_time_6

### Integration Version

2025.05.3

### Homeassistant core version

2025.5.3

### Inverter brand

Growatt

### Plugin used

plugin_growatt.py

### Serial prefix

SPH

### Connection Method

RS485-Wifi

### Detailed Error Log

This error originated from a custom integration.

Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:725
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 09:04:30 (1 occurrence)
Last logged: 09:04:30

ModbusIOError: Modbus Error: [Input/Output] Modbus value 16 not found in scale mapping for grid_first_time_6

### Additional context

![Image](https://github.com/user-attachments/assets/8897f51b-4406-4b5b-b457-3d581c5cc6b4)

JanSekanina:
The lastest working version in my environment is 2025.05.1

JanSekanina:
Solved in ticket #1409
Build 2025.06.1

Issue #1404, opened 2025-06-01
[Bug]: Inverter connection lost after upgrade to 2025.05.3


ezfrag2021:
### Describe the bug

All entities unavailable after update to latest version.

### Integration Version

2025.05.3

### Homeassistant core version

2025.5.3

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H3DE05

### Inverter firmware versions

Unknown

### Connection Method

Ethernet

### Dongle firmware

2.034.03

### Detailed Error Log

This error originated from a custom integration.

Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:725
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 07:35:19 (6 occurrences)
Last logged: 07:47:34

ModbusIOError: Modbus Error: [Input/Output] Modbus value 5 not found in scale mapping for meter_1_direction

ezfrag2021:
Rolling back the integration version to 2025.05.2 or 2025.05.01 did not fix the issue. Problem may be due to upgrade of HA Core.

Edit: rolled back to the backup from Saturday and the problem is resolved so no idea what caused it.

godfreym29:
I'm also seeing this

**Detailed Error Log**
This error originated from a custom integration.

Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:725
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 10:10:59 (8 occurrences)
Last logged: 10:28:14

ModbusIOError: Modbus Error: [Input/Output] Modbus value 48 not found in scale mapping for backflow_power_switch


also getting a lot of these errors in the log


```
2025-06-01 10:33:48.986 ERROR (MainThread) [homeassistant.components.sensor] Error adding entity sensor.solis_battery_charge_current for domain sensor with platform solax_modbus
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 657, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 979, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1368, in add_to_platform_finish
    await self.async_internal_added_to_hass()
[...]

PR #1409, opened 2025-06-02
Revert Modbus value validation from enum.

julienbourgainpro:
Cause regression on some inverters #1408 #1407 #1404 #1411 #1410


JosephAbbey (approved):
This is definitely needed to ensure that we don't reject all values simply because one value seems erroneous and thus throws an error, which causes the integration to crash.

josephabbey:
I have tested this on my own instance. It works!

PR #1338, opened 2025-04-27
New Growatt entities, typo

flopp999:
I have made some big changes for Growatt. New entiteis and also some renames.
Manually added the changes in init.py from 2025.04.2

PR #1460, opened 2025-07-07
Update plugin_growatt.py

flopp999:
Communication Board Temperature
new total work time in hours, old total work time in seconds is still available
peak-entites
inverter fault-entites
inverter warning-entites


flopp999:
@wills106 i'm checking the errors

flopp999:
@wills106 errors fixed
For GEN4, is that I have so I cannot answer if other version of Growatt is the same
Lots of entities are named Battery....
I have an idea to rename all those to BMS 1 and BMS 2. For example keep/create "battery charge power" but that will be a sum of BMS1 charge power and BMS2 charge power.
But there are many more entities that may be removed/renamed

Is this OK for you?

flopp999:
@wills106 wait, found an error

flopp999:
> @wills106 wait, found an error

I have now tested the code in this PR and it works without any errors.

wills106:
> But there are many more entities that may be removed/renamed
> 
> Is this OK for you?

It's difficult to know how to answer. Previously when entities have changes people have been upset.

Depends if they are entities people class as critical for long term stats / logging and how they effect automations.

flopp999:
> > But there are many more entities that may be removed/renamed
> > Is this OK for you?
> 
> It's difficult to know how to answer. Previously when entities have changes people have been upset.
> 
> Depends if they are entities people class as critical for long term stats / logging and how they effect automations.

I understand and I agree with them :)

OK, then I will create new entities with new names but have same data, I don't think it will be that many duplicates.
Will this be OK you think?

w4iq3:
Does 2025.07.1 also include the work from https://github.com/wills106/homeassistant-solax-modbus/pull/1338 @flopp999 ? I have copied the plugin_growatt.py from that PR, but don't want to break that.

flopp999:
> Does 2025.07.1 also include the work from https://github.com/wills106/homeassistant-solax-modbus/pull/1338 @flopp999 ? I have copied the plugin_growatt.py from that PR, but don't want to break that.

Not all changes yet.
I am doing small changes for each PR, so it is easier to check the changes.
First PR was more than 2000 changes

w4iq3:
Got it, let me know if you need help writing change logs

PR #1462, opened 2025-07-09
Update plugin_growatt.py

flopp999:
* changed from allowedtypes = GEN4 to GEN4 | HYBRID, for all entities that I have added. Since GEN4 can be an inverter PV only, so with HYBRID added those will only be visible for Hybrid inverters.
* renamed throughput -> output energy, because this value is only how much energy that has been outputed, not in and out
* BMS status entity
* removed "Total Worked Time Seconds", entity in hours instead
* renamed "...module cycle" to "...module cycles", in plural


flopp999:
Tested with an clean and empty HA:  
Core 2025.7.1  
Supervisor 2025.07.1  
Operating System 16.0  
Frontend 20250702.1

no errors
tested with MID 15 KTL3-X with two BMS

PR #1463, opened 2025-07-10
Solax X1-IES Gen5 Inverter Temperature Incorrectly Scaled

TCWORLD:
It seems that a Gen 5 specific entry for inverter temperature entry was added with 'scale=0.1' as discussed in #813 to fix an issue with X3-IES inverters reporting temperature 10 times higher.

However this fix appears to make the X1-IES versions report the temperature as 1/10th of what it should be. I going to assume that this scaling is therefore only needed for the X3-IES variant.

This includes a patch to split them up.


wills106:
I have decided to merge it, as not everyone can alter code locally.

PR #1465, opened 2025-07-12
Solax: add remotecontrol autorepeat mechanism for Mode 8

infradom:
Also correct an initialization error for LOCAL DATA entities

I have made a mess, while merging the mainline updates in my fork, then I tried to rebase it all into one commit to make it more transparent, but this just added another commit, even to files I did not touch, so it might be very confusing to read the history. The files should be up to date however.


infradom:
removing this pull request, to be replaced by one with fewer commits

Issue #1473, opened 2025-07-18
[Bug]: Sofar - HYD 20KTL-3PH UnboundLocalError: cannot access local variable 'action' where it is not associated with a value


anton4:
### Describe the bug

After upgrading from 2025.06.02 to 2025.07.02 Passive update battery charge/discharge button (button.sofar_passive_mode_battery_charge_discharge) does not work anymore and giving this error in the logs - 
`2025-07-18 09:42:35.663 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [140009461365024] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 250, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2802, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2845, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1009, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
        hass, entity, func, data, call.context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1081, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/button/__init__.py", line 132, in _async_press_action
    await self.async_press()
  File "/config/custom_components/solax_modbus/button.py", line 97, in async_press
    if action == WRITE_MULTI_MODBUS:
       ^^^^^^
UnboundLocalError: cannot access local variable 'action' where it is not associated with a value`

### Integration Version

2025.07.02

### Homeassistant core version

2025.6.3

### Inverter brand

Sofar

### Plugin used

plugin_sofar.py

### Serial prefix

SP1ES120

### Connection Method

LSE-3, Modbus RTU over TCP

flopp999:
[...]

PR #1476, opened 2025-07-18
correct bug #1473 and potential other errors introduced in latest version

infradom:
The bug affected non-autorepeat  buttons that do a write_multi
Seems to have affected 

- sofar, 
- growatt
- solax rtc button


infradom:
issue is #1473 not #1773

PR #1477, opened 2025-07-20
Growatt update

flopp999:
Updated the code structure for time entities

renamed so it is consistent for timers
Time X function
e.g.
Time 1 Update
Time 1 Clear

remake for module soc, 
using a function instead

removed a duplicate entity: "bms_1_module_2_combined_current"

function for combined current
bms amp entities renamed to "combined current", now shows negative and positive values

Added combined power for modules, shows negative and positive values


TonyMathiesen:
Tested it on my setup 😊 Works nice with the individual time update buttons etc. 👍

TonyMathiesen:
A suggestion could be to set the buttons 2-9 as `entity_registry_enabled_default = False` as the other time entities. Then people can add the relevant ones in their setup 😊

flopp999:
> A suggestion could be to set the buttons 2-9 as `entity_registry_enabled_default = False` as the other time entities. Then people can add the relevant ones in their setup 😊

Great idea. I will set False for 4-9. Then you can have one Time for each function.
Load, Battery, Grid.
Then you only need to Enable/Disable

PR #1474, opened 2025-07-18
Modify sensor.py to ignore disabled entities

TonyMathiesen:
This is work in progress. It checks if known in Home Assistant before appending it to the reading blocks.

Issue #1440, opened 2025-06-25
[Bug]: Calculated Values with multiple scan groups


gec75:
### Description

This is a generic issue with calculated values (value_function) and multiple scan groups (I use Solinteg plugin).

async_read_modbus_registers_all / computedSensors / value_function is called for *EACH* scan group.
The "data" dictionary (inside async_read_modbus_registers_all) is always new, i.e. populated only for the *current* group.
This means that all value_function() calls fail when they need/use other sensor values (from "data" parameter).

### Suggested Solution

Is there a reason NOT to use a persistent dictionary? Either directly self.data, or a different one?

infradom:
Maybe I do not understand the issue well, but as far as I know, in the case of the solax plugin, this problem does not occur. Is this typical for the solinteg plugin ?

gec75:
No, the issue is generic.
I searched and indeed only solinteg and sofar plugins are using multiple scan groups. Maybe the maintainers of other plugins could also add support...
How is your scanning interval?

infradom:
Sorry, I have been away for too long from this forum, so I was not aware of the scan group concept, which probably introduced this bug. 
I think my scan interval is set to 5 seconds (or maybe 3 seconds).
If I find some time in the coming days, I will review the source code to see if I can find the cause of the problem.

infradom:
Ok, I think I understand the problem, but as I am in the middle of a big change in my development fork, it is hard to issue a PR right now. 
In my opinion, in __init__.py,  async_read_modbus_registers_all, we should modify the line:
CORRECTION: THIS DOES NOT WORK AS THE SEQUENCE MODIFICATION CREATES SIDE EFFECTS

```
        for reg in self.computedSensors:
            descr = self.computedSensors[reg]
            data[descr.key] = descr.value_function(0, descr, data)

        if group.readFollowUp is not None:
            if not await group.readFollowUp(self.data, data):
                _LOGGER.warning(f"device group check not success")
                return True

        for key, value in data.items():
            self.data[key] = value
```

into


```
       # sequence changed
        if group.readFollowUp is not None:
            if not await group.readFollowUp(self.data, data):
                _LOGGER.warning(f"device group check not success")
                return True

        for key, value in data.items():
            self.data[key] = value

        for reg in self.computedSensors:
            descr = self.computedSensors[reg]
            self.data[descr.key] = descr.value_function(0, descr, self.data) # use self.data instead of local data
```


[...]

PR #1478, opened 2025-07-20
issues #1440 #1473 #1482; dont poll disabled sensors, better handling of offline situations, multiple_scan_groups enhancement

infradom:
EDITED
- implement copy of suggested #1440 issue  solution - thanks @gec75 
- reduce messages when inverter is offline or asleep issue 

- copy/implement #1474 do not poll registers that are disabled. Thanks @TonyMathiesen. If a  sensor is enabled, the read_block structure needs to be recomputed, this caused some significant code moves. It which seems to work now. 
Needs to be reviewed and tested further by developers of the scan_group mechanism. 

- set modbus RETRIES = 0 instead of 6 (did not make sense in my opinion as the polling loop will retry automatically) 
- make some code more readable
- suppress autorepeat writes when inverter is offline
- Some code (e.g. splitInBlocks) moved from sensor.py to __init__.py
- modify value_function_pv_power_total, so that it does not scan entire datadict (and catches the old pv_power_total). The function now takes 8 times less compute cycles.
- I guess this solves issue #1484 also (probably a duplicate of #1482)

As this is only tested on Solax, scan groups are not really tested. Edit: tested with a local solax plugin version with scan groups. 
Enabling sensors seems to work, Disabling sensors: works also now after some debugging. 

You can also find the source code in my development fork: https://github.com/infradom/homeassistant-solax-modbus

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -358,6 +358,7 @@ class SolaXModbusHub:
         self.entry = entry
         self.device_info = None
         self.blocks_changed = False
+        self.initial_groups = {} # as returned by the sensor setup 
 
         _LOGGER.debug("solax modbushub done %s", self.__dict__)
 
@@ -464,13 +465,14 @@ class SolaXModbusHub:
             async def _refresh(_now: Optional[int] = None) -> None:
                 await self._check_connection()
                 await self.async_refresh_modbus_data(interval_group, _now)
-
+            _LOGGER.info(f"starting timer loop for interval group: {interval}")
             interval_group.unsub_interval_method = async_track_time_interval(
                 self._hass, _refresh, timedelta(seconds=interval)
             )
 
         device_key = self.device_group_key(sensor.device_info)
         grp = interval_group.device_groups.setdefault(device_key, self.empty_device_group())
+        _LOGGER.debug(f"adding sensor {sensor.entity_description.key} ")
         grp.sensors.append(sensor)
         self.blocks_changed = True # will force rebuild_blocks to be called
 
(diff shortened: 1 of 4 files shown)


infradom:
The latest update also solved the problem when sensors are disabled by the end user. Cause: SimpleNamespace usage was inconsistent between __init__.py and sensor.py. I now shared the definition between both files.

infradom:
In my local copy ( not yet on github), I have introduced a 3 second fast scan group for solax, and changed the default solax scangroup to 15 second. All seems to work well, so I change from draft to normal PR, but review and testing by the solinteb, sofar, growat ... developers is very welcome

infradom:
found another bug I want to solve first ...

gec75:
Unfortunately I can't help... I'm on vacation... 😊

infradom:
> Unfortunately I can't help... I'm on vacation... 😊

No problem, I will take some more time to test in-depth. Enjoy

TonyMathiesen:
Excellent work @infradom 👏🏼 Regarding the fast scan group, is it only required to add it in the plugin file?

infradom:
Yes, is sufficient to add scan_group=SCAN_GROUP_xxx statements to the sensor entities,  ..
Dont forget to reconfigure your integration instance to adapt the polling speeds.
For some reason, I believe it is best to use it on the version on which PR #1478 is implemented (I cannot remember why I  ..)

I would also like to make the introduction of can groups easier by providing plugin-specific defaults for holding and input registers (see discussion #1480,, this discussion is not only relevant for solax plugin )

andbad:
Yesterday I encountered a problem with 2025.07.4 (the values update was stuck or unstable), I think it was related to this change, so I reverted to the previous one.
Inverter Sofar HYD.

I'm tring update 2025.07.5

By(t)e

gec75:
I also tried a remote update (.4,.5), and some values are not reported anymore.
I didn't see any log entries.
As I am not at home to investigate, I had to downgrade to .3

infradom:
I am fully aware this PR causes problems for some users. Unfortunately these users do not send their (debug) logs.
I cannot simulate on inverter models that I do not own.

infradom:
For solis, I received relevant logs, which helps finding a solution (for solis). Draft PR issued, awaiting testing by a solis owner.
For other models, I can only hope to get more logs.
Removing the integration, rebooting and reinstalling the integration helps for some users.

andbad:
Here are the logs for a Sofar inverter.
Are they sufficient or do you need more?
Thanks for your work.
[log.txt](https://github.com/user-attachments/files/21466513/log.txt)

By(t)e

PR #1509, opened 2025-07-30
Add kW value to Solax H34 inverter models

rosenrot00:
This adds the kW value to the inverter name if it has a H34 serial number. To my knowledge there is only integer values from 5-15kW. Please correct me if I'm wrong.

To my knowledge there are H34A/B/C/T with the same 5-15kW ranges. Please correct me if I'm wrong and we need to differentiate between the different types.


wills106:
> H34A/B/C/T with the same 5-15kW ranges.

As far as I am aware they are all the same. The A-T are probably just minor hardware revisions?

Issue #1503, opened 2025-07-29
[Bug]: Sofar - HYD 20KTL-3PH missing values on text/dropdown boxes


anton4:
### Describe the bug

After upgrading to from 2025.06.3 to 2025.07.x integration does not read values for dropdown boxes after HA restart.
Changing values do work and are populated after manually changing, but after HA reboot they all are empty.

<img width="1012" height="1007" alt="Image" src="https://github.com/user-attachments/assets/8de67f48-e4b3-42fb-a99d-7cc4a8e573db" />

### Integration Version

2025.07.6

### Homeassistant core version

2025.7.4

### Inverter brand

Sofar

### Plugin used

plugin_sofar.py

### Serial prefix

SP1ES120

### Connection Method

LSE-3 with Modbus RTU over TCP

infradom:
Should be the same cause as the other similar issues recently reported (issue #1496)
Fix is on its way with PR #1501

wills106:
I have just released [2025.07.7](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2025.07.7) with these changes.

marbell90:
The Problem is in 2025.07.7 still there.

infradom:
> The Problem is in 2025.07.7 still there.

Hard to believe, it looked so much similar to what we experienced. If you want us to investigate further, we will need a debug log that includes a restart of HA

Have you checked if the select and number entities have not received a new name (with a _2 suffix) ? In that case, renaming them without _2 will solve the problem. you need to open the settings dialog of a failing entity to see this

marbell90:
> > The Problem is in 2025.07.7 still there.
> 
> Hard to believe, it looked so much similar to what we experienced. If you want us to investigate further, we will need a debug log that includes a restart of HA
> 
> Have you checked if the select and number entities have not received a new name (with a _2 suffix) ? In that case, renaming them without _2 will solve the problem. you need to open the settings dialog of a failing entity to see this

I had to switch back to 2025.07.3. I attached the Debug-Log. Hope it helps. Thanks. The entities seem all fine.

[home-assistant_solax_modbus_2025-07-29T21-55-50.717Z.log](https://github.com/user-attachments/files/21497972/home-assistant_solax_modbus_2025-07-29T21-55-50.717Z.log)

infradom:
> The Problem is in 2025.07.7 still there.

@marbell90 :
[...]

PR #1512, opened 2025-07-31
is_entity_enabled now works without dirty fix; sofar #1503; consider sensor_key dependencies

infradom:
- the unique_id of some entities separate platform from key with '_' others with '.' . Now we try with both. Unclear why. Maybe older versions may have used a different separator

- Sofar: modify entity_enabled_by_default=False by internal=True. There might be another issue causing #1503 a certain block returns modbus read errors. To be examined separately.

- Thanks @thomas7475 for contributing the .sensor_key dependency mechanism. Some sensors were not read because they were referred to by a sensor_key statement in a control with a different key


infradom:
@wills106 : could solve some remaining problems with disabled registers, so best to merge, I suggest to create another beta release

PR #1520, opened 2025-08-04
Generalize the X3-HYB-G4 PRO inverter model identification

rosenrot00:
now all X3-HYB-G4 PRO inverters should be recognized. Their kW value is incorporated in hex format in the serial number.

10K12 is falling out of this scheme but my best guess is that this is either weird from Solax or a mistake?

Worst that happens with my code is that the 10K12 would show as 18kW version.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -8501,7 +8501,7 @@ class solax_plugin(plugin_base):
             invertertype = HYBRID | GEN6 | X3  # X3-HYB-G4 PRO
             if kw_value >= 8:
                 invertertype |= MPPT3
-            self.inverter_model = f"X3-G4PRO-{kw_value}kW"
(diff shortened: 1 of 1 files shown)


rosenrot00:
By the way, I was asking myself if we should change the inverter model so it matches the datasheet naming? in this case from X3-G4PRO-4kW to X3-HYB-4.0-P. It is not a big thing but why introducing new naming if there is one from the users bill/datasheet/manual.

infradom:
@wills106 : looks ok, but have not tested this myself

wills106:
I would have to revisit if it's possible to match the data sheet name. But previously there has been overlap between Generations with X3-HYB-4.0-x etc So not so easy to know exactly which Inverter we are talking about.

Sometime they are just named after what people have previously requested aswell.

rosenrot00:
> Sometime they are just named after what people have previously requested aswell.

I saw that. I think it might be good to follow the vendor naming. However, I can’t tell if they overlap. It isn’t the most important thing in the world in any case as I don’t know how many people would pay attention.

PR #1521, opened 2025-08-04
Generalize inverter model detection also for X1-VAST-6K and X3-MIC-3K-G2

rosenrot00:


infradom:
@wills106 : looks ok to merge, but have not tested this myself

PR #1525, opened 2025-08-05
ESP8266 support added to ESPHome modbus bridge adapter

rosenrot00:
I added support for the ESP8266 and Arduino framework.


infradom:
@wills106 : documentation update

Issue #1533, opened 2025-08-12
[Bug]: Integration faiiled HASS 2025.8


tiaxter:
After upgrading Home Assistant to the version 2025.8 the integration loading failed showing these errors:
```
Logger: homeassistant.setup
Source: setup.py:343
First occurred: 12:47:37 PM (1 occurrence)
Last logged: 12:47:37 PM

Setup failed for custom integration 'solax_modbus': Unable to import component: No module named 'pymodbus.payload'
Traceback (most recent call last):
  File "/nix/store/p2b2wgc2d6b5nymp1fibacf92mq9smw6-homeassistant-2025.8.0/lib/python3.13/site-packages/homeassistant/setup.py", line 343, in _async_setup_component
    component = await integration.async_get_component()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/p2b2wgc2d6b5nymp1fibacf92mq9smw6-homeassistant-2025.8.0/lib/python3.13/site-packages/homeassistant/loader.py", line 1023, in async_get_component
    self._component_future.result()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/nix/store/p2b2wgc2d6b5nymp1fibacf92mq9smw6-homeassistant-2025.8.0/lib/python3.13/site-packages/homeassistant/loader.py", line 1003, in async_get_component
    comp = await self.hass.async_add_import_executor_job(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self._get_component, True
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/nix/store/cr7c6cf8y7sbqwq9bj7lrm977x7c4mv9-python3-3.13.5/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/nix/store/p2b2wgc2d6b5nymp1fibacf92mq9smw6-homeassistant-2025.8.0/lib/python3.13/site-packages/homeassistant/loader.py", line 1063, in _get_component
    ComponentProtocol, importlib.import_module(self.pkg_path)
                       ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
[...]

PR #1539, opened 2025-08-16
Update const.py to adopt to pymodbus >= 3.0

ITDancer13:
Adopt imports to support both pymodubus < 3.0 and >= separately.

Solves https://github.com/wills106/homeassistant-solax-modbus/issues/1533


infradom:
@wills106 : Seems to be safe to merge in my opinion

PR #1551, opened 2025-08-20
Use custom Endian enum class if cannot import it

tiaxter:
Reading the release changes of `pymodbus` I noticed that after version 3.10 the `Endian` enum class was deleted. And Home Assistant 2025.8 decided to update its `pymodbus` version higher than 3.10 and so it broke the integration on latest update

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -76,7 +76,6 @@ _LOGGER = logging.getLogger(__name__)
 # else:
 #    Endian_BIG = Endian.BIG
 #    Endian_LITTLE = Endian.LITTLE
-from pymodbus.constants import Endian
 from pymodbus.exceptions import ConnectionException, ModbusIOException
 from .payload import BinaryPayloadBuilder, BinaryPayloadDecoder, Endian
 from pymodbus.framer import FramerType
--- a/custom_components/solax_modbus/const.py
+++ b/custom_components/solax_modbus/const.py
@@ -12,10 +12,17 @@ from homeassistant.components.select import SelectEntityDescription
 from homeassistant.components.switch import SwitchEntityDescription
 from homeassistant.components.button import ButtonEntityDescription
 from homeassistant.helpers.entity import EntityCategory
+from enum import Enum
 try:
     from pymodbus.constants import Endian  # for pymodbus < 3.0
 except ImportError:
-    from pymodbus.payload import Endian    # for pymodbus >= 3.0
+    try:        
+        from pymodbus.payload import Endian    # for pymodbus >= 3.0
+    except Exception:
+        class Endian(str, Enum):
+            AUTO = "@"
+            BIG = ">"
+            LITTLE = "<"        
 from datetime import datetime, timedelta
 from dataclasses import dataclass, replace
 import pathlib
--- a/custom_components/solax_modbus/payload.py
+++ b/custom_components/solax_modbus/payload.py
@@ -18,7 +18,7 @@ from array import array
 # pylint: disable=missing-type-doc
 from struct import pack, unpack
 
-from pymodbus.constants import Endian
+from .const import Endian
 from pymodbus.exceptions import ParameterException
 from pymodbus.logging import Log
 


rosenrot00:
Thanks for bringing my idea into a PR. After looking at the code, I realized that the previous approach with payload.py was essentially just a backward-compatibility shim, duplicating deprecated logic and keeping old code paths alive. For me, that doesn’t go far enough. What I’d suggest instead is introducing a pymodbus_compat.py that is future-proof by trying to use the new pymodbus 3.11 functions first and only falling back to emulated ones if they are not available. The advantage of this approach is that once all systems are running on pymodbus ≥3.11, the wrapper can simply be removed and the code will continue to work unchanged.

This is untested but should be more or less it:
```
# pymodbus_compat.py
from __future__ import annotations
from enum import Enum

# Endian: prefer modern location; provide a tiny shim if neither is present.
try:
    from pymodbus.constants import Endian  # ≥3.5 (Enum with BIG/LITTLE)
except Exception:
    class Endian(Enum):
        BIG = "big"
        LITTLE = "little"

def _to_str_endian(e) -> str:
    if isinstance(e, Enum):
        v = getattr(e, "value", None)
        if isinstance(v, str):
            return v
        return getattr(e, "name", "BIG").lower()
    s = str(e).lower()
    return "big" if "big" in s else "little"

# Prefer the new API
try:
    from pymodbus import convert_from_registers as _cfr, convert_to_registers as _ctr

    def convert_to_registers(value, data_type: str, byteorder=Endian.BIG, wordorder=Endian.LITTLE):
        return _ctr(value, data_type, _to_str_endian(byteorder), _to_str_endian(wordorder))

    def convert_from_registers(regs, data_type: str, byteorder=Endian.BIG, wordorder=Endian.LITTLE):
        return _cfr(regs, data_type, _to_str_endian(byteorder), _to_str_endian(wordorder))

except Exception:
    # Fallback to legacy payload API (available in old pymodbus)
    from pymodbus.payload import BinaryPayloadBuilder, BinaryPayloadDecoder  # type: ignore
    try:
        from pymodbus.constants import Endian as _OldEndian  # type: ignore
    except Exception:
        class _OldEndian(Enum):
            BIG = "big"
            LITTLE = "little"

    def _old_endian(e):
        return _OldEndian.BIG if _to_str_endian(e) == "big" else _OldEndian.LITTLE

    def convert_to_registers(value, data_type: str, byteorder=Endian.BIG, wordorder=Endian.LITTLE):
        dt = data_type.lower()
        b = BinaryPayloadBuilder(byteorder=_old_endian(byteorder), wordorder=_old_endian(wordorder))
        if   dt == "uint16":   b.add_16bit_uint(int(value))
        elif dt == "int16":    b.add_16bit_int(int(value))
        elif dt == "uint32":   b.add_32bit_uint(int(value))
        elif dt == "int32":    b.add_32bit_int(int(value))
        elif dt == "uint64":   b.add_64bit_uint(int(value))
        elif dt == "int64":    b.add_64bit_int(int(value))
        elif dt == "float32":  b.add_32bit_float(float(value))
        elif dt == "string":   b.add_string(str(value))
        else:
            raise ValueError(f"Unsupported data_type: {data_type}")
        return b.to_registers()

    def convert_from_registers(regs, data_type: str, byteorder=Endian.BIG, wordorder=Endian.LITTLE):
        dt = data_type.lower()
        d = BinaryPayloadDecoder.fromRegisters(
            list(regs),
            byteorder=_old_endian(byteorder),
            wordorder=_old_endian(wordorder),
        )
        if   dt == "uint16":   return d.decode_16bit_uint()
        elif dt == "int16":    return d.decode_16bit_int()
        elif dt == "uint32":   return d.decode_32bit_uint()
        elif dt == "int32":    return d.decode_32bit_int()
        elif dt == "uint64":   return d.decode_64bit_uint()
        elif dt == "int64":    return d.decode_64bit_int()
        elif dt == "float32":  return d.decode_32bit_float()
        elif dt == "string":   return d.decode_string(len(regs) * 2)
        else:
            raise ValueError(f"Unsupported data_type: {data_type}")
```
In the Solax relevant files we could replace
`from .payload import BinaryPayloadBuilder, BinaryPayloadDecoder, Endian`
with
`from .pymodbus_compat import convert_to_registers, convert_from_registers, Endian`
and change the calls to the new APIs.

That should be future proof and in case pymodbus introduces new API changes and we need to cover more than one pymodbus version we could simply adapt the pymodbus_compat.py. When all systems would be on pymodbus ≥3.11, we could simply remove the pymodbus_compat.py and the new APIs could be used directly (if that ever happens).

We could introduce this to Solax and in case it works roll it out to the rest. But this is a more complex PR that I would be happy to hear oppinions from @wills106 and @infradom before working on it.

infradom:
Thanks a lot ! 
Once I have resolved some other issues, I will spend some time on review and testing

user86000:
please approve this pull request
integration doesn't load on 2025.9.0 beta

wills106:
@user86000 Are you running this code change locally to confirm that it's working correctly?

tiaxter:
I'm running it locally and I can confirm that's working

Il lun 1 set 2025, 20:08 wills106 ***@***.***> ha scritto:

wills106:
Has this been tested on the [2025.09.1](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2025.09.1) codebase? Also which Inverters have you tested this on @tiaxter and also @user86000 
Supporting payload.py is becoming a bit of a hack on top of another hack.

Part of July and all of August was spent trying to clean up previous code changes.
There is already discussion in #1533 to look at correctly addressing the changes made in pyModbus to use the proper methods where possible. Unless you are using the built-in HA Modbus integration for something else, you can always downgrade pyModbus.

user86000:
Yes works on 2025.9.1 codebase locally on two different systems
Inverter ist m-tec , intentically with solinteg

wills106:
I'll approve this. But work to correctly address the pyModbus changes outlined in #1533 needs working on.

HellixFix:
hi, today tried 2025.9.4 (from 8.7) and have Endian problem with my Growatt MOD 10KTL...  9.1 works OK
should I create the issue?

wills106:
> have Endian problem with my Growatt

Could you put the error log into #1533
Thanks

Issue #1546, opened 2025-08-19
[Bug]: Solis - To/From battery no longer displayed


J14111:
### Describe the bug

Hi,

Something has changed recently, the to\from battery shows 0 for me all the time now

Has something changed or what do I need to check?

<img width="662" height="130" alt="Image" src="https://github.com/user-attachments/assets/8e13e2b7-86d1-4ea6-a170-c2a35f7f7faa" />

### Integration Version

2025.08.1

### Homeassistant core version

2025.8.2

### Inverter brand

Solis

### Plugin used

plugin_solis.py

### Serial prefix

Unsure

### Connection Method

TCP Eth adapter

infradom:
Which version were you coming from?
Can you check whether battery_charge_direction is enabled? As a quick fix, it should be enabled (was disabled by default).
In the next version, we will make it internal to that this issue disappears completely.

wills106:
Just released [2025.08.2](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2025.08.2) with this change.

J14111:
Thanks infradom and wills106

Im not sure what version I came from but I update fairly regularly so it would have been a recent one.

I have just updated to 2025.08.2 and still have the same issue, I have also tried reverting to 3 or 4 previous versions but im still seeing the same.

Where do I check the battery_charge_direction setting?

infradom:
it should be in the list of entities for your inverter. Click show disabled entities. Then you should see it. From there, you can enable it (will take 30 seconds)
In version 2025.8.2, it might be invisible as it is declared internal.
If we do not find a solution, I will need a debug log that includes a HA restart (see the FAQ in the readthedocsp

isorin:
@infradom: This change should also apply to "solis_fb00" which is the same inverter with a newer firmware.
We should try to keep common changes in sync.

DKen71:
I am having the same issue, as are other users on one of the Solis Facebook groups. I update my version every time a new version is released. The most recent version 2025.08.2 still shows both entities as 0W. The last version where these entities displayed correctly is 2025.07.3. If I revert to that, then the 2 entities show the correct data and the gauges on my dashboards are visible. I have restarted Home Assistant after the install and done a full system reboot since installing 2025.08.2.

atecey:
[...]

PR #1552, opened 2025-08-21
correct part of issue #1546

infradom:
was a copy paste error


infradom:
@wills106: merging this exceptionally myself as it is a real stupid copy-paste error that absolutely needs to be corrected.

PR #1556, opened 2025-08-22
Add relevant modes to Solax remotecontrol settings

rosenrot00:
I'm currently trying to figure out how remotecontrol works and found that it would be much better if it would be clearly stated which controls I have to adjust.

Please have a look if my mode assignments are correct.


infradom:
Sorry, I do not like Incompatible changes. Automations rely on the old identities

infradom:
I will have a closer look later

rosenrot00:
> Sorry, I do not like Incompatible changes. Automations rely on the old identities

I completely understand that of course. I didn’t change the entities, just the names as my thought was that automations are bound to entities rather than names.

If we shouldn‘t do that it is also fine with me. I learned a lot looking and re-labeling the remote control settings. That is already enough for me :-)

infradom:
Ok, I understand but I find it confusing when the names do not correspond more or less to the id's
I admit that the chosen id's could have been better. 
Let me study your proposal.  
Have you read the readthedocs documents about power control ? maybe  these can be enhanced also.
I will have a clearer opinion once I have spent some more time looking at the change .
Thanks a lot

infradom:
Correction to my previous remarks:
I looked into the change in more detail and I must admit I like it, so I am in favor of implementing it.
Please be aware that I was not involved in the _direct alternatives, and I do not use those, so I cannot judge the implications (but these seem also minimal). I believe the documentation only refers to the key names, so there are probably no major changes to make in the readthedocs documents.

rosenrot00:
The names are still the same, I just added „(mode 1-9)“ to the the names to make it clear which modes the setting concern. I found it highly confusing to have remote control names without indicated modes because some are for 1-7 only for example. It was hard to find out which settings were important for me to run mode 8. With these changes I know at least which settings are relevant.

I read the readthedocs about remote controls and it helped me to initially understand better. However, I ended up consulting the Solax documents to understand what I needed. I‘m still puzzled with the „Set“ and „Update“ setting as I don’t understand why this isn‘t handled by the integration internally. When hitting the trigger button it should be „Set“ and if the settings don‘t change internally set it to „Update“. Really confusing. So I guess there is room for improvement in both the docs and the code but I‘m just using remote control since 15 minutes, maybe I shouldn’t judge.

I‘m not using _direct neither as I don’t see an advantage from it. Honestly, I would love to understand a use case for _direct because the _direct duplicate settings make it really crowded and this - at least to me - without any gain.

infradom:
In my opinion, _direct makes only sense in 2 cases:

- for power loops that do not need quick updates every few seconds (e.g. a single fixed power battery discharge)
- for users that use an external control loop (but in that case, the communication overhead may be high)

In my automations, I believe I never use update, but always use set (not 100% sure, would have to verify)

infradom:
One more remark: although I started with the mode1 autorepeat, I believe I can do everything now with the mode 8  (and my summer automations). I still have to adapt some older winter automations that need to be converted to mode8 (if possible)

rosenrot00:
> In my opinion, _direct makes only sense in 2 cases:
> 
> * for power loops that do not need quick updates every few seconds (e.g. a single fixed power battery discharge)
> * for users that use an external control loop (but in that case, the communication overhead may be high)
> 
> In my automations, I believe I never use update, but always use set (not 100% sure, would have to verify)

Ad1) From what I got you need to send updates anyway to prevent it from going into timeout and falling out of the remote control. So if you need an update every 300s anyway, I don’t see an advantage or I just get it wrong.

Ad2) But if you set the duration to 21600 seconds and you send Set all the time my bet is that it restarts the duration. You can do that if you don’t care about the duration but in this case the autorepeat functionality of the integration doesn’t work as intended. I guess.

But I‘m still uncertain.

infradom:
2) I send set commands with a duration of 20 seconds, so with that short time, it does not matter much whether we do a set or update. The autorepeat duration is set to 3600 or even more...
My fastest polling cycle is currently 3 seconds

rosenrot00:
In your case, _direct could make sense as you seem to change values in a very short amount of time. In this case writing directly to the inverter would directly update the settings without sending an entire "Set" update.

But I guess it doesn't matter much if you change the setting locally and then trigger it or write it directly and only trigger initially. Just different way of doing the same thing.

HA-Supi:
First I have to say, i can not contribute to this in case of programmer skills.
4 months ago we had a discussion at German photovoltaic forum about remote control. 

https://www.photovoltaikforum.com/thread/246018-solax-x3-g4-lade-und-entladestrom-über-rc-per-skript-in-ha-steuern/?postID=4273871#post4273871

But since April I never heard from nea. But he set up a fork.  

https://github.com/Nea74/solax_smart_charge

Maybe this give some inspiration... 

My experience about remote control was, we need many extra helper entity to support automated remote control.. 

image

infradom:
@HA-Supi  : Interesting; I dont think this is a fork, just an application on top of our integration as far as I can see. I cannot read the forum page since I use adblockers ;-( and am reluctant to disable them.
I did not yet check the source code yet to see if it uses remotecontrol and which mode.

Thanks

rosenrot00:
@HA-Supi thanks for sharing. I was thinking about something like this as well.
@infradom I just checked and he seems to use mode 8, altering the push power. However, I can't see that he can prevent the battery from charging from grid. The only way I see is calculating the PV surplus and adjust the push power accordingly. However, I think this might get tricky as clouds come faster than a 3s measurement cycle. Do you have any experience with such approaches?

Actually my goal would be to check if there is a PV surplus on the current day. At sunrise I check if the energy needed to charge the SOC to 100% is available as PV surplus beyond what I'm allowed to feed into the grid. If so, it should do house > grid > battery. However, it would be great to have a limiting factor that charges the battery in a way that if enough energy is available later in time it reduces to charging current to increase the battery lifetime. However, doing PV surplus calculations every 3s seems not to be precise enough and lead to importing more from the grid on light cloudy days.

I think the "Enabled Feedin Priority" setting in this integration does something like this - charging PV surplus?

Update: I was wrong, it seems it uses "Enabled Battery Control" which is mode 1.

infradom:
I use such a model in my negative price submode. Cloud and house load change quickly, but in my experience a 3 second reaction time is good enough. I could probably go down to 1 second, but I am already pleased with 3 second reaction

rosenrot00:
Could you tell how much grid import increase you see by running this approach? Also, how do you determine the PV surplus? There isn't a PV surplus indicator as far as I know. How do you probe the PV surplus?

infradom:
> Could you tell how much grid import increase you see by running this approach? Also, how do you determine the PV surplus? There isn't a PV surplus indicator as far as I know. How do you probe the PV surplus?

With my dynamic hourly changing prices (day ahead prices) I dont really care about PV surplus, my main objective is load batteries when grid prices are low (during daylight), and discharge batteries when prices are high in the evening. My winter algoritm is completely different  (geothermal heatpump) : I need to buy electricity when prices are low ; no selling in winter.

With my 3 second approach, from end of february till middle of november, I dont consume anything from grid with my simple algoritm (or the amount I consume is really minimal).

infradom:
As nowadays half of the days have negative injection prices, the remotecontrol loop throttles the PV down to the houseload level (plus battery charge) so that there is no export to grid during negative price times.

rosenrot00:
I have to think about that. I find it overly complicated if I'm honest. The best HA automation couldn't mitigate a house load that jumps from 300 W to 2 kW in 2 seconds. Maybe I'm wrong here. However, why wouldn't mode 8 not simply have a charging and discharging register that tells how much watts are allowed in both ways and keep the "allow discharge to grid" flag.

That would allow much more control to mode 8 and could simulate Feedin priority and self consume.

infradom:
Instead of throttling down the PV power, I am even thinking of sending my excess PV power into my geothermal underground loop, so that the underground warms up ; my heathpump can than recover that energy in the autumn/winter. My geothermal heatpump installer believes this idea will not be worth the effort; so I have not done it yet.

HA-Supi:
@infradom : nea started with a fork, the link is a new approach, mostly for better battery charge (not always 100% after 1-2h)

I would say, remote control is a complex functionality, so keep it simple. 
Integration in wills solax addon should be secure. 
I wish you good luck with remote control.

infradom:
> The best HA automation couldn't mitigate a house load that jumps from 300 W to 2 kW in 2 seconds

I would not worry about these 2 seconds mismatch, my experience shows that when I want to have no grid export during a negative price weekend day, at the end of the day, I may have exported less than 0.2 kWh . Same for grid import during summer, almost nothing and my system runs probably more than 4 hours a day in remotecontrol power loop mode.

rosenrot00:
@wills106 is there a reason this hasn't been merged yet? Any remarks? I didn't change any entity names so this is "only" cosmetics.

@infradom As far as I can read, you gave your blessing already? :)

Happy to change or cancel if it isn't wanted.

wills106:
Not really kept up with the conversation on this PR. So wasn't sure if it needed merging yet or not.

infradom:
sorry for the late reaction, indeed, it was fine form me ! Thanks

wills106:
I'll hold off releasing it until the morning as already had two releases today.

infradom:
If it is merged already, I can test it this evening, before you release it tomorrow or whenever

infradom:
Does not work yet: I get an error "suggested_display_precision" does not exist (or something similar) line 679
So please do not release yet. I cannot resolve this this evening.. other tasks

rosenrot00:
@infradom you are right, I missed the change in the const.py. Did it. https://github.com/wills106/homeassistant-solax-modbus/pull/1609

infradom:
Tested, ok now. My remotecontrol automations (mode8) continue to work as expected.
thanks

ralfna:
Hi all, just saw here the discussion about remote control and I am always looking for advice and sample scripts/automations about remote control. So I like to ask is there a good source to learn and to get better with this. I did my own automation, but I had some trouble with it, thanks for any advice. br Ralf

infradom:
@ralfna : 
For the time being, I can only refer to the readthedocs documentation about remotecontrol [mode1](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control) and remotecontrol [mode8](https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode8-modbus-power-control) . I have added a couple of example automation scripts in these pages (some may be outdated).
Nowadays, I almost exclusively use mode8 in my summer automations. My winter automations are still using mode1, but I might adapt them to mode8 one day.

Zdenek45:
Hi.
Is description of entity "remotecontrol_power_control" as "Remotecontrol Power Control (mode 1-9)" correct? I think this is Select only for Mode 1.

>     SolaxModbusSelectEntityDescription(
>         name="Remotecontrol Power Control (mode 1-9)",
>         key="remotecontrol_power_control",
>         unit=REGISTER_U16,
>         write_method=WRITE_DATA_LOCAL,
>         option_dict={
>             0: "Disabled",
>             1: "Enabled Power Control",  # battery charge level in absense of PV
>             11: "Enabled Grid Control",  # computed variation of Power Control, grid import level in absense of PV
>             12: "Enabled Battery Control",  # computed variation of Power Control, battery import without of PV
>             110: "Enabled Self Use",  # variation of Grid Control with fixed target 0
>             120: "Enabled Feedin Priority",  # variation of Battery Control with fixed target 0
>             130: "Enabled No Discharge",  # missing HL from grid
>             # 2: "Enabled Quantity Control",
>             # 3: "Enabled SOC Target Control",
>         },
>         allowedtypes=AC | HYBRID | GEN4 | GEN5,
>         initvalue="Disabled",
>         icon="mdi:transmission-tower",
>     ),

infradom:
@Zdenek45 : I believe you are right, this currently only applies to mode 1 and its computed submodes (with mode2 and 3 commented out as not yet implemented)
It certainly cannot be used for mode 8 and 9 that use a different mechanism.

infradom:
yes, you are right, this theoretically only applies to modes 1-7, and only mode 1 and its submodes is supported.
So mode 8 and 9 should not be included in the list.
If we ever implement modes 2-7, they probably belong to this list.
Mode 8 and 9 use a different register set.
Thanks. Awaiting @rosenrot00 's opinion on this

rosenrot00:
@Zdenek45 you are absolutely right. So renaming it to mode 1 is the right thing to do. @infradom I would suggest not adding mode 1-7 as the rest isn't implemented? I have only played with mode 8/9 so far this is why I'm not so familiar with the other modes yet. Sorry for the confusion.

https://github.com/wills106/homeassistant-solax-modbus/pull/1623

PR #1568, opened 2025-08-28
Implement dependency mechanism for sensors with value_functions

infradom:
- Introduces depents_on attribute for computed sensors. This assumes that plugins declare the dependencies with such a stratement

- Some computed entities did nog get any value after initial installation. The entities became available after a first HA restart or after an integration reload. This should be solved now

- Continued enhanced debug output, showing the inverter name
- adapted solax, solis, sofar, growatt pluigin to use the depends_on method. This work may not be complete
-


wills106:
Is this ready for merging now?

Edit:
Just looked at the code, it looks like Copilot is making suggestions. Do you need addressing or are they false suggestions?

infradom:
I committed some of copilot's suggestions and fixes, but not all of them.  I did not like the suggestion to remove some brackets, especially since the version without the brackets caused problems.

PR #1577, opened 2025-08-31
Introduce auto bad register handling

rosenrot00:


infradom:
@rosenrot00 : you mentioned that the mechanism does not work in some cases
I have been trying to reproduce errors, but in each test case, your solution did remove the failing register correctly. 
So my latest PR added some extra output to find out why we still see failing registers in sofar issue #1503 
As soon as @wills106 releases this, we hopefully can proceed testing further with real world data

My tests introduced only one failing register; maybe I should test also with multiple (consecutive) failing registers

rosenrot00:
Thanks for taking the time to test. I asked one user to enable an entity that was a bad register for him, and he still reported that entities didn't read; this is why I blamed our solution. Adding more debug hopefully helps us to find out. I suggest that we actively approach users with bad registers to get debug logs. I think some Sofar users are still facing issues.

infradom:
Indeed, as soon as the new code is released, I or we can approach our users.

infradom:
The 2 sofar logs I received since release 2025.9.7  are interesting. It may take some time for me to analyze, but they should help me understand what happens. Unfortunately, people do not wait long enough after startup, so I fear last part of the log is missing (or there was no final re-check phase)

One log seems to use 2 devices (same hub name): inverter and battery pack

infradom:
Just a remark: the pymodbus exception output looks a bit strange to me, the output seems to include previous requests also. I did not reallize that at first. So I believe we have to look at the last line(s) only.

```
2025-09-05 09:44:54.767 ERROR (MainThread) [pymodbus.logging] No response received after 1 retries, continue with next request
>>>>> send: 0x0 0x1 0x0 0x0 0x0 0x6 0x1 0x3 0x4 0x45 0x0 0x7
>>>>> recv: 0x0 0x1 0x0 0x0 0x0 0x11 0x1 0x3 0xe 0x53 0x53 0x32 0x45 0x53 0x31 0x30 0x34 0x4e 0x35 0x53 0x34 0x34 0x35 extra data: 
>>>>> send: 0x0 0x2 0x0 0x0 0x0 0x6 0x1 0x3 0x4 0x4 0x0 0x57
>>>>> Repeating....
```

the first request 01 seems to be the old request that requested the serial number
the second request 02 looks fine, but receives no response

I analyze now why the bisect does not return a failing register

rosenrot00:
Interestingly in this log from here https://github.com/wills106/homeassistant-solax-modbus/issues/1605#issuecomment-3257859563 I can see that bisect identified a bad register but it doesn't probe it afterwards.

```
2025-09-05 12:13:02.956 ERROR (MainThread) [custom_components.solax_modbus] Error: device: 1 address: 1051 -> Modbus Error: [Input/Output] No response received after 1 retries, continue with next request
2025-09-05 12:13:02.956 INFO (MainThread) [custom_components.solax_modbus] sofar: probe not fully ok: depth 2/10 len: 6
2025-09-05 12:13:02.956 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x41b-0x422  
2025-09-05 12:13:03.144 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x445-0x45b  
2025-09-05 12:13:03.442 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x484-0x4bd  
2025-09-05 12:13:03.586 DEBUG (MainThread) [custom_components.solax_modbus] sofar: scan_group timer initiated refresh_modbus_data call - interval 1
2025-09-05 12:13:03.586 DEBUG (MainThread) [custom_components.solax_modbus] sofar: skipping poll – initial probe not done yet
2025-09-05 12:13:03.812 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x584-0x589  
2025-09-05 12:13:04.069 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x5c4-0x5c5  
2025-09-05 12:13:04.324 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x684-0x694  
2025-09-05 12:13:04.586 DEBUG (MainThread) [custom_components.solax_modbus] sofar: scan_group timer initiated refresh_modbus_data call - interval 1
2025-09-05 12:13:04.587 DEBUG (MainThread) [custom_components.solax_modbus] sofar: skipping poll – initial probe not done yet
2025-09-05 12:13:04.604 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x1023-0x1025  
2025-09-05 12:13:04.864 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x488-0x4a4  
2025-09-05 12:13:05.174 INFO (MainThread) [custom_components.solax_modbus] sofar: probing holding 0x586-0x58a  
2025-09-05 12:13:05.431 INFO (MainThread) [custom_components.solax_modbus] sofar: initial bisect found no suspect registers.
```
It seems not probing 0x404 which later on fails.

```
2025-09-05 12:13:17.559 DEBUG (MainThread) [custom_components.solax_modbus] sofar: ** trying to read holding block 0x404 previous res:True
2025-09-05 12:13:17.562 ERROR (MainThread) [pymodbus.logging] No response received after 1 retries, continue with next request
```

infradom:
Bisecting leads to success on each smaller block. Maybe we should reduce the block size for sofar, it is a plugin-specific setting
It is not register 0x404 that fails, but block 0x404 fails if it is read in a large block
If it is split up, 0x404-0x40F succeed (and all sub-blocks succeed)

the @darkrain-nl log shows that the initial probing has no bad registers anymore (after splitting). Same for the final recheck

infradom:
I would suggest to set the maximum blocksize for sofar to 0x30 (since a 0x40 length block still fails)
This may not bring the solution, but you never know ...

infradom:
@darkrain-nl : if you feel capable of modifying one line of code at the end of plugin_sofar.py and generate a new log, we would not have to wait for a new testing release. The line somewhere at the end should become:
```
    block_size=48, # was 100
```

darkrain-nl:
> @darkrain-nl : if you feel capable of modifying one line of code at the end of plugin_sofar.py and generate a new log, we would not have to wait for a new testing release. The line somewhere at the end should become:
> 
> ```
>     block_size=48, # was 100
> ```

Sure thing will do :-)

wills106:
I'm sure Sofar used to have a small block_size and not 100?
As we have always had issues with Sofar Inverters.
I think the docs mention only 60 registers at a time and we reduced it down to 40 or 50 to be on the safe side.

darkrain-nl:
With block_size=48, # was 100 it works for me, but I will still make a debug log for you...

rosenrot00:
@infradom nice catch. So are all these bad registers a hoax?

infradom:
> So are all these bad registers a hoax?

could be, but in some cases, they are real, so your work has real value in my opinion and will save us a lot of time

darkrain-nl:
new log here https://github.com/wills106/homeassistant-solax-modbus/issues/1605#issuecomment-3257983544

rosenrot00:
> could be, but in some cases, they are real, so your work has real value in my opinion, and will save us a lot of time

Could you look into this changed approach with your tests? It should not only check if there is a reading issue, but also replace large blocks that fail with bisect blocks that suddenly work. Again, untested so be patient with me :) I was working on this offline but I thought it might not be needed. If it would work, we could enhance debug outputs if needed. https://github.com/rosenrot00/homeassistant-solax-modbus/blob/auto-bad-register/custom_components/solax_modbus/__init__.py

infradom:
@rosenrot00 I will have a look at your new version; does it contain already my enhanced info/debug logging (already in @wills106 main line) ? I really find this info logging very useful to track what your algoritm does.
I dont like to have debug logging on all the time, so this is why I dont mind to have some info logging during startup phase

rosenrot00:
yes it is the latest released version.

infradom:
Thanks, fairly complicated piece of code on first sight.
I will try to understand it; testing may be simpler than understanding
I wonder if it is worth the complexity, but this is just an initial reaction ...

infradom:
update: nice code, I understand it now.
I would suggest to add some more info logging, that it becomes clear that we are applying a time consuming action during normal polling. I can imagine that some error conditions will activate this mechanism on every polling cycle. This should at least be visible in the logs (but contradictory, we also need to reduce logging after startup phase)
I will make a proposal for slightly enhanced logging (that makes analysis easier and that makes clear if there are recurrent problems)
I may not have time immediately to do this and to do some testing. I need to give some non-github related todos priority.

rosenrot00:
To be the building blocks should only happen once and thus the bisect should only happen initially. However, you have much more knowledge about it. Feel free to add logging info and hope that all this helps to mitigate bad register readings.

PR #1579, opened 2025-09-01
Added missing depends_on for buttons

rosenrot00:


infradom:
Although unlikely that selects and numbers will need depends_on, I believe it makes sense to add it also

infradom:
Added this for select, number and switch also. Thanks

PR #1581, opened 2025-09-01
solinteg: implement depends_on

gec75:


infradom:
thanks for adding this, I did not check the solinteg code for value_functions aparently. Can you add solinteg in the title?

PR #1587, opened 2025-09-02
Modbus compatibility layer  - developed by @rosenrot00, initial tests by @infradom

infradom:
This PR with code base from @rosenrot00 aligns our integration with the most recent pymodbus API.
It adds a very lightweight compatibility layer.
The compatibility layer is intended to support also older pymodbus versions, but emulates the new API
Payload.py can then later be removed once proven it is not needed anymore.


Saentist:
https://github.com/wills106/homeassistant-solax-modbus/issues/1594
error in ENDIAN

PR #1588, opened 2025-09-02
Modbus compat: const.py should not contain Endian, copy to payload 

infradom:


infradom:
payload now contains a copy of Endian; could not import from compat due to a dependency loop

PR #1599, opened 2025-09-04
Solinteg fix after pymodbus changes

rosenrot00:


rosenrot00:
@wills106 please also merge this one. We have used the wrong variable name :) This PR can't do any harm and just makes things better even if it wouldn't work (even though it should).

PR #1604, opened 2025-09-04
Remove dependency from payload.py

rosenrot00:
We have a new pymodbus compat layer and we don't need payload.py anymore. If someone has such an old pymodbus, she/he should update it. I hardly doubt anyone uses such an old pymodbus with the latest solax integration.


infradom:
Small remark: The error message  is  a bit strange:
if BinaryPayloadBuilder cannot be found, it means that the user is using a more recent version of pymodbus, so not the very old ones. So a local copy was added afaik.
Maybe I misunderstand something. Does not matter much. 

I have the impression we now support BinaryPayloadBuilder from very old pymodbus versions (with BinarayPayloadBuilder) the more recent ones dont need it as we use convert_).

rosenrot00:
I tried to put it together how I see it - but please feel free to comment on that. What you describe is what was done with payload.py, which was not future-proof, while now we follow the inverse approach, which hopefully is future-proof.

**Fast-Path (≥ 3.8, including 3.11.1):**
When pymodbus provides the convert_* helpers with word_order support (e.g. 3.8+, 3.9.x, 3.10.x, 3.11.1), we call them directly. This is the minimal-overhead path.

**Fallback (Legacy, ≤ 3.7):**
If the helpers are missing or do not support word_order (typical in 3.6/3.7), we fall back to the BinaryPayloadBuilder/BinaryPayloadDecoder API. Compatibility is maintained as long as the pymodbus.payload module is available.

**Not Supported:**
Builds without usable convert_* helpers and without the payload module are not supported. In this case we raise a clear error message and instruct the user to install a compatible pymodbus release (≥ 3.8 preferred).

infradom:
Sorry for wasting your time; I just  wanted to point out that the error message will probably never occur (and if it does, it is for a recent pymodbus). It does not really matter; I am probably wrong

rosenrot00:
No, you are very welcome. Discussing topics allows everyone to gain different perspectives - it is never a waste of time.

If ever we see this error message on a newer pymodbus version, I will be proven wrong.

PR #1609, opened 2025-09-05
Remote control naming & precision

rosenrot00:


infradom:
Tested after this update; looks ok now.
My remotecontrol (mainly mode8) automations continue to work as expected.
 Thanks

PR #1630, opened 2025-09-30
Added polling protection

rosenrot00:
I added functionality that a new poll doesn't start before the previous one was finished.

I'm hunting down a bug that leads to negative houseload values when I decrease the fast polling time from 5 to 3 seconds.

I see polling times of about 300ms but this might be handy in case polling takes longer on slower machines or when there are connectivity issues.

@infradom you know the polling better than I do, maybe you can double check if what I did make sense? On a side node, currently there seems not to be an initial polling, meaning slower polled data show empty until the slow interval measurements are done. Wouldn't it make sense to have all poll groups read initially so that the integration doesn't show empty values for 15-60s (or whatever the slower intervals are)?

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -8,6 +8,7 @@ import importlib
 import json
 import logging
 from time import time
+import time as _mtime
 from types import ModuleType, SimpleNamespace
 from typing import Any, Optional
 from weakref import ref as WeakRef
@@ -131,7 +132,9 @@ PLATFORMS = [Platform.BUTTON, Platform.NUMBER, Platform.SELECT, Platform.SENSOR,
 empty_hub_interval_group_lambda = lambda: SimpleNamespace(
             interval=0,
             unsub_interval_method=None,
-            device_groups={}
+            device_groups={},
+            poll_lock=asyncio.Lock(),
+            pending_rerun=False,
         )
 empty_hub_device_group_lambda = lambda: SimpleNamespace(
             sensors=[],
@@ -534,6 +537,7 @@ class SolaXModbusHub:
         return key
 
 
+
     # following function is the added_to_hass callback for sensors, numbers and selects
     @callback
     async def async_add_solax_modbus_sensor(self, sensor: SolaXModbusSensor):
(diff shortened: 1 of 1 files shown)


infradom:
I have not checked your change yet,  but indeed it would be good to start the first slower polling loop immediately.
I was not the author or the multiple polling cycles, but I will have a look at your change(probably this evening).

rosenrot00:
Thank you, I tried to implement an initial polling but it didn't work out initially polled data got overwritten by the fast polling cycle. Maybe you have an idea.

infradom:
Strange, I do not understand how a fast register can be overridden. Even if the range blocks overlaps, the register itself should not be treated. But I guess I misunderstand your remark

rosenrot00:
I didn't put much time in understanding it. I just saw that the homeassistant entries are getting filled and just afterwards become empty again. I just talk from plain observation, without having spent 5 minutes in understanding it. I also didn't want to overload the PR also with other mechanisms beside the poll protection.

infradom:
I had a quick look, it looks fine. I intend to test it on my system later this evening.

rosenrot00:
> I had a quick look, it looks fine. I intend to test it on my system later this evening.

Great thank you!
I’m seeing negative houseload on my installation from time to time. Also before this change. I will add a debug output in cases house load is negative to track down which value might be wrong. It happens for one poll and is gone on the next always. Not super critical but annoying. Maybe it is firmware related as I didn’t see it before the latest 1.56/1.54 firmware.

infradom:
I assume these negative values are rather small, it is possible that inverter_power, pv_* and measured_power are sampled at a different time, so that fluctuations result in short-lived negative values. I saw these also from time to time, but was never worried about this, due to the way this value is computed from different parameters. Measured_power is gathered through another modbus, so small delays are possible.

infradom:
I examined further: the negative values are not always small. This evening, my automation discharged my battery with 4200 Watt at 18:30.
At that moment, I see a short lived negative value of 4000 Watt:

House load graph: 
net1

Battery charge graph:
neg2

So this is just a data collection timing issue in my opinion.

rosenrot00:
I will double check with usage and delivery values. If your assumption is correct these should happen on rapid changes of the related variables.

rosenrot00:
My observation confirm your assumption.

image

I'm just wondering if it would make sense in that case to clamp houseload to 0 or even leave the last value intact. Not sure if for remote control that would be an advantage as not doing completely off things?

On a second thought, we do currently split the reading blocks right? Could it be that this splitting is strengthening to possibility of these value missmatch to happen as if we would read the entire block in one go, the MCU would copy all values to a buffer at the same time. Do you know if the variables to calculate house load are found within the same block?

infradom:
It is unclear to me if clamping to 0 would enhance convergence or make it worse. I could test this for a while.
Regarding your second point: I dont believe the splitting in blocks is the key issue, the time differences are in my opinion already in the firmware of the inverter. Also: the registers needed for house_load are 0x02 and 0x46 ; they are in the same block (for my Gen4): 

```
2025-10-02 09:27:03.425 INFO (MainThread) [custom_components.solax_modbus] SolaX - interval 3s: adding input block: 0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x9, 0xa, 0xb, 0x14, 0x15, 0x16, 0x1c, 0x46
```

Personally, I dont think that using  (fast - slow) differentiated polling was a good idea for Solax. Unless we do what you suggest and read bigger blocks, but this will make the algoritm more complicatad (bad register detection). And since our house_load registers are it the same block, it won't help.
I myself do not mind seing a negative value from time to time, it shows that things may change rapidly over time.

rosenrot00:
Thanks for looking into it. I'm aligned with your opinion. These negative readings don't do any harm for now.

This PR is still valid as it makes sure that if one sets the polling period to 1s and polling takes longer that it polls as fast as it can.

infradom:
yes, I completely agree that the PR makes sense. I am even thinking of making a couple of sensors that report the maximum, average, minimum polling time (for each of the scan groups)

infradom:
I have run this code for a couple of days, no problems observed during this period.

Issue #1648, opened 2025-10-11
[Bug]: Sensors keep their old value when communication is lost


Zdenek45:
### Describe the bug

This issue was initially discussed in posts:
 _Originally posted by @Zdenek45 in [#1641](https://github.com/wills106/homeassistant-solax-modbus/issues/1641#issuecomment-3392337617)_
 _Originally posted by @infradom in [#1641](https://github.com/wills106/homeassistant-solax-modbus/issues/1641#issuecomment-3393064121)_

In case the communication with inverter is broken, all entities remains in last state instead of change to Unavailable state.

### Integration Version

2025.10.2

### Homeassistant core version

2025.10.1

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H34A15

### Inverter firmware versions

DSP v1.46 ARM v1.44

### Connection Method

Waveshare RS485 TO POE ETH 

### Dongle firmware

NA

### Detailed Error Log

Logs uploaded in other post:
https://github.com/wills106/homeassistant-solax-modbus/issues/1641#issuecomment-3392337617

infradom:
I am attaching the log file you provided in the other thread (interuption created manually): 

[home-assistant-b.log](https://github.com/user-attachments/files/22865432/home-assistant-b.log)

I believe I have a solution ready, but cannot upload it yet as I want my other PR to be merged first, otherwise the changes get mixed and it is difficult to pinpoint source of potential future problems.

If you are not afraid of editing a couple of lines of source code, you could test my potential solution. Let me know if you want me to explain the simple change I would like to test.

Zdenek45:
You can point me to changes. I can test it.

infradom:
In __init__.py around line 1036 : please remove the "not" in this piece of code.
I still need to test it myself, but the code below that block was meant to invalidate data in case of read errors: (it may not have been executed as expected in the past due to a faulty 'not')
```
            _LOGGER.debug(f"{self._name}: failed {typ} block {errmsg} start 0x{block.start:x} {firstdescr.key} ignore_readerror: {firstdescr.ignore_readerror}")
            if (firstdescr.ignore_readerror is not False):  # ignore block read errors and return static data
                _LOGGER.debug(f"{self._name}: failed block analysis started firstignore: {firstdescr.ignore_readerror}")
```

should become: 

```
[...]

Issue #1641, opened 2025-10-08
[Bug]: Integration causing HA core to reboot.


dougle03:
FIX: Check you don't have any additional devices in the integration that have no or unavailable entities. 

### Describe the bug

I gave ChatGPT my logs and this is what it's saying.

*********************************
At 18:54:28, Home Assistant throws a fatal exception:

Error doing job: Task exception was never retrieved (None)
Traceback (most recent call last):
  File "/config/custom_components/solax_modbus/__init__.py", line 568, in _refresh
    agg_res, updated_sensors = await self.async_refresh_modbus_data(...)
TypeError: cannot unpack non-iterable NoneType object

That’s a coding bug inside your solax_modbus custom integration, not a configuration issue.
It tries to unpack two return values but gets None, which raises an unhandled exception in the event loop.
That exception stalls the main loop → HA watchdog (or Docker’s restart policy) kicks in → forced reboot.

So right now, solax_modbus is crashing Home Assistant.

🧩 What led up to it

The same component failed earlier:

Modbus Error: [Input/Output] No response received after 1 retries
SolaX: reading serial number from address 0x0 failed

Then it immediately hit the unpacking error.

### Integration Version

2025.10.1

### Homeassistant core version

2025.9.3

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

XM4A25

### Inverter firmware versions

 DSP 17 ARM 20

### Connection Method

Waveshare eth adapter

### Dongle firmware

Not wifi

infradom:
This chatgpt info is irrelevant to me./
It would make more sense to give us your debug log or describe whether this is a new installation that fails or some more context information.  See General FAQ on how to generate a debug  log.

I myself tend to ignore bug reports without significant (debug) log. Debug logs contain much more interesting information (like modbus version ...)

infradom:
@rosenrot00 : this may be related to your latest change, but I miss context information as explained in my previous message.
On first sight, it is just a failing modbus communication (first time connection ?)

rosenrot00:
@infradom how do you know? I would need more context to pass such a bold statement to you :D

@dougle03 please provide log and tell with which older version it still works.

dougle03:
I've had it running for over a year. But noticed rebooting of HA (Core in docker) over the last maybe week or so.
Not a new installation, and has been fine.
Not done any regression to see if it works with older versions.
[...]

PR #1650, opened 2025-10-13
ignore_readerror problem: entities kept their old value when communication was lost

infradom:
reported first off-topic in thread #1641 by Zdenek45. Deserves a separate issue thread. issue #1648


infradom:
Also implemented implemented in #1651

PR #1654, opened 2025-10-14
Introduce export-first battery limit submode

rosenrot00:
I added a VPP 8 submode that feeds-in first and charges the battery with the PV surplus. The battery charging power can be dynamically adjusted by using a new control "Export-First Battery Charge Limit (mode 8/9)"

This submode enables running a logic similar to Solax’s Datahub within Home Assistant: which analyses the solar forecast and, when sufficient sunlight is expected, prioritises grid feed-in before charging the battery to maximise feed-in benefits. By allowing the battery charging power to be adjusted, it also makes it possible to slow down charging when plenty of solar energy is available – helping to maintain battery health.


infradom:
Maybe we should also adapt the mode8 documentation (reead_the_docs). Does not seem easy to explain.

One remark: it may help to maintain battery health, but for people on a dynamic (15 minute) contract, we will probably be  exporting energy when it is worth (almost) nothing (at times when there is plenty of solar). So I continue to prefer to export my energy in the morning and evening, when prices are at the highest level. I have a fixed limit on the battery charge max current to avoid excessive load on the battery.

But every use case is different, I just wanted to state I cannot test it (in depth) since the algoritm does not match my strategy for solar. I am sure people with a less dynamic contract will prefer to use your model. 
It would be nice to be able to limit PV to avoid export (but then it may become too complicated to understand; the negative prices submode already handles this in a way that does not consider battery wear (except when nearly full))

I also have an open PR for mode_8 gridcontrol, so it is nice to see the VPP submode set growing...

rosenrot00:
I'm completely with you on your remark. I see a lot of different use cases and I think we can provide different submodes for them. At least this is my current understanding, there is no "one fits all" strategy. As I'm not on a dynamic price plan, I want to export as much to the grid as possible. If the forecast tells there is enough solar (beyond the grid export limit) available, I turn this submode on and I gain 15% more energy export per day as I'm maximising my export. Besides I limit the battery charge power if I have the day long to charge the battery.

I'm currently looking into dynamic price plans but the issue is that these on the solar export give very little per kWh. Which is to be expected as the pricing for import reflects my own solar production volume, which means in a matter effect that prices for export are almost zero when I generate the most solar power. I did some calculations that state I can provide 4MWh per year to the grid which equals to 400 Euro. When I look into dynamic pricing, I see that during winter the cheapest dynamic prices are on average (adding up several hours with lowest prices) 0.03 Euro below the usual pricing model. I would massively loose on the grid export and I'm not sure all the complexity is worth gaining 300 Euro a year. I would have to cycle the batteries much more during, which would reduce their lifespan. My battery capacity is only big enough to bridge 3-4h during the winter months, which is adding complexity as more capacity would make the system more flexible. So, still not sure how I can make sense out of it.

I added a description for this new submode to the mode8 documentation

infradom:
I agree with your analysis: dynamic prices are not proven to be more economic. I have 22kWh of battery (theoretical max), and in the spring/summer, by selling in the morning/evening I tend to earn almost enough to cover my winter consumption (that includes heating with a  geothermal heatpump). Winter prices are very unpredictable, so the danger is certainly there. Last winter, I  probably was lucky (import prices remained reasonable at night).

One remark: I dont loose on the grid export, my automation goes to negative_price mode when prices are negative.
Yes indeed, I almost cyle my batteries once a day. This will certainly not be good for the official 6000 cycles lifetime.

wills106:
Not sure if I have merged the previous PR's wrong, but there seems to be quite a large conflict?
When I go on the Files changed tab it looks ok, but then on the Resolve conflicts button it doesn't look so great?

rosenrot00:
@wills106 I will have a look.

wills106:
I have an option to rebase and merge. It might be something as simple as that. But I don't want to go and mess things up even worse as I'm not the greatest with GitHub.

rosenrot00:
Give it a try. In the worst case I add the changes to the latest version. It is a rather isolated change so no worries.

infradom:
my recent changes are also easy to verify (and to redo in case they get lost)

rosenrot00:
I think I solved it. It was because @infradom made changes at exactly the same spots I did. I made a minor change and now it seems to be ok.

PR #1663, opened 2025-10-20
Fix Docs Mode 8 Options List Capitalisation

TCWORLD:
"BAT" is in caps in the integrations select options, so should be capitalised in the documentation.

This also corrects the "Edit on Github" link to point to the correct branch. See: https://www.mkdocs.org/user-guide/configuration/#edit_uri for details. Basically the default for `edit_uri` is `edit/master/docs/`, however as this repo uses `main` as the branch name, need to set the URI to `edit/main/docs/`. Otherwise you get a nice 404 error when clicking the edit link.


infradom:
Thanks !!

Issue #1658, opened 2025-10-18
[Bug]: Remotecontrol causes oscillation switching from forced zero battery power to disabled (Solax X1-IES Gen5)


TCWORLD:
### Describe the bug

Kind of a weird bug, not sure if its in this integration, or in the inverters firmware, so bear with. It's very intermittent, so still trying to debug what the sequence of events that triggers it.

Basically I'm using the [predbat](https://github.com/springfall2008/batpred) integration to control charging/discharging of the battery using the remotecontrol interface in the SolaX Modbus integration. For the most part this is working well.

However last night (same thing happened at the same time two weeks ago), when exiting the remote control mode (selecting mode of Disabled), the inverter started oscillating between Import and Export every minute. See screenshot below. 

Once it goes into this weird mode, even flipping the main breaker for the inverter off for a few moments and and back on again doesn't recover it (goes back to oscillating when the breaker comes back on). I found both times the only way I could stop it doing this was to issue a "Force Discharge" remote control command (has to be high enough power, e.g. 2000W) for a couple of minutes before issuing the "Disabled" command again. 

<img width="1427" height="935" alt="Image" src="https://github.com/user-attachments/assets/6f026310-48f1-46b9-b896-4cbe40570ecb" />

Once it goes into this weird mode, even flipping the main breaker for the inverter off for a few moments and and back on again doesn't recover it (goes back to oscillating when the breaker comes back on). I found both times the only way I could stop it doing this was to issue a "Force Discharge" remote control command (has to be high enough power, e.g. 2000W) for a couple of minutes before issuing the "Disabled" command again. 

<img width="1158" height="935" alt="Image" src="https://github.com/user-attachments/assets/43cf39e3-1a23-4fa7-8503-82233bf8fa43" />

Some extra info:
* The first time it was switching from "Enabled No Discharge" to "Disabled"
* The second time it was going from "Eanbled Battery Control" with a power of 0, to "Disabled"
* Both times this has happened, it has been at just after midnight.

It's entirely possible that there is an issue in the inverters firmware itself, but wanted to check if anyone has seen similar behaviour and whether it's possibly something up with the issuing of commands by this integration. If not, I will contact SolaX support and see if they've seen it or if they can upgrade the firmware.

### Integration Version

2025.10.4

### Homeassistant core version

2025.10.02

### Inverter brand

Solax X1-IES Gen5

### Plugin used

plugin_solax.py

### Serial prefix

H5602AK

[...]

PR #1667, opened 2025-10-21
Solax VPP: mode1 and mode8: provide timeout parameter correctly

infradom:
Previous versions did not supply the timeout parameter of the write_multi in the correct location in the payload
This PR should correct this for both mode1 and mode8.


infradom:
maybe issue #1658 is related to the misuse of timeout in the past. still unsure

Issue #1640, opened 2025-10-07
[Bug]: Solax X3-Hybrid / House Load calculation wrong


Goracc:
### Describe the bug

Hey all,

I've been upgraded to the latest firmware on 09.09.2025. Since then I've noticed that the House Load calculation is always wrong as soon as I have any Solar/Battery usage. If it's Grid only then it seems right and matches what e.g. EVCC is showing me. It's always off by like 50-80 W. Before the Firmware upgrade everything was looking right.

<img width="501" height="373" alt="Image" src="https://github.com/user-attachments/assets/06503d9a-2403-4a0a-8508-02fb38ed6c50" />

<img width="508" height="366" alt="Image" src="https://github.com/user-attachments/assets/80bb5196-93fc-422a-886c-8b93e057c3f3" />

I've been in contact with Solax about this and they've sent me the latest Modus documentation. I've attached it to this post.  

[(Solax)Hybrid X1&X3-G4 Modbus RTU V3.47-English_250115.pdf](https://github.com/user-attachments/files/22735974/Solax.Hybrid.X1.X3-G4.Modbus.RTU.V3.47-English_250115.pdf)

### Integration Version

2025.10.1

### Homeassistant core version

2025.10.1

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H34A10

### Inverter firmware versions

DSP v1.56 ARM v1.54

### Connection Method

Solax Pocket WiFi 3.0

### Dongle firmware

3.019.01

infradom:
Thanks for the document, it may be more recent than the one I was using.
I am not sure there is much we can do: 
House_load is not provided by solax as a register, so we have to compute an approximation (derived from measured_power and inverter_power).
I believe it is normal that it can be off by 80-100 Watt.
There are multiple interpretations of house load possible: 

- just the house load
- the house load including the energy consumption of the inverter

The more solar or battery usage, the more losses in the inverter, so this can vary between 20 and 200 Watt.

We could try to improve this with empiric corrections or with Machine Learning (sometimes called AI), but it did not seem important enough until now.

I will examine the protocol document to see if there are new registers that provide a more reliable way.
There is also a disabled entity house_load_alt, that computes house load in a different way.

See the source code for value_function_house_load and value_function_house_load_alt.
```
def value_function_house_load(initval, descr, datadict):
    return (
        datadict.get("inverter_power", 0)
        - datadict.get("measured_power", 0)
[...]

PR #1679, opened 2025-10-25
Add Gen4 meter direction configuration support

fxstein:
This PR adds meter direction configuration support for Gen4 inverters. Gen3 already had meter direction support at registers 0x116/0x117, but Gen4 uses different registers (0x010B/0x010C).

## What's Added

**Two configuration select entities:**
- `select.solax_X_meter_1_direction` (register 0x010B)
- `select.solax_X_meter_2_direction` (register 0x010C)

**Options:** Positive (0) or Negative (1)

**Two internal sensors:**
- `sensor.solax_X_meter_1_direction` (register 0x010B, internal)
- `sensor.solax_X_meter_2_direction` (register 0x010C, internal)

## Key Features

- **Safe by default:** Entities are disabled by default (`entity_registry_enabled_default=False`)
- **User opt-in:** Users must explicitly enable these entities if needed
- **Non-breaking:** Does not modify any existing entities or behavior
- **Gen4 specific:** Only appears on Gen4 inverters (allowedtypes = AC | HYBRID | GEN4)
- **Configuration category:** Properly categorized for UI organization
- **Icon:** Uses `mdi:meter-electric` for clarity

## Testing

Tested with 3x SolaX X3-Hybrid-15kW Gen4 inverters (serial H34A15K):
- ✅ Entities created successfully when enabled
- ✅ Read actual values from inverter registers
- ✅ Options (Positive/Negative) available and functional
- ✅ No errors in logs
- ✅ Existing entities unaffected
- ✅ Integration loads and operates normally

## Why This Matters

Meter direction configuration is needed for proper power flow calculations when meters are configured with positive meter flow direction instead of the default negative direction. This PR provides the foundation for accurate power measurements regardless of meter configuration.

## Compatibility

- No breaking changes
- Backward compatible
- Only affects Gen4 inverters with these entities enabled
- Gen3 meter direction (0x116/0x117) remains unchanged

## Note on Previous PR

This supersedes the previously reverted PR #1082. This version is intentionally conservative:
- Only adds meter direction entities
- Does not modify `measured_power` sensor
- Does not add power correction logic (will be separate PR)
- Ensures no breaking changes for existing users

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1998,24 +1998,26 @@ SELECT_TYPES = [
     SolaxModbusSelectEntityDescription(
         name="Meter 1 Direction",
         key="meter_1_direction",
-        register=0x010B,
+        register=0x00A4,
         option_dict={
             0: "Positive",
             1: "Negative",
         },
         allowedtypes=AC | HYBRID | GEN4,
+        entity_category=EntityCategory.CONFIG,
         entity_registry_enabled_default=False,
         icon="mdi:meter-electric",
     ),
     SolaxModbusSelectEntityDescription(
         name="Meter 2 Direction",
         key="meter_2_direction",
-        register=0x010C,
+        register=0x00A5,
         option_dict={
             0: "Positive",
             1: "Negative",
         },
         allowedtypes=AC | HYBRID | GEN4,
+        entity_category=EntityCategory.CONFIG,
         entity_registry_enabled_default=False,
         icon="mdi:meter-electric",
     ),


wills106:
Shouldn't the selects be writing to A4 & A5?
Just checked the Gen4 - Gen6 docs and all 3 say to write to A4 & A5.

image

fxstein:
Good catch. I was only focused on reading the values. Missed the writes. Will fix.

Also looking for the very latest docs - mine are a bit our of date.

wills106:
Have a look at #1640

fxstein:
ok write registers are now correct. Also moved to config section as this very much feels like a config thing and not something that one would change frequently.

PR #1687, opened 2025-10-27
Add support for 15...24KTLX-G3.

mldebian:
This works nicely on my 24KTLX-G3.

The only nit I have is that Load consumption today is actually hourly production, and resets every hour, but it seems modbus describes it as such.

### DIFF
--- a/custom_components/solax_modbus/plugin_sofar.py
+++ b/custom_components/solax_modbus/plugin_sofar.py
@@ -682,7 +682,7 @@ SELECT_TYPES = [
             0: "Off",
             1: "On",
         },
-        allowedtypes=HYBRID,
+        allowedtypes=HYBRID | PV,
         write_method=WRITE_MULTISINGLE_MODBUS,
     ),
     SofarModbusSelectEntityDescription(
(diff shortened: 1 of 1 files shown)


mldebian:
Oh one more thing I found is that "PV Power Total" has a separate definition for PV inverters. It uses the same step size as 'PV Power 1', but that is incorrect and needs to use the same definition as is defined for hybrid inverters. With the extra definition removed and the hybrid definition extended for PV, the correct values show.

mldebian:
Thanks for pulling @wills106 . Any idea what to do about 'PV Power Total'?

wills106:
We could alter the scaling using a method similar to what we do for the odd SolaX Inverter:

https://github.com/wills106/homeassistant-solax-modbus/blob/fe9a4e56b39221390670a0481d58e36a52dc5379/custom_components/solax_modbus/plugin_solax.py#L1154-L1157

https://github.com/wills106/homeassistant-solax-modbus/blob/fe9a4e56b39221390670a0481d58e36a52dc5379/custom_components/solax_modbus/plugin_solax.py#L5294-L5306

Based on your Serial SS1

So something like ("SS1", 0.1)

Hopefully that makes sense.

PR #1690, opened 2025-10-27
fix: PM register overflow by treating corrupted data as errors

fxstein:
## Problem

PM registers in SolaX parallel mode systems were showing massive U32 overflow values (4294967295) instead of legitimate sensor readings. This occurred when the Master inverter's PM communication system failed, returning uninitialized register data.

## Root Cause Analysis

- **Issue**: Master inverter returning uninitialized register data with high word = 0xFFFF
- **Pattern**: Raw register data showed patterns like [0xfff4, 0xffff], [0xfff5, 0xffff], etc.
- **Impact**: Users saw corrupted sensor values instead of last known good values
- **Affected Hardware**: SolaX X3-Hybrid-15kW (DSP v1.72 ARM v1.50) in Master mode

## Solution

Treat corrupted register data as data errors by:
1. **Detection**: Check if high word is 0xFFFF (indicates corrupted register)
2. **Error Treatment**: Set val=None to trigger existing error handling
3. **Result**: Sensor preserves last known good value instead of showing overflow
4. **Recovery**: Automatic recovery when valid data returns
5. **Debug Logging**: Additional debug logging available for dev and testing

## Implementation

```python
# PM register corruption detection - detect uninitialized high word pattern
if is_pm_register and raw_words[1] == 0xFFFF:  # High word is 0xFFFF
    _LOGGER.warning(f"PM_CORRUPTED_REGISTER: {descr.key} raw:[{hex(raw_words[0])}, {hex(raw_words[1])}] - treating as data error")
    val = None  # This triggers the existing error handling
```

## Benefits

- ✅ **No Retry Logic**: Leverages existing error handling mechanism
- ✅ **Performance**: No additional Modbus calls
- ✅ **Reliability**: Uses proven error handling approach
- ✅ **Simplicity**: Minimal code changes required
- ✅ **User Experience**: Clean sensor values instead of overflow values
- ✅ **Automatic Recovery**: No manual intervention needed

## Testing

- **Before**: Sensors showed 4294967295 (overflow values)
- **After**: Sensors preserve last known good values (e.g., 0.0, 3000W)
- **Logging**: Corruption events logged for monitoring
- **Recovery**: Normal updates resume when valid data returns

## Files Changed

- `custom_components/solax_modbus/__init__.py`: Added PM register corruption detection in treat_address() function

Fixes PM sensor data corruption in parallel mode systems.

## Checklist

- [x] Initial verified implementation in init.py
- [x] Modified initial approach in plugin_solax.py per @infradom 
- [x] Review recommendation by @infradom regarding plugin simplification
- [x] Implement final design

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -504,6 +504,7 @@ class SolaXModbusHub:
         _LOGGER.debug(f"{self.name}: ready to call plugin to determine inverter type")
         self.plugin = plugin.plugin_instance  # getPlugin(name).plugin_instance
         self.plugin_module = plugin  # Store plugin module for accessing module-level functions
+        self._validate_register_func = getattr(plugin, 'validate_register_data', None)  # Cache function reference
         self.wakeupButton = None
         self._invertertype = None
         self.localsUpdated = False
@@ -1233,8 +1234,8 @@ class SolaXModbusHub:
         """
 
         # Plugin-level validation hook
-        if hasattr(self.plugin_module, 'validate_register_data'):
-            val = self.plugin_module.validate_register_data(descr, val, data)
+        if self._validate_register_func is not None:
+            val = self._validate_register_func(descr, val, data)
 
         if val == None:  # E.g. if errors have occurred during readout
             #_LOGGER.warning(f"****tmp*** treating {descr.key} failed")


infradom:
I don t like having solax-specific code for pm_ in the shared __init__.py file
Furthermore, I believe we should tackle the root cause instead of trying to fix this in a non-generic way.
So my opinion on this:  lets please think twice before we merge this PR

If we really cannot find and solve the root cause, the solution should be in the plugin_solax.py file. If small generic framewark changes are needed, we can think about that, but only the plugins are allowed to contain inverter-specific code. One solution path would  be to introduce an entity attribute "check_function" that verified the value and returns whether it is valid or not. The pm_* entities in plugin_solax would than get this attribute (that can point to a single shared function in the plugin)

fxstein:
Happy to discuss alternatives. Right now the Parallel Mode implementation is broken and I am taking my fixes apart and submitting them piece by piece as PRs to keep the scope in check. The reason this is currently in init.py because the PM block code is there and it can and will receive corrupt data from inverters. These PM attributes are corrupted as much as 6-10% per day. The inverters seem to have internal issues with the PM mode and sending uninitialized data back in the form of 0xFFFF 0xFFFx pairs which our code turns into 4B numbers. That means that for every inverter setup that runs two or more inverters the PM sensors are heavily impacted. 

That 0xFFFF represents uninitialized values will hold true for any manufacturer. No matter what sensor, nothing that reads 4 Billion will ever be right, Solax or not. Aside from the debug logging that is in this PR, the 0xFFFF check is actually highly localized and uses all of the existing logic on whether to update the sensor with a new value or not. It is a single if statement that checks for un-initialized data and is only performed in the PM block portion of the code.

On top of that these PM sensors are a pre-requisite to get the remote control calculations right as currently the calculations in PM mode are catastrophically wrong and will throw fuses because inverters will blow any fuse limits as they go through their oscillations. Measuring 57-60kW import spikes on a 45kW setup that is fused for 40kW imports and should be limited to 25kW exports.

Hope this helps explain the need for a fix. How we fix it is obviously up for debate. Happy to discuss alternative strategies. Just trying to avoid duplicating error handling code that already exists in the PM block in init.py

infradom:
Thanks for explaining...

I am not aware of a PM block code  in init.py; Can  you hint what you mean? if there is one, it should be very generic.
I dont believe 0xFFFF will be valid for any manufacturer. Furthermore you use register  addresses in init.py
I do believ you need the PM sensors, but that should be realized in the plugin, not in the core. 

I still believe we should implement this part outside __init__.py to keep the framework independent.
Let me know if I can help making this separation (with an additional attribute/plugin function).



Forgive me for following  thought:

- If PM mode is so unreliable as you explain, I would not like to use it, seems like really dangerous to build further on rubbish data... Safety first !

fxstein:
Parallel Mode in the current implementation is completely broken. I have fixed it a while ago - just never got around to PR it all. 

But you have a valid point that this could be moved out of init and I will look at that.

fxstein:
@infradom Please take another look. Moved this into the plugin and applied only to the affected U32 sensors. If an inverter sends us an uninitialized reading from its internal PM stack, we use the prior value. This has proven a safe solution and allows all other functions to rely on that data.

fxstein:
Actually hold off - this new design is not working as planned the functions aren't getting called. I need an alternative way to accomplish this outside of core.

fxstein:
Turns out this needed to be way more involved to reliably work. Catch Solax specific register overruns for U32 registers and applies last known good value. 

The tradeoff to doing 5 lines as a generic fix in init are  100+ lines in plugin_solax as we have to do it sensor by sensor and need both a raw sensor and a computed sensor template.

infradom:
> Turns out this needed to be way more involved to reliably work. Catch Solax specific register overruns for U32 registers and applies last known good value.
> 
> The tradeoff to doing 5 lines as a generic fix in init are 100+ lines in plugin_solax as we have to do it sensor by sensor and need both a raw sensor and a computed sensor template.

I have not looked into detail, but I believe I can solve it without exploding the code in the plugin. Do you want me to give it a try? 
I would add a new attribute  to the sensor entity_descriptor( const.py). That attribute could even be added automagically by the plugin if the name starts with pm_  (and type is 32)
The attribute , lets call it "validate_data" could be None  (default) or a function that checks /corrects the value (or pops it from datadict). All pm_sensors could share the same validate_data function, so the plugin code remains minimal

Optional extra: Upon startup, an iniitialization callback could scal all entities and add this "validata_data" attribute to all entities that match pm_ ...(and other criteria)

Let me know if I should make a draft implementation ...

infradom:
See Edit of previous message

fxstein:
Let me give that thought a try. Would be great if we can make this work cleanly.

infradom:
> Let me give that thought a try. Would be great if we can make this work cleanly.

If you want to do the optional automatic attribute assignment, you need to modify the entity_descriptor, which is possible but requires a little trick. I will explain later ..

infradom:
> I will explain later ..

If we want to add the automatic assignemet of this new attribute, we might need to give the plugin class an additional "on_startup" method that modifies the declared SensorEntityDescription objects. Such modificaitions are already done in the LocalDataCallback method of the plugin (the replace statement).
But I believe we can first test with adding the new attribute manually before we add this automatic assignment for pm_ entities. Manual declaration might even be better as it does not try to hide the problem.

fxstein:
Added a simple checklist to the PR so we can track of what has been completed as we redesign this implementation. Also cross referenced the PR in the parallel mode PR #1694.

Planning to get this done today if my schedule permits.

infradom:
Very nice work, just see some preliminary remarks in the PR #1694 comment

fxstein:
Ok - I think this is a solid compromise. I checked out sensor level validation functions and ultimately ended up with the smallest and I think most elegant solution:

init.py checks if a plugin hook exists and if so calls it:
``` python
# Plugin-level validation hook
if hasattr(self.plugin, 'validate_register_data'):
  val = self.plugin.validate_register_data(descr, val, data)
```

The original logic now sits in plugin and is called and does its magic.

Least amount of code and complexity, keeps the init.pu core clean and allows every plugin to implement a validate_register_data() function as needed.

Added a summary developer doc into the commit with the details of what was implemented.

In this case we then check for `pm_` registers that are U32 and apply the overflow logic.

fxstein:
@infradom Did a quick performance test to get a ballpark for the overhead this adds:

📊 Performance Results:
## Key Findings:
Average overhead: ~0.9-1.0 µs per sensor
Multiple inverters reporting
Consistent across all measurements
PM sensors: 28 out of 2000 calls (~1.4%)
## Analysis:
Per sensor: ~1.0 microsecond
Per polling cycle (assuming ~200 sensors):
200 × 1.0 µs = 200 µs = 0.2 milliseconds
## Context:
Modbus read: ~10-50 milliseconds (network I/O) (system setup at 115200 baud)
Validation: ~0.2 milliseconds
Overhead: ~0.4% of total polling time

infradom:
@fxstein Thanks a lot !
I have not done this kind of tests myself, I just intended to load your version and see if the overall average load of my Rpi4 increased (now at 7% with 3 second fastest polling group)

fxstein:
Ok and found a tiny change that saves 35% of that. I think that is it. Most simplistic elegant solution for register validation without hidden complexity that other devs might find hard to understand or trace, and a very minimal overhead. In fact this type of caching we should consider across the entire plugin as the savings could be substantial overall - but that is beyond this scope. Will cleanup the code and removed my testing hooks and commit to this PR. This is my final piece here - need to move on to bigger fish to fry... 😉

# Plugin Validation Performance Optimization

## Measurements

**Before optimization:**
- Average overhead: ~1.0 µs per sensor
- Per polling cycle (200 sensors): 200 µs = 0.2 ms
- Overhead: ~0.4% of total polling time

**After optimization (function caching):**
- Average overhead: ~0.65 µs per sensor
- Per polling cycle (200 sensors): 130 µs = 0.13 ms
- Overhead: ~0.26% of total polling time

**Improvement:** 30-35% overhead reduction

## Code Changes

### 1. Cache function reference at initialization

**File:** `__init__.py` (line 507)

```python
self._validate_register_func = getattr(plugin, 'validate_register_data', None)
```

### 2. Use cached reference in validation hook

**File:** `__init__.py` (line 1240)

**Before:**
```python
if hasattr(self.plugin_module, 'validate_register_data'):
    val = self.plugin_module.validate_register_data(descr, val, data)
```

**After:**
```python
if self._validate_register_func is not None:
    val = self._validate_register_func(descr, val, data)
```

## Result

**Total changes:** 2 lines  
**Performance gain:** 30% overhead reduction (1.0 µs → 0.65 µs)  
**Complexity cost:** None

Eliminated per-call `hasattr()` check and attribute lookup overhead while maintaining plugin independence.

infradom:
> This is my final piece here - need to move to bigger fish to fry... 😉

Agree, the code is clean and easy to understand. Adding more tricks will not make it better (and the performance gain has not been proven yet). I still will test your version, but I am convinced that there will be no performance loss.

fxstein:
Ok this is good to go to be merged.

infradom:
Installed on my system; seems to work well . Thanks.  
Thanks also for guidance on performance testing !!!

PR #1693, opened 2025-10-28
Solax VPP Mode 8 - do not send duration (as dissussed in #1658)

infradom:
As specified in the protocol docs , new VPP register set , mode 8 section
See @TCWORLD remark in #1658

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -288,7 +288,7 @@ def autorepeat_function_powercontrolmode8_recompute(initval, descr, datadict):
     setpvlimit = datadict.get("remotecontrol_pv_power_limit",10000)
     pushmode_power = datadict.get("remotecontrol_push_mode_power_8_9", 0)
     target_soc = datadict.get("remotecontrol_target_soc_8_9", 95)
-    rc_duration = datadict.get("remotecontrol_duration", 20)
+    #rc_duration = datadict.get("remotecontrol_duration", 20)
     import_limit = datadict.get("remotecontrol_import_limit", 20000)
     battery_capacity = datadict.get("battery_capacity", 0)
     rc_timeout = datadict.get("remotecontrol_timeout", 2)
@@ -516,10 +516,10 @@ def autorepeat_function_powercontrolmode8_recompute(initval, descr, datadict):
             "remotecontrol_push_mode_power_8_9",
             pushmode_power,
         ),
-        (
-            "remotecontrol_duration",
-            rc_duration,
-        ),
+        #(
+        #    "remotecontrol_duration",
+        #    rc_duration,
+        #),
         (   "remotecontrol_timeout",
             rc_timeout,
         ),


rosenrot00:
@infradom and @TCWORLD I didn't follow all the discussion however, sending the duration when using SET is the right thing to do from my perspective. 

As it allows to use VPP in a way that if homeassistant restarts/breaks I can set the duration to 20s and when homeassistant isn't renewing it it simply stops. Using SET even tells that the new mode will immediatly start but for this to happen we need the duration to be renewed. What this change does is the following, I have to use a longer duration like 1h and I need to use an automation to trigger it over and over again. There is no way I can stay in mode 8 without an automation. Who wants that?

If I want this behavior that duration is just set once, I can use UPDATE.

Why do you see this as "wrong"?

TCWORLD:
@rosenrot00 The PR isn't to do with the set/update register - that is still always being sent as "set".

The mode 8 loop was simply sending one too many values, this fixes that. There is no duration register in the mode 8 command, only the timeout value. The timeout being sent here is not the auto-repeat time, it's the command timeout value, so you can still set a sensible timeout and have vpp stop if home assistant stops running.

In other words, use `number.solax_remotecontrol_timeout` as the safety value for timing out if home assistant locks up.

rosenrot00:
> @rosenrot00 The PR isn't to do with the set/update register - that is still always being sent as "set".
> 
> The mode 8 loop was simply sending one too many values, this fixes that.

I don't get it. So what I want is to select a mode89 submode, hit the trigger and it should auto-renew itself until I set the mode8/submode to disabled. How do I do that? The inverter should fall out of the VPP mode by itself if homeassistant isn't updating for 10s. Could you tell me how to set variables:

solax_remotecontrol_autorepeat_duration = ?
solax_remotecontrol_duration = ?
solax_remotecontrol_timeout = ?

TCWORLD:
@rosenrot00 You can still accomplish this as before.

* Set `solax_remotecontrol_autorepeat_duration` to how long you want the autorepeat loop to keep going. E.g. if you want it to run for 5 minutes, set to 300s. You can also set this to be arbitrarily long if you intend to disable it manually (trigger with mode=disabled). I think the default (28800) is 8hours!
* For mode 8/9, `solax_remotecontrol_duration` is unused. The parameter does not exist in the mode 8 VPP command set (it does exist for Mode1).
* Set `solax_remotecontrol_timeout` to a safe value longer than the register update rate (default update is 15s) - it's basically how long you want the mode to keep running in the event of home assistant crashing.

rosenrot00:
Thanks for these numbers. I just tried them and I set `solax_remotecontrol_autorepeat_duration` to 60s. The inverter stayed in Normal (R) even beyond the 60s. Even worse, it stays in Normal (R) even after a homeassistant restart.

solax_remotecontrol_autorepeat_duration = 60s
solax_remotecontrol_timeout = 10s

TCWORLD:
@rosenrot00 What state is the `sensor.solax_modbus_power_control` in after a restart? 

Also 10s might be too short for timeout - try something like 30s - if it's too short the inverter will stop before the autorepeat loop next runs.

rosenrot00:
`solax_modbus_power_control  = Self Consume - C/D Mode`

If my fast polling loop is 3s, 10s should do the job, no?

TCWORLD:
Ah ok, 10s should be fine.

Just tested setting autorepeat to 60s, the pressed trigger. My inverter went from modbus mode being disabled to the selected VPP mode, stayed there for ~75 seconds and then went back into disabled.

Which inverter do you have? Also maybe this conversation should be moved into an issue.

infradom:
Maybe the remotecontrol_next_motion parameter should be examined more closely: it determines what happens next after the chosen VPP mode expires:

image

Edit: (sorry this is for the mode8 new register set)

PR #1700, opened 2025-10-31
Add autorepeat support to selects & add manual battery control support for Solis inverters

Gregy:
Hi,
I need to be able to tell my inverter when to charge and when to discharge the battery and I did not want to deal with the timed configuration.

I implemented a support for manual battery control which I found in the inverter protocol documentation. This way I can order the inverter to charge/discharge in an hass automation.

The PR is a bit complex because I needed to add autorepeat support to selects.

Please let me know if I should take a different approach somewhere.

I am running this code successfully with Solis S6-EH3P10K-H

Have a nice day!

### DIFF
--- a/custom_components/solax_modbus/plugin_solis.py
+++ b/custom_components/solax_modbus/plugin_solis.py
@@ -120,6 +120,17 @@ def mutate_bit_in_register(bit: int, state: int, descr: str, datadict: dict):
     new_value = (value & ~(1 << bit)) | (state << bit)
     return new_value
 
+def value_function_battery_control_override(initval, descr, datadict):
+    """Value function for battery_control_override autorepeat - resends force charge/discharge commands."""
+    current_option = datadict.get(descr.key)
+    if current_option and hasattr(descr, 'reverse_option_dict'):
+        payload = descr.reverse_option_dict.get(current_option)
+        # Only repeat for force charge (1) and force discharge (2), not for off (0)
+        if payload in (1, 2):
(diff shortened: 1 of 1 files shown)


infradom:
Thanks, I only quickly checked the changes in init.py and other generic parts. 
Your work looks fine !

wawyed:
Is this not available for solis_fb00?

ajan1:
> Is this not available for solis_fb00?

I tested copypaste plugin_solis components it to solis_fb00 and it works.

bilalhp:
@Gregy wow, manual battery control is an extremely useful feature for me and thanks for this discovery!

I had a lot of complex logic to manage this with "timed charge/discharge" and even with that precise control is not possible because you can only set a current.

I also confirm this works with my rhi-3p10k-hves-5g.

cdensel:
Hello,

Thanks @Gregy, this could be really useful for me. I have the 5Kw model of Solis S5-EH1P(3-6)K-L and I often want to charge the battery to 100% if we have been notified by the power company that there will be a power outage.
Do you know if my model of inverter will support this?

Would someone have an example of an automation they are using to set the manual charge please? I don't want to risk damaging the inverter or battery by setting the values incorrectly.

Thanks,
Conor

bilalhp:
@Gregy @wills106  one thing I just realized is that the values for "Battery control override", "Battery control override charge power" and "Battery control override discharge power" are not being read after an HA reboot. They stay unknow until I set the values. See:

image

image

slate-1:
@Gregy - This sounds great, if possible can the changes also be implemented for plugin plugin_solis_fb00.py

Would be much appreciated.

Thanks

PR #1714, opened 2025-11-05
Add Modbus device ID field and option to isolate Modbus bus per hub

rosenrot00:
Adds a configurable Modbus device ID (unit/slave address) and an option to isolate the Modbus bus per hub by creating a dedicated Pymodbus client.
This improves flexibility and stability in multi-device or multi-bus setups.

@wills106 I only have limited possibilities to test it as I have only one inverter. After replacing the changed files my previously installed hub was still working. Maybe it is worth while having a none official release and test it with people with two inverters like here? https://github.com/wills106/homeassistant-solax-modbus/discussions/1696


rosenrot00:
I guess this one got obsolete meanwhile.

PR #1750, opened 2025-12-07
grid power: add state class

gec75:
Home Assistant requires for energy dashboard


TCWORLD:
Just FYI even with this, the polarity of the sensor may be wrong for the energy dashboard (it is for mine with a CT). I'm hoping a PR on the HA front-end to select the polarity will be approved allowing them to play nicely.

gec75:
It's true... Inverters with CT usually have positive value for "producing" = "exporting"... the opposite of normal grid meters...

gec75:
Would it make any sense to change the reported power sign from the integration??

I just used a template sensor to invert the sign for energy dashboard/power.
I also have EDL21 with standard/compatible readings, but it's too slow.

TCWORLD:
@gec75 I started looking at proposing adding inverted sensors in #1754.

PR #1753, opened 2025-12-10
Add Gain/Offset Calibration to Measured Power

TCWORLD:
The CT on my system has inaccuracies, both in terms of gain, but also a ~40W offset meaning the inverter thinks it is exporting more than it actually is.

There is currently no way to calibrate this in the SolaX hybrid inverter itself which is somewhat annoying.

For the remote control modes, they require reasonably accurate `measured_power` values as they are basing their operation on "house load" which is calculated from the measured power value. So here we add the ability to apply gain/offset correction to the measured power value in the form of a dynamic scale factor based on two new (disabled by default) entities:

image
image

The measured power has the offset added and then the scale factor (%) applied using the sensor dynamic scale option with a new helper function so no need for an additional sensor.

While this doesn't help with the energy statistics produced by the inverter, it improves the remote control handling at low powers.

### DIFF
--- a/docs/solax-entity-description.md
+++ b/docs/solax-entity-description.md
@@ -101,6 +101,31 @@ Shows internal inverter temperature, you can use this information to know if you
 ## lease_mode
 
 
+## measured_power
+
+Current power as measured by main grid connection CT or meter. Positive values indicate export and negative values indicate import.
+
+This entity is used in remote control (VPP) modes to monitor grid power such as negative-injection/zero-export modes.
+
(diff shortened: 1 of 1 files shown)


wills106:
Thanks, I think it might be worth adding a page to the docs explaining this function?

TCWORLD:
Agreed. I'll add something after Christmas.

TCWORLD:
Added a note in the solax entities documentation.

Issue #1751, opened 2025-12-07
[FR]: 'Reverse battery combined power'


TychoGold:
### Description

Hello,
at first thank you for your work on the integration. Works really great.
With the new HA 12.2025 the energy dashboard is capable of showing current power values. The logic of the combined battery discharge unfortunately is reversed from the logic in this integration.
Is there the plan to reverse the logic or add additional sensors to the integration to support both logics (I could create a PR if answer is yes).

Regards,
Tycho

### Suggested Solution

Either implement second sensor or reverse the current one

TychoGold:
To me, as the battery is power delivery, it makes more sense to reverse the current logic to match the new HA dashboard logic, but I know that people may rely on the current logic in their automations.

rosenrot00:
You can simply create a template sensor and invert it yourself - I wouldn't change the integration as it would mess with peoples statistics.

TCWORLD:
I've proposed the following PR to home assistant frontend which if accepted will rectify this: https://github.com/home-assistant/frontend/pull/28397

TCWORLD:
Looks like they powers that be at HA prefer to push the burden back to users to create extra template sensors to negate the measured power and battery power sensors. 

We shouldn't change the polarity of the existing sensors here as it would mess up history and all existing use cases. Though I'm tempted to add an additional set of inverse sensors into this integration to provide the inverted values for the power chart.

WillCodeForCats:
I don't like it either, and I think creating another entity in the integration just for inverse power is dumb, but they are convinced their way is the only way so I don't suppose there is a choice between offering it in the integration and forcing users to make custom template sensors. I also think it's ridiculous that after they deploy a feature that clearly excludes systems that, when installed correctly, present grid power opposite of their feelings, now they want to be careful and slow about changes when they could have done that in the first place before releasing the grid power feature.

This is my solution for solaedge-modbus-multi: https://github.com/WillCodeForCats/solaredge-modbus-multi/pull/928

I'm not going to invert all of the power sensors, just the one summary power sensor for an attached meter.

TCWORLD:
Not stale.

wills106:
Just had a read through that PR and the linked comments.
Seems silly they won't consider having an inverse option, when there is clearly a demand!

fxstein:
[...]

PR #1754, opened 2025-12-12
WIP: Adding "Dashboard Grid/Battery Power" helper sensors

TCWORLD:
HA power dashboard requires all power sensors to be "+ve to Load". i.e. a positive value indicates power flowing from a source to a load, and negative value indicates from load to source. The Solar, Battery, and Grid are all considered Sources, so:

 * Battery discharge would be +ve and charge would be -ve
 * Grid import would be +ve and export -ve
 * Solar power would always be +ve

Certainly the battery and grid sensors in a number of (if not most) inverters seem to be the other way around from the direction HA favored.

To cope with the currently predefined direction required for the home assistant power dashboard, begin adding additional sensors which if neccessary present negated versions of grid power and battery power in the home assistant expected flow direction.

This will probably need to be implied to all inverter files for consistency. Currently this is just a proposal (see #1751), and depending on the direction HA devs finally settle on, may or may not be a necessity. Please comment on the linked issue to indicate if your inverter needs any of battery/grid/solar inverting (and the sensor names).

---

@wills106 Your thoughts on this as ever would be appreciated. I've tried proposing to HA to correct for this in the dashboard itself but they seem to be insistent on pushing the mess back on to the integrations to deal with.


fxstein:
Question: What Meter Direction does your device show? Its usually a hidden sensor that needs to be enabled manually.  

In my case I am running with an external meter and it is showing `Meter 1 Direction: Positive`

In my case `Measured Power` is negative for imports and positive for export. <- your point

Here is how my energy dashboard is setup: (ignore the broken price sensor - its a HA issue I am dealing with)
Screenshot 2026-01-10 at 11 00 31

This seems to work fine without the need to invert and of the numbers.

TCWORLD:
@fxstein I'll check measured power direction, but seems to be the same as yours. This, along with the battery charge power sensor are the opposite direction to what HA requires for the energy dashboard "power" sensors (the energy sensors are fine). 

If you look on the "now" tab, the graphs are wrong unless you invert both sensors. 

Whether we also need to take into account meter direction when deciding to invert the measured power value is an interesting question.

fxstein:
@TCWORLD Got it. Thank you! Did not use the realtime metrics on the Energy dashboard until now. I will take a look later tonight to verify - but in principle I support the need for a solution. Hopefully this is a quick one.

fxstein:
@TCWORLD I think this is a great idea. Have some thoughts on how to possibly implement that.

The issue is sensor duplication and naming choices here is an simple example: (threw it together very quickly)

| Plugin | Grid Sensor | Battery Sensor | In PR | Priority |
|--------|------------|---------------|-------|----------|
| plugin_solax.py | ✅ measured_power | ✅ battery_power_charge | ✅ Yes | High |
| plugin_solinteg.py | ✅ measured_power | ❓ Unknown | ✅ Yes | High |
| plugin_solis.py | ❓ Unknown | ✅ battery_power | ✅ Yes | High |
| plugin_solis_fb00.py | ❓ Unknown | ✅ battery_power | ✅ Yes | High |
| plugin_sofar.py | ❓ Unknown | ❓ Unknown | ❌ No | Medium |
| plugin_growatt.py | ❓ Unknown | ❓ Unknown | ❌ No | Medium |
| plugin_alphaess.py | ❓ Unknown | ❓ Unknown | ❌ No | Low |
| plugin_solax_ev_charger.py | ❓ N/A | ❓ N/A | ❌ No | Low |
| plugin_solax_lv.py | ❓ Unknown | ❓ Unknown | ❌ No | Low |
| plugin_solax_a1j1.py | ❓ Unknown | ❓ Unknown | ❌ No | Low |
| plugin_solax_mega_forth.py | ❓ Unknown | ❓ Unknown | ❌ No | Low |

Just for the main two sensors we could be writing a ton of code and the sensors could be confused amongst the long list of sensors already in each device.

It gets worse if you consider parallel mode implementations as some of the sensors are different e.g. solax_1_pm_...
It would mean even more sensors to be implemented, further cluttering the plugins and the user experience.

After some playing around with ideas I think I came up with a possible alternative that could simplify this from both an implementation per plugin perspective as well as user experience:

Create a virtual device as part of the integration that specializes on the dashboard metrics. Each standalone or master (for pm) would automatically get a matching virtual device with only the sensors required for the dashboard. It would keep those sensors cleanly separated and easy to pick

```
Devices:
├── SolaX 1 (Main Device)
│   ├── Grid Power (original)
│   ├── Battery Power Charge (original)
│   └── ... other sensors
└── SolaX 1 - Energy Dashboard (Virtual Device)
    ├── Grid Power (Energy Dashboard)
    └── Battery Power (Energy Dashboard)
```

On the implementation side, rather than writing tons of new sensors, we would have a simple mapping document in each plugin. It would map both default and pm sensors. The virtual device logic would pick the proper one depending on config and setup. 

```python
# In plugin_solax.py
ENERGY_DASHBOARD_MAPPING = EnergyDashboardMapping(
    plugin_name="solax",
    mappings=[
        EnergyDashboardSensorMapping(
            source_key="measured_power",  # Single inverter mode
            source_key_pm="pm_total_measured_power",  # Parallel mode Master
            target_key="grid_power_energy_dashboard",
            name="Grid Power (Energy Dashboard)",
            invert=True,
        ),
        EnergyDashboardSensorMapping(
            source_key="battery_power_charge",  # Single inverter mode
            source_key_pm="pm_battery_power_charge",  # Parallel mode Master
            target_key="battery_power_energy_dashboard",
            name="Battery Power (Energy Dashboard)",
            invert=True,
        ),
    ],
)
```

We could further enhance that and have both Power and Energy sensors cleanly mapped. That way the setup of the HA Energy dashboard would become super easy and error proof.

I'd be happy to help and draft up the virtual device logic. Should be only a day or two of work. In return it would give us great flexibility and simplicity on the sensor mapping side.

Thoughts or reactions on this?

TCWORLD:
@fxstein I agree it would be a complete pain to add all the necessary duplicate logic across all the various files - I started doing it but then realised it was a load of effort so held off doing the rest.

Your virtual device mapping approach sounds like a much better design choice. If you're happy to implement that, I'll probably close this PR and go for your approach instead.

fxstein:
@TCWORLD I will take a shot at it - but lets leave this PR draft open for now until we are sure.

@wills106 Would you mind taking a quick look and providing some feedback if you agree with the potential approach?

Thank you!

fxstein:
Quick update: I am almost ready to publish a draft PR on this work. Put more effort into it than I first planned but got to see the power of this approach when done right. Have mapped all GENs and a few of the plugins to start and it does support full parallel mode. (that was a lot more work than planned - but well worth it)

TCWORLD:
I'm happy to test out the changes once you're ready.

TCWORLD:
Closing this in favour of #1790 .

PR #1757, opened 2025-12-18
Growatt - Detect model

flopp999:
I have add a new way to detect model. If the new way don't find the model it will use the old way


flopp999:
It will now show serial numer correctly
image

PR #1768, opened 2025-12-29
Update plugin_growatt.py by removing repeated use of 1044 register

brad5505:
I have a GEN3 inverter and from my understanding, the priority  (register 118 and 1044) for battery first, grid first, load first can not be directly set as these registers are read only.  Therefore, the select entity code for register 1044 is ignored by the inverter. Priority can only be set by timer registers (begin time, end time, enable time) or via VPP registers

Code for two of these registers are currently in plugin_growatt.py (see below).

    GrowattModbusSelectEntityDescription(
        name = "Priority",
        key = "priority",
        register = 1044,
        option_dict = {
                0: "Load First",
                1: "Battery First",
                2: "Grid First", },
        allowedtypes = GEN3,
        entity_category = EntityCategory.CONFIG,
        icon = "mdi:run",
    ),

    GrowattModbusSensorEntityDescription(
        name = "Priority",
        key = "priority",
        register = 118,
        scale = { 0: "Load First",
                  1: "Battery First",
                  2: "Grid First", },
        register_type = REG_INPUT,
        allowedtypes = GEN2 | GEN3 | GEN4,
        icon = "mdi:run",
    ),

For my GEN3 inverter, I have removed the  “Priority” select code using register 1044, now  monitor using the priority sensor (register 118) code (already in plugin_growatt.py). All seems to work well.

The PR is to remove the  “Priority select code” using register 1044.


peteS-UK:
Thanks @brad5505 .  Yes, the growatt register doc I dug out show 1044 being R not R/W

image

As you say, you need to set the time based registers, but as discussed elsewhere, that's not currently an option as they need to be set in a group with 3039/3039

image

PR #1770, opened 2025-12-29
📚 Offer a tip on configuring Ethernet-enabled Wareshare devices without Windows

webknjaz:
I struggled with this for a bit but some amount of experimentation and studying the docs revealed that Waveshare devices come pre-configured with 192.168.1.200/24, making the Windows binary redundant for many. Actually, it was a few smart home enthusiasts from a local community who dug it out of the FAQ on a different page. @ByteStorm to be precise.

The PoE Wiki page fails to mention the defaults but the non-PoE doc at https://www.waveshare.com/wiki/RS485_TO_ETH_(B)#FAQ has this:

> **Question:** What is the default access IP of the webpage?
> **Answer:**
> The default is from 192.168.1.200;
> If the factory settings have been restored, the default IP is 192.168.1.254.


webknjaz:
Hope I didn't mess up the syntax anywhere as I don't normally use mkdocs. I've filed #1771 to let future PR authors see how their docs updates actually show up.

PR #1773, opened 2026-01-03
Fix: align Solax EV charger device info and Modbus telemetry

KIvosak:
### Overview
Align Solax EV charger device info and telemetry with the inverter: surface model/HW/ARM firmware, decode firmware correctly, fix Modbus map mismatches (RSSI, charging duration).

#### What’s Added / Changed
- Shared decoder value_function_firmware_decimal_hundredths in const.py for firmware values encoded as integer hundredths (611 → 6.11).
- EV charger detection now sets model and HW generation (Gen1 EVC, Gen2 HAC) and exposes them via getModel/getHardwareVersion; firmware uses the shared decoder and Device Info shows “ARM v<fw>”.
- RSSI sensor aligned to the Modbus map: percentage units (drops invalid signal_strength warnings).
- Charging duration sensor aligned with the map: duration in seconds.

#### Why This Matters
- Device Info for the charger now mirrors the inverter (model, HW gen, ARM firmware).
- Firmware displays the real value instead of mis-decoded numbers.
- Telemetry matches the Modbus map; HA logs stay clean.

#### Changes
- Files: const.py, plugin_solax_ev_charger.py.
- Logic-only updates; no config changes required.
- Charging duration sensor reports seconds per map (HA will prompt to update the unit; confirm to migrate without data loss).

#### Benefits
- Consistent device info across inverter and EV charger.
- Accurate firmware and duration reporting per Modbus map.
- Cleaner Home Assistant logs (RSSI unit fixed).

### DIFF
--- a/custom_components/solax_modbus/strings.json
+++ b/custom_components/solax_modbus/strings.json
@@ -2,21 +2,19 @@
   "config": {
     "step": {
       "user": {
-        "title": "Define Modbus connection",
+        "title": "Define your inverter modbus-connection",
         "data": {
           "name": "The prefix to be used for your inverter sensors",
-          "read_modbus_addr": "The Modbus address of the Inverter",
-          "interface": "Interface",
+          "read_modbus_addr": "The modbus address of the Inverter",
+          "interface" : "Interface",
           "read_eps": "Emergency Power Option (EPS)",
-          "read_dcb": "Dry Contact Box (SolaX Hybrid Gen4 & Gen5)",
+          "read_dcb": "Dry Contact Box (Gen4)",
           "read_pm": "Parallel Mode (Master-Slave)",
           "plugin": "Select Inverter Type",
-          "scan_interval": "Default polling interval (s)",
-          "scan_interval_medium": "Medium polling interval (s)",
-          "scan_interval_fast": "Fast polling interval (s)",
-          "time_out": "Request timeout (s)",
-          "inverter_name_suffix": "Name suffix for the inverter",
-          "inverter_power_kw": "Max inverter power in kW (values above are ignored)"
+          "scan_interval": "The polling interval of the modbus registers in seconds",
+          "scan_interval_medium": "Medium polling interval",
+          "scan_interval_fast": "Fast polling interval",
+          "time_out" : "Timeout in seconds"
         }
       },
       "serial": {
@@ -29,7 +27,7 @@
       "tcp": {
         "title": "TCP/IP Parameters",
         "data": {
-          "host": "The address of your inverter or Modbus interface",
+          "host": "The IP-address of your Inverter or Modbus Interface",
           "port": "The TCP port on which to connect to the inverter",
           "tcp_type": "The Modbus TCP variant"
         }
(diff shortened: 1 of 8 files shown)


KIvosak:
[x1-x3-hac-ev-charger-modbus-tcprtu-v1.0.0-en.pdf](https://github.com/user-attachments/files/24418970/x1-x3-hac-ev-charger-modbus-tcprtu-v1.0.0-en.pdf)

fxstein:
Hi @KIvosak, I noticed we are both working on the EV charger integration. I opened PR #1778 which adds some enhanced debug logging and additional model mappings. I also have several uncommitted fixes running locally that I would want to commit and merge. (e.g. latest firmware for some GEN1 chargers makes them run GEN2 firmware, a bug in the charging time mapping and on top of all have done extensive testing with the Solax Datahub1000 for multiple EVCs on a single property)
Any preference on how to merge our work? Normally I would wait for the PRs to get merged, but I cant predict when that will happen and I need the combined fixes for my setup. 
I have created a local test branch that combines both PRs and applies extra fixes and enhancements on top of that, but did not want to put that into my PR until you and I have had a chance to sync up.

KIvosak:
Hi @fxstein,

I would suggest keeping the detection logic and model strings as fixed in #1773, and building any further changes on top of that.

Ideally, additional changes would be added as commits on top of this PR, limited to debug logging only, without modifying device detection or model mapping. After #1773  is merged, rebasing and extending with debug should be straightforward.

Regarding the serial number prefix 5070: this corresponds to a 22 kW X3-HAC Gen2 unit. I have a real installation with an SN starting with 5070, so mapping it to 11 kW does not match the actual device. The incorrect handling of the 22 kW variant was introduced in this [commit](https://github.com/wills106/homeassistant-solax-modbus/commit/43570c509e7389c6389ca6f2cd51ae8442d79dd3).

One more small naming note: Gen1 units use the EVC designation, while Gen2 units use HAC. In your PR #1778, everything is unified under HAC, so it would be good to align the naming with the actual generation to avoid confusion in the logs.

Overall, it makes sense to merge #1773  first as a behavior fix, and only then layer extended debug logging on top of it, to avoid reintroducing already fixed issues.

fxstein:
Thanks! Normally I would just wait for the PRs to get merged but it has been since November for the last PR to get merged. I changed my PR into draft mode to combine our changes but to keep your commits. You are correct that your strings are the ones to keep. But there are additional issues because the latest GEN1 chargers come with GEN2 FW and need to be clarified as such. The debug logging was just the tooling to get to that. Please take a look at my draft PR with your commits merged in.

KIvosak:
Thanks for the clarification and the context so far, @fxstein.

I do have some concerns about potentially affecting support for EVC (Gen1) chargers, so I want to be careful not to introduce regressions for devices that are currently working correctly.

My latest verified firmware information as of 2026-01-08 is:

- G1 EVC: 1.17 / 3.012.03
- G2 HAC: 009.05 / 10.02

The EVC firmware version you mentioned [here](https://github.com/wills106/homeassistant-solax-modbus/pull/1778/commits/5483458eb43d574c35edd44b6261ae4954b3dccb) does not align well with the known EVC firmware line and clearly deviates from what we see on real EVC installations. Because of that, I’m hesitant to generalize this behavior to all EVC devices.

Given this mismatch, could it be a specific or transitional hardware variant? For example, an older EVC chassis combined with a HAC control board internally, or another mixed configuration used during a transition period. That would explain why the firmware versions don’t fit the standard EVC sequence.

From my perspective, the safest approach would be to introduce a new HW variant for this specific case, instead of modifying the existing EVC logic, ideally based on a combination of the serial number (SN) and the detected firmware version. This would preserve the current, working EVC support and minimize the risk of breaking existing installations.

fxstein:
Agreed. Have been thinking along the same lines. I know that my installer swapped the earlier models out for me, as the consumer version of these did not have an ethernet port and they could not get three of them to work with my inverters. These might be their commercial versions since they have ethernet port, respond to 502 and just got upgraded to firmware 7.07 by the vendor. It was 3.x before the fw upgrade. It might also be that 7.07 will become available more widely for GEN1 EVCs but that's a speculation. Will put something together for us to review.

fxstein:
Ok added the dynamic firmware detection and only upgrade C311/C322 to GEN2 when fw >7.0
Also added docuemntation about the EVC detection and operation and a little bit on EVC <-> Datahub interaction and what it means for the solax_modbus integration.

KIvosak:
> Ok added the dynamic firmware detection and only upgrade C311/C322 to GEN2 when fw >7.0 Also added docuemntation about the EVC detection and operation and a little bit on EVC <-> Datahub interaction and what it means for the solax_modbus integration.

So far, i’ve done basic validation on two real installations and didn’t encounter any issues:

- X3-EVC-22 (Gen1) with firmware 1.17
- X3-HAC-11 (Gen2) with firmware 6.11

Both setups behaved as expected with the current changes, and i didn’t observe any regressions related to EVC support during this testing.

dskvml:
> Regarding the serial number prefix 5070: this corresponds to a 22 kW X3-HAC Gen2 unit. I have a real installation with an SN starting with 5070, so mapping it to 11 kW does not match the actual device. The incorrect handling of the 22 kW variant was introduced in this [commit](https://github.com/wills106/homeassistant-solax-modbus/commit/43570c509e7389c6389ca6f2cd51ae8442d79dd3).

I have a 5070 with 11kW, thats why i made this change. Maybe serial number alone does not help here.

KIvosak:
> I have a 5070 with 11kW, thats why i made this change. Maybe serial number alone does not help here.

This does not affect functionality of the integration — only the displayed model name/power may be incorrect.
I’m verifying the SN-based detection pattern directly with SolaX and will update once I have a response.

KIvosak:
Fixed #1798

PR #1781, opened 2026-01-10
i18n: fix missing translations, normalize labels, and update docs

KIvosak:
## Overview

This PR completes missing translation entries, normalizes configuration/option labels, and includes a small documentation update.

Scope remains limited to translations and documentation only — no functional or behavioral changes.

### i18n
- Added missing translation keys based on the current config and options flow.
- Normalized polling interval and request timeout labels, including units.
- Updated host wording to avoid IP-only phrasing (hostnames are supported).
- Unified “Modbus” capitalization and general label wording.
- Aligned inverter power label format.

### Documentation
- Updated documentation to reflect the current configuration wording and options.

## Languages

DE / FR / IT / NL / SE translations were generated via machine translation to ensure coverage.  
Follow-up corrections by native speakers are welcome.


KIvosak:
Resolved the conflict introduced by commit [32f5f8e](https://github.com/wills106/homeassistant-solax-modbus/commit/32f5f8e66aa10d24326266e23e33048be5aa8bea).
Propagated the updated inverter_power_kw label to all translation files and strings.json.

KIvosak:
Hi @wills106,

while filling in the missing translation strings, I got an idea regarding a possible improvement to the configuration flow.

What do you think about moving plugin-specific parameters into a final configuration step, similar to how this is currently handled for Sofar with battery reading?

This would simplify and shorten the initial configuration dialog, which currently contains Solax inverter–specific options that are not relevant for Solax EV chargers or potentially other inverter types.

If such a refactor sounds reasonable to you, I can put it on my TODO list and work on it when I have some spare capacity.

wills106:
That makes sense. The configuration has just evolved as time has progressed, so covers things some Inverters don't support or not implemented yet.

PR #1783, opened 2026-01-11
Solax add meter1 variables

rosenrot00:
I added the power factor as well as energy (today, total) related variables. They are off by default, I just would like to have them to compare measurements between the inverter and the meter.


custom_components/solax_modbus/plugin_solax.py
```diff
@@ -6649,6 +6649,93 @@ def value_function_battery_voltage_cell_difference(initval, descr, datadict):
...
+        native_unit_of_measurement=PERCENTAGE,
+        device_class=SensorDeviceClass.POWER_FACTOR,
+        state_class=SensorStateClass.MEASUREMENT,
+        register=0xC4,
+        register_type=REG_INPUT,
+        unit=REGISTER_S16,
+        scale=1,
+        rounding=0,
+        entity_registry_enabled_default=False,
+        allowedtypes=AC | HYBRID | GEN3 | GEN4 | GEN5,
```
wills106: Doesn't appear in the docs for Gen3

wills106:
I'm a bit confused about this PR.

A lot of the Meter 1 entities are existing entities, so we are reading the same registers twice.

rosenrot00:
@wills106 now that you say it - you are right. I blindly looked in the latest modbus documentation without checking the existing entities. I will close this down.

PR #1790, opened 2026-01-15
feat: Add Energy Dashboard virtual device, mappings, and diagnostics

fxstein:
# Energy Dashboard virtual device

## Summary

This PR introduces a virtual Energy Dashboard device that provides curated grid, battery, and solar sensors tailored to Home Assistant’s Energy Dashboard requirements. The implementation standardizes naming and sign conventions, supports parallel systems (Primary + Secondary), and adds diagnostics plus debug helpers to simplify validation and troubleshooting.

## Key Outcomes

- Virtual Energy Dashboard device per inverter hub with curated sensors
- Clean, consistent energy and power sensors for Energy Dashboard and “Now” views
- Parallel mode support with system-wide totals and per‑inverter detail
- Diagnostics to surface mode, inverter count, and configuration context
- Debug infrastructure to enable targeted development and validation

### Virtual devices
Energy Dashboard virtual devices

### Example standalone virtual device
Energy Dashboard virtual device standalone

### Example parallel mode virtual device
Energy Dashboard virtual device parallel mode

### Virtual device config flow
Energy Dashboard virtual device config flow

### HomeAssistant Energy dashboard config example
HomeAssistant Energy dashboard config example

## User Experience Highlights

- Energy Dashboard setup becomes straightforward: select the curated sensors created by the integration
- “Now” view is enabled by correct, signed power sensors without custom helpers
- Parallel systems gain a unified set of totals plus individual inverter visibility
- Diagnostics explain how the device is interpreting parallel mode state

## Energy Dashboard Metrics and Inversions

The integration adapts raw inverter data to the Energy Dashboard requirements:

- **Grid power**: uses `measured_power`, inverted so import is positive and export is negative
- **Battery power**: uses `battery_power_charge`, inverted so discharge is positive and charge is negative
- **Solar power**: uses `pv_power_total` as-is (always positive production)

For energy counters:

- Uses daily energy registers where available (avoids spikes from lifetime totals)
- Falls back to totals if daily is unavailable
- Uses Riemann sum integration from power for GEN1 inverters lacking energy counters

## Parallel Mode Architecture

Parallel mode creates a consistent device layout:

- Primary inverter hosts an **All** sensor set (system‑wide totals)
- Per‑inverter sensors are created for detailed tracking
- Secondary devices are skipped unless explicitly enabled via debug override
- Aggregation rules:
  - **Power totals**: use PM totals when provided by the inverter
  - **Energy totals**: aggregate Primary + Secondary values where needed

## Diagnostics and Debug Infrastructure

The Energy Dashboard device exposes diagnostics (entity category: diagnostic) to show:

- Mode (Standalone vs Parallel Primary/Secondary)
- Inverter count and detected secondaries
- Debug override state (only when enabled)
- Mapping summary (plugin + count)

Debug settings are configured via `configuration.yaml` for development/testing only, allowing targeted overrides (e.g., treat a Secondary as Standalone for validation).

## Documentation Added/Updated

- User docs: `docs/energy-dashboard.md`, `docs/energy-dashboard-setup.md`
- Developer docs: `docs/developer/energy-dashboard.md`, `docs/developer/debug-settings.md`

## Test Plan

- [x] Standalone inverter: enable Energy Dashboard device and verify curated energy + power sensors appear
- [x] Home Assistant Energy Dashboard: select curated energy sensors and confirm “Now” view uses power sensors
- [x] Diagnostics: verify mode, inverter count, and mapping summary entries display as expected
- [x] Parallel mode: verify Primary exposes All + per‑inverter sensors and Secondary behavior matches configuration

### DIFF
--- a/custom_components/solax_modbus/sensor.py
+++ b/custom_components/solax_modbus/sensor.py
@@ -516,6 +516,7 @@ class SolaXModbusSensor(SensorEntity):
             #try:    val = self._hub.data[self.entity_description.key] *self.entity_description.read_scale # a bit ugly as we might multiply strings or other types with 1
             #except: val = self._hub.data[self.entity_description.key] # not a number
             #return val
+        return None
 
     @property
     def extra_state_attributes(self):


fxstein:
@TCWORLD This is regarding our discussion in #1754

Was a lot more work than I thought to do it right. It covers all energy and power metrics used in the HA Energy dashboard, contains all possible mapping for SolaX GEN1-GEN6 inverters (and basic mapping for other inverter types), full support for parallel mode (that was the majority of my time spent on it), adds some better debugging tooling and a bunch of documentation.

Have it running on my system and it seem to work well. I am sure I missed a few things or left some bugs but hopefully they are quick to iron out.

Let me know what you think.

@wills106 FYI please provide feedback.

fxstein:
@TCWORLD This is awesome feedback. Yes I had the PV variant power sensors in there, but removed them because they had no energy equivalent (Duuuh! 🙈) - never crossed my mind to use the very same function I added for GEN1 for it. Also I am on a parallel system with 3 inverters and am already getting 3 Solar sub buckets because of that. 

You are also right that we don't get good household energy tracking from the inverters.

The funny thing - with the infrastructure I built, these all are just mappings. Only need to bring back my variant logic so one does not have to map PV1, PV2, PV3, PV4 ... up to PV6 for some inverters.

I will take a look.

fxstein:
@TCWORLD Ok - added the sensors as optional ones and created the infrastructure for dynamic optional settings. When you test this you can simply flip the switches below the sensors and the ones you enable should appear within seconds. When you disable them they will disappear a few seconds later. The state of the settings is preserved over restarts and the statics do not get removed. So removing them by accident does no harm, just re-enable them and they are back to where they were before. The energy integration sensors only calculate and increase when activated.

ha-solax-virtual-device-optional-disabled
ha-solax-virtual-device-optional-enabled

Please give it a try and provide feedback.

TCWORLD:
I'll give it a go over the weekend

TCWORLD:
Just set it up (used the Samba addon to copy all the new files across). The virtual device has loaded correctly and showing all the various sensors. I'll let it run over the course of today to see how they update relative to the various sensors I'd manually created before.

---

<strike>
One request is that it would be nice to have a way to set the current value of the riemann sum energy sensors - similar to how the built-in "Utility Meter" sensor has a "calibrate" action so that I can set the value to match the existing totals I have.

Setting the state in the developer tools doesn't work as it simply gets overridden on the next recalculation.
</strike>

Never mind, I now see the riemann sums are handled as today sensors resetting at midnight, so no need to calibrate them.

fxstein:
@TCWORLD Found a bug in the Riemann calculations for the PV variants. It was writing to the wrong counter. It's fixed now and PV1 and PV2 should now be correct.

fxstein:
@TCWORLD To help with debugging there is a hidden debug setting that you can temporarily enable in `configurations.yaml`:

```
# SolaX Modbus Integration - Debug Settings (DEVELOPMENT/TESTING/DEBUGGING ONLY)
solax_modbus:
  debug_settings:
    "Solax 1": # Your name here
      reset_riemann_sums_on_restart: true
```

Next time you restart it will reset the counters to ZERO - you should do that after taking the fix I just provided. Do not forget to remove the debug setting after the restart or it will rest them EVERY time you restart HA.

custom_components/solax_modbus/sensor.py
TCWORLD: Lines 735 to 827 are a duplicate of lines 444 through 536 - I assume something odd happened during a merge at some point.

TCWORLD:
For the Mapping Summary sensor on the energy dashboard, I wonder if its worth also outputting the actual map to the `state_attributes` for that sensor to allow seeing what sensor was picked as the source data?

custom_components/solax_modbus/plugin_solax.py
```diff
@@ -9943,6 +9943,223 @@ def localDataCallback(self, hub):
...
+            source_key="battery_output_energy_total",
+            target_key="battery_energy_discharge",
+            name="Battery Discharge Energy",
+            needs_aggregation=True,
+            allowedtypes=GEN2,
+        ),
+
+        # Grid to Battery Energy (per inverter, aggregate in parallel)
+        EnergyDashboardSensorMapping(
+            source_key="e_charge_total",
```
TCWORLD: For consistency with the other sensors which are daily tallies, should this be `e_charge_today`?

fxstein: Good catch. You are correct. for the dashboard we want as much daily counters, as any resets or issues would otherwise lead to extreme swings in the charts. Fixing...

custom_components/solax_modbus/sensor.py
fxstein: Correct - I messed up during my testing. just committed a cleanup. Thank you!

fxstein:
Update: Added ED mapping metadata to sensor state attributes so each Energy Dashboard entity shows its resolved source key and hub. Verified on live entities (non-Riemann + Riemann). Commit: 89347dc.

custom_components/solax_modbus/sensor.py
```diff
@@ -587,7 +516,10 @@ def native_value(self):
             #try:    val = self._hub.data[self.entity_description.key] *self.entity_description.read_scale # a bit ugly as we might multiply strings or other types with 1
             #except: val = self._hub.data[self.entity_description.key] # not a number
             #return val
-        return None
```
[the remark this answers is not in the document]
fxstein: @TCWORLD Thank you! Default is None - hence why I missed it. Better to have it there explicitly.

fxstein:
Unless there are any more bugs for me to consider - this is good to go from my side. I wont add more functionality into this PR as it is already BIG enough. Any followup requests will happen once this PR has been approved and merged. Thank you!

Special Thanks to @TCWORLD for helping me with this!

fxstein:
@wills106 Let me know when you had a chance to review this. Any questions - just ask. I realize this is big and touches a lot of things - but hopefully for the better. There are some new patterns in here that could be very helpful for other parts of the integration (debug settings, dynamic settings) - Happy to discuss.

TCWORLD:
It's looking good from my end. I will keep the new changes running through tomorrow to double check a full day of reimann sum data.

The individual pv energy sums recorded 0 today - but there was very little solar, and you've already made a fix to rounding in the calculation that would account for it.

---

Edit: Yep, the PV energy sensors are now recording values 👍.

wills106:
I have just updated my HA to match the current GitHub Master.

I haven't added anything to the Dashboard yet as it's too dark for PV etc.
I'll hopefully get chance to setup the Dashboard during the day tomorrow.

image

Probably just needs some tidying up under Diagnostics for Inverters that don't offer PM mode.

Edit:

Thanks @fxstein for all of this hard work.
Also thanks @TCWORLD for helping in testing.

wills106:
Dashboard sensors have worked as expected today 👍

fxstein:
Agreed - some of the diagnostics and attributes we should fine tune.

gurkburk76:
Under configuration in the pic there are 3 thigns i can enable, should those when added add further sensors? It does not on my solinteg, but i think that inverter is a bit of a WIP.

Also, on my solinteg i have an external meter hooked (retrofitted) to my solar panel inverter (solaredge) that dosenät should up, this might be a topic of it's own but it sure would be nice to have that in this integration.

fxstein:
> Under configuration in the pic there are 3 thigns i can enable, should those when added add further sensors? It does not on my solinteg, but i think that inverter is a bit of a WIP.
> 
> Also, on my solinteg i have an external meter hooked (retrofitted) to my solar panel inverter (solaredge) that dosenät should up, this might be a topic of it's own but it sure would be nice to have that in this integration.

Yes - they should add additional sensors - and in realtime. When you enable one, it should create the additional sensors within seconds. But I have written this solely on SolaX inverters, so no way for me to test on others. Do you get any errors or warnings in your system log when you try to enable these?

gurkburk76:
I tried to disable and enable those 3 options and idon't see anyything in the log, this screenshot might tell something:
image

gurkburk76:
The integration itself works, i use it with predbat and it's just fine.

TCWORLD:
@gurkburk76 Can I refer you to issue #1821.

Essentially at the moment only the SolaX plugin has had its energy dashboard fully populated. Over time we'll need to add full mappings for each of the other plugins. I've posted a full list of mapped sensors in the SolaX plugin in the issue above.

The switches in the configuration link to the names of `target_key` entry in the mappings:

* "Enable Grid to Battery Sensors" refers to mapping targets with a prefix of `grid_to_battery_`
* "Enable Home Consumption" is mapping targets with a prefix of `home_consumption_`
* "Enable PV Variant Detail Sensors" is mapping targets with a prefix of `pv_power_` or `pv_energy_`

Currently the `solinteg` plugin has only had two sensors added to its mapping so the switches won't do anything yet until the rest have been populated.

Issue #1793, opened 2026-01-17
[Bug]: EV Charger "Charging duration" shows extremely large value shortly after charging starts


fxstein:
## Summary
`sensor.solax_evc_2_charging_duration` reports an unexpectedly large value after only a few minutes of charging.

## Observed behavior
- Entity: `sensor.solax_evc_2_charging_duration`
- Reported value after a few minutes: `21,233,664.00 s`

## Expected behavior
- Duration should be on the order of minutes, not millions of seconds, after only a few minutes of charging.

## Environment
- Home Assistant OS (linux 6.12.51-haos)
- Integration: SolaX Modbus custom_component
- Integration version: 2026.01.1 (from manifest)
- Minimum HA version per HACS metadata: 2025.1.0
- Charger: SolaX EV Charger (GEN2)
- Sensor key: `charge_duration` (register 0x2B, input, 32-bit)

## Device Info (SolaX EVC 2)
- Model: X3-EVC-11kW (serial prefix C311 per integration mapping)
- HW version: Gen1 (GEN2 FW)
- FW version: ARM v7.07

## Notes
- The value appears immediately after charging starts and grows from there.

fxstein:
I will be opening a PR for this shortly - just wanted to start documenting these things better.

TenFiliman:
Hello, I’m not using an EV charger, but I’ve noticed similar “spikes” with my Wattsonic inverter when integrated via the SolaX Modbus component in Home Assistant. Occasionally, the sensors report completely unrealistic values for a few seconds and then return to normal. Examples I’ve seen include:

PV panels reporting several million kW, sometimes even at night
Battery state of charge exceeding 100% (e.g., 174%)
SOH briefly dropping from the normal 99.55% to 0.02%

These spikes do not appear in the Wattsonic Cloud, only in Home Assistant, so it seems related to the Modbus integration, not the physical hardware. I’m not sure if it’s the same underlying issue, but the behavior is very similar.

fxstein:
> Hello, I’m not using an EV charger, but I’ve noticed similar “spikes” with my Wattsonic inverter when integrated via the SolaX Modbus component in Home Assistant. Occasionally, the sensors report completely unrealistic values for a few seconds and then return to normal. Examples I’ve seen include:
> 
> PV panels reporting several million kW, sometimes even at night Battery state of charge exceeding 100% (e.g., 174%) SOH briefly dropping from the normal 99.55% to 0.02%
> 
[...]

PR #1794, opened 2026-01-17
fix: correct EV charger 32-bit word order

fxstein:
## Summary
- switch EV charger plugin 32-bit word order to big-endian
- aligns 32-bit values like `charge_duration` with expected seconds
- addresses issue #1793

## Test plan
- sanity-check 32-bit word-order math (little=21233664, big=324 for sample regs)
- not tested on live hardware in this change


fxstein:
Closing per request; will reopen after review.

PR #1795, opened 2026-01-17
fix: allow per-sensor endian override for EV duration

### LINKED ISSUE #1793: [Bug]: EV Charger "Charging duration" shows extremely large value shortly after charging starts

fxstein:
## Summary
- add optional per-sensor 32-bit word-order override
- apply override for EV charger `charge_duration` register
- keep device default little-endian per SolaX docs

## Rationale
SolaX documentation states 32-bit values are little-endian, but on GEN2 EV chargers the `charge_duration` register at `0x2B` is verifiably flipped in live data. The override lets this single register be read correctly without changing the default endianness for the rest of the device.

## Test plan
- [x] Verified in HA: `sensor.solax_evc_2_charge_duration` now reports correct seconds.
- [x] No other sensor behavior changes observed.

Fixes #1793


fxstein:
@wills106 Not sure this was intentional but the merge ended up cherry picking only part of the PR for merge. It breaks the code because the other file changes got lost.

I will create another PR and link to this one to make it an easy fix.

wills106:
Sorry if I did something wrong merging it? Didn't know you could even do a partial merge?

fxstein:
I have no idea how it happened. Could be because of other PRs. git sometimes gets confused with when files change file endings. (hence the separate PR  that we merged on that).
I opened #1802 to fix it.

PR #1800, opened 2026-01-20
fix: ignore zero battery SoC when prior value exists

fxstein:
# fix: ignore zero battery SoC when prior value exists

## Problem
Battery SoC occasionally reports `0%` for SolaX inverters, even though the inverter should not discharge below the minimum threshold. This produces a transient but invalid SoC reading in Home Assistant.

## Root Cause Analysis
- **Observation**: SoC can briefly drop to `0%` and immediately recover to the prior value.
- **Pattern**: Single-sample anomaly with no sustained drop.
- **Likely cause**: Intermittent Modbus read issues producing invalid values.

## Solution
Leverage the same plugin validation hook introduced in PR #1690 to detect invalid SoC values and reuse the last known good reading:

1. **Detection**: Treat `battery_capacity == 0` as invalid only when a prior SoC > 5% exists.
2. **Error Treatment**: Substitute the last known good SoC value.
3. **Result**: Avoid transient SoC drops to 0 while preserving valid zero readings for batteryless systems.
4. **Recovery**: Automatic recovery when valid data returns.

## Implementation
- Reuse the `validate_register_data` hook from PR #1690 (framework remains unchanged).
- Extend `plugin_solax.py` validation to cover `battery_capacity` (SoC) values.
- Store last known SoC values and only override when a valid prior reading exists (>5%).

## Benefits
- ✅ **Consistency with PR #1690**: Uses the same validation infrastructure.
- ✅ **Data Integrity**: Prevents invalid SoC drops to `0%`.
- ✅ **Safety for Batteryless Systems**: Leaves `0%` untouched if no valid prior SoC exists.
- ✅ **Automatic Recovery**: No manual intervention needed.

## Testing
- Live system validation planned (monitoring for SoC=0 events after deployment).

## Files Changed
- `custom_components/solax_modbus/plugin_solax.py`

## Checklist
- [x] Use PR #1690 validation infrastructure
- [x] Detect invalid SoC=0 when prior SoC > 5%
- [x] Preserve valid zero when no prior SoC exists
- [x] Test on live system


fxstein:
This is confirmed working for now on my systems. No more SoC drops to ZERO that cause charts to flip and derived calculations to fluctuate.

Screenshot 2026-01-20 at 10 37 53

Zdenek45:
Last year, one of my slave battery died (Solax X3-Hybrid-G4 with two T30 units). This caused the battery SoC to drop suddenly to 0%. I believe this behavior is expected given the hardware failure. While I don’t have a specific recommendation on how to handle this, I believe this scenario should be taken into account.

fxstein:
@Zdenek45 Much appreciate the feedback! The current implementation would catch that on the next restart. We could also add a counter into the fix to limited the number of continuous fixes applied and say drop to ZERO if the past 10 readings all where zero. Given it is a rare scenario, that slight delay might be a decent compromise. Thoughts?

PR #1822, opened 2026-01-31
Populate Energy Dashboard Mappings for Solinteg

TCWORLD:
Add most of the energy dashboard mappings for the Solinteg plugin.

The following are not yet populated:
 * 'Grid to Battery Power'
 * 'Grid to Battery Energy'

### DIFF
--- a/custom_components/solax_modbus/plugin_solinteg.py
+++ b/custom_components/solax_modbus/plugin_solinteg.py
@@ -1636,7 +1636,6 @@ ENERGY_DASHBOARD_MAPPING = EnergyDashboardMapping(
             source_key="battery_power",
             target_key="battery_power",
             name="Battery Power",
-            invert=True,
         ),
         # Home Consumption Power
         EnergyDashboardSensorMapping(
--- a/custom_components/solax_modbus/plugin_solinteg.py
+++ b/custom_components/solax_modbus/plugin_solinteg.py
@@ -1638,6 +1638,20 @@ ENERGY_DASHBOARD_MAPPING = EnergyDashboardMapping(
             name="Battery Power",
             invert=True,
         ),
+        # Home Consumption Power
+        EnergyDashboardSensorMapping(
+            source_key="house_total_load",
+            target_key="home_consumption_power",
+            name="Home Consumption Power",
+        ),
+        # Grid to Battery Power (TODO)
+        # EnergyDashboardSensorMapping(
+        #     source_key="inverter_power",
+        #     target_key="grid_to_battery_power",
+        #     name="Grid to Battery Power",
+        #     filter_function=lambda v: max(0 - v, 0),
+        #     icon="mdi:transmission-tower-export",
+        # ),
         # ===== ENERGY SENSORS =====
         # PV Variant Energy (per string).
         EnergyDashboardSensorMapping(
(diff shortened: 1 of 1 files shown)


TCWORLD:
@gurkburk76 
Could you test the changes here with your inverter and make sure all entities in the virtual dashboard make sense?

Furthermore, could you let me know if any of your sensors represent either power from the grid to the battery (i.e. charging the battery of the grid), and whether there is an energy sensor that shows amount the battery has been charged from the grid (not by other sources such as solar).

gurkburk76:
Not sure how to pull this specific version via hacks, but my dash looks like this with the latest released release: 
image

TCWORLD:
@gurkburk76 I don't think branches appear on HACS unfortunately.

The way I've been testing such changes is to apply the changes to the installed files directly (in custom_components). I've update the branch to be based on the latest release (2026.03.2), so you should be able to download the file here:

https://github.com/TCWORLD/homeassistant-solax-modbus/raw/111801f4d942f864a1bf34af445681a6bc5f0fa0/custom_components/solax_modbus/plugin_solinteg.py

And replace the file in your installation using any of the file editor options in home assistant.

gurkburk76:
Looks good to me, bottom diag might be a bit "ugly" with unknown values but oh well.
Also, i don't have PV hooked up to this inverter so hard to test that part.

image
image

TCWORLD:
@gurkburk76  Looks like the polarity of house load is wrong - seems to be negative. Could you check polarity of all three power values is correct? 

Grid should be +ve for import, -ve for export
Battery should be +ve for discharge, -ve for charge
House load should be positive.

gurkburk76:
> @gurkburk76 Looks like the polarity of house load is wrong - seems to be negative. Could you check polarity of all three power values is correct?
> 
> Grid should be +ve for import, -ve for export Battery should be +ve for discharge, -ve for charge House load should be positive.

Currently around 7kw is being generated from a solaredge PV inverter so maybe that's why?

TCWORLD:
Ah ok, so they might be a bit weird.

Is the battery currently charging or discharging?

gurkburk76:
> Ah ok, so they might be a bit weird.
> 
> Is the battery currently charging or discharging?

Currently the house is using what it needs and with the leftovers it's charging the battery, can take a new screenshot later today when the sun is down and we'll get a better view, pretty sure the clamps are as they should as i did have reversed clamps when my solaredge got installed witch gave very odd reasings, clumpsy installers... :) 


EDIT: Actually, i took this from my solinteg app, the direktion/flows are animated so i made a artistic addition to it so it's clear what's going on.

image

Yeah third party meters makes it a bit weird, i'll show tonight after sunset.

gurkburk76:
Looks fine to me, phases also could possibly be of interest to have in the dash ? 
(it's obviously running off the batt. in the picture :) )
image

TCWORLD:
Ok, so it looks like battery power doesn't need inverting for the Solinteg ones - it should be -ve for charging and +ve for discharging, but as your last screenshot was house running on battery it needs reversing.

gurkburk76:
> Ok, so it looks like battery power doesn't need inverting for the Solinteg ones - it should be -ve for charging and +ve for discharging, but as your last screenshot was house running on battery it needs reversing.

Just so i understand what you're saying, do you think my clamps need reversing?

To me it makes sense that the powerdraw of the house is positive (solinteg house compumption in the screenshot) and that the power drawn out of the battery (solinteg battery power in picture) is negative as it's energy leaving the battery.

Maybe i'm wrong, but it makes sense to me.
Also, values in the solinteg app would be all wack if the clamps were reversed but it's fine there.

TCWORLD:
@gurkburk76 No, sorry nothinv you need to change/not your clamps. Just that I don't think we need to invert the value for the dashboard, that's all. 

HA is expecting by default positive for discharge and negative for charge (HA is home centric, where discharge is positive power into the home), which seems to be what your battery already was (if you look at the sensor in the main device not the virtual dashboard device).

It's kind of moot now as HA added internal support for this after work had already been done in this integration.

TCWORLD:
Closing as this will be merged as part of #2172.

PR #1824, opened 2026-01-31
Neaten ordering of Energy Dash Mappings for SolaX

TCWORLD:
Two of the "Power" sensors were in the section labelled for "Energy" sensors.


fxstein:
Thank you for the cleanup!

PR #1831, opened 2026-02-06
fix growatt modbus register ids for BMS 1 Module 1

jap:
This fixes what appears to be a transcription error, where register 5080 ended up as 5880, and the rest of the sensors for BMS 1 Module 1 also being off by 800.

With this fix, I get sensor readings on my setup (which is a TL-XH inverter with 1 BMS controlling several ARK batteries) that match Growatt's. Also note that module 2 starts at 5120, module 3 at 5160, module 4 at 5200, etc, which would make it logical for module 1 to start at 5080.


flopp999:
This is not working with APX modules. Notice that it was HOLDING and not INPUT.
I tried getting data for HOLDING 5082 it returns 0, HOLDING 5882 returns 97 and INPUT 5082 returns 97.
I remember that I couldn't find BMS 1 Module 1 Modbus registers but now I got a newer PDF for all registers and BMS 1 Module 1 is located from INPUT 5080-5119.
The PDF says that INPUT 5080-> 7639 is only for APX. Not each BDC but APX PM. ARK and APX is working different. APX has a BMS in each PM, but ARK has one BMS for all PMs. Correct?!
@jap is it working with HOLDING 5082 in your setup?

image

jap:
Good catch! I'm alternating between a factory dongle+solax and a custom esphome dongle, and my (esphome) config is indeed using the input (reading) registers and then it does work. I can make a quick patch later, but if you want to s/REG_HOLDING/REG_INPUT/ for these registers I'd say go for it..

Is that new PDF something you can share?

flopp999:
[2.1 GROWATT VPP COMMUNICATION PROTOCOL OF INVERTER_V2.01.pdf](https://github.com/user-attachments/files/25717033/2.1.GROWATT.VPP.COMMUNICATION.PROTOCOL.OF.INVERTER_V2.01.pdf)
[2.2 Growatt Inverter Modbus RTU Protocol_II V1_39-English-20240416.pdf](https://github.com/user-attachments/files/25717034/2.2.Growatt.Inverter.Modbus.RTU.Protocol_II.V1_39-English-20240416.pdf)
[1. ShineMaster Modbus diagram.pdf](https://github.com/user-attachments/files/25717035/1.ShineMaster.Modbus.diagram.pdf)
[0. Modbus system Diagram.pptx](https://github.com/user-attachments/files/25717036/0.Modbus.system.Diagram.pptx)

w4iq3:
This seems to have broken a few things on Module 1 info indeed..

image

PR #1799, opened 2026-01-19
chore: standardize line endings to LF

fxstein:
# Line Ending Standardization

## Problem
Line ending differences (CRLF vs LF) have been causing noisy diffs that obscure real changes and create recurring cleanup commits.

## Root Cause Analysis
- Git was not consistently normalizing line endings for text files across different operating systems.
- Contributors on Windows and Unix/Mac were unintentionally overwriting line endings.

## Solution
- Update `.gitattributes` to enforce LF normalization using `text=auto eol=lf`.
- Normalize all remaining tracked text files to LF in this PR.

## Implementation
- `.gitattributes` updated to standardize text file handling.
- Full repository renormalization performed so all tracked text files are now LF.
- No functional or content changes beyond line-ending normalization.

## Benefits
- **Cross-Platform Consistency:** Prevents "line ending wars" and noisy diffs.
- **CI/CD Stability:** Avoids CRLF-related issues in Linux-based tooling.
- **Binary Safety:** Binary assets remain protected from line ending conversion.
- **Cleaner History:** This should be the last line-ending-only cleanup commit.

## Testing
- Tested on a live system.

## Files Changed
- `.gitattributes`
- Tracked text files normalized to LF (line-ending-only changes).

## Checklist
- [x] `.gitattributes` updated to `text=auto eol=lf`
- [x] All remaining tracked text files converted to LF
- [x] No functional changes included
- [x] Tested on a live system

PR #1701, opened 2025-11-01
feat: Add code linting and formatting standards proposal

fxstein:
## Overview

Proposes establishing automated code linting and formatting to maintain consistent code quality across multiple developers.

## What's Proposed

**New Document:** `docs/developer/linting-proposal.md` (480 lines)

A comprehensive proposal for implementing coding standards with:
- Tool selection and configuration
- 7-phase implementation plan
- Conservative rollout strategy
- Progress tracking

### Python
- ~~**Black**: Code formatter~~ (line-length=120)
- ~~**Flake8**: Style checker~~ (lenient configuration)
- ~~**Pylint**: Deferred for future consideration~~
- **Ruff** much faster - replaces Black and Flke8
- **mypy** Type checking

### Other Formats
- **yamllint**: YAML validation
- **markdownlint**: Markdown formatting
- **codespell**: Spell checking

## Key Decisions

1. **Line length:** 120 characters
2. **Initial scope:** __init__.py and plugin_solax.py only (test on subset)
3. **Flake8 rules:** Lenient initially, tighten gradually
4. **Enforcement:** Optional initially, evaluate after Phase 6
5. **Pylint:** Deferred until Black/Flake8 are stable

### 7-Phase Rollout

1. **Phase 1:** Setup configuration files and manual scripts
2. **Phase 2:** Test on 2 core files (__init__.py, plugin_solax.py)
3. **Phase 3:** Add GitHub Actions (informational, won't block PRs)
4. **Phase 4:** Expand to remaining plugin files
5. **Phase 5:** Add optional pre-commit hooks
6. **Phase 6:** Apply to documentation files
7. **Phase 7:** Evaluate feedback, decide if making required

**Conservative approach:** Start small, validate, expand gradually.

### Configuration Files (Examples)
- `.editorconfig` - IDE consistency
- `pyproject.toml` - Black configuration
- `.flake8` - Linter rules
- `.yamllint` - YAML rules
- `.markdownlint.json` - Markdown rules
- `.codespellrc` - Spelling dictionary
- `.pre-commit-config.yaml` - Pre-commit hooks
- `.github/workflows/lint.yml` - GitHub Actions

### Manual Scripts
- `scripts/lint.sh` - Run all checks on demand
- `scripts/format.sh` - Auto-fix formatting issues

### Code Quality
- Consistent style across all files
- Catch common errors early
- Improve readability

### Developer Experience
- Clear standards for contributors
- Automated feedback in PRs (when enabled)
- Less manual code review time

[...]

PR #1833, opened 2026-02-06
feat: Add mypy type checking infrastructure (Phase 1)

fxstein:
## Problem

The integration lacks static type checking, leading to runtime type errors that could be caught during development, unclear function signatures, and reduced IDE effectiveness.

## Solution

Add mypy infrastructure with lenient Phase 1 configuration:
- Enables type checking without blocking development
- Creates baseline for measuring progress
- Opt-in via manual pre-commit hook
- Prepares for incremental strictness increases

### 1. Add mypy Dependency

**File:** `pyproject.toml`

```toml
[dependency-groups]
dev = [
    "mypy>=1.8.0",  # Added
]
```

### 2. Configure mypy (Lenient Settings)

**File:** `pyproject.toml`

```toml
[tool.mypy]
python_version = "3.12"
strict = false
ignore_missing_imports = true
check_untyped_defs = true
files = ["custom_components/solax_modbus"]
exclude = ["^tests/", "^scripts/", "^\\.venv/"]
```

**Philosophy:** Start permissive to avoid workflow disruption. Strictness increases in future phases.

### 3. Add Pre-commit Hook (Manual Stage)

**File:** `.pre-commit-config.yaml`

```yaml
- repo: local
  hooks:
    - id: mypy
      name: mypy type checking
      entry: uv run mypy
      language: system
      args: ["custom_components/solax_modbus"]
      stages: [manual]  # Opt-in: pre-commit run mypy --hook-stage manual
```

## Baseline Results

**Command:** `uv run mypy custom_components/solax_modbus`

**Output:** 1597 type errors across 26 files (expected for lenient config)

**Top Files:**
- `plugin_solax.py`: 842 errors
- `const.py`: 178 errors  
- `plugin_growatt.py`: 156 errors

## Testing

- ✅ All 250 existing tests pass
- ✅ No integration issues
- ✅ mypy runs successfully
- ✅ Pre-commit hook works as expected
- ✅ No workflow disruption

## Compatibility

- No code changes to integration logic
- All existing functionality preserved
- Development workflow unchanged (mypy is opt-in)
- No new runtime dependencies

## Next Steps

Phase 2 will add type hints to critical interfaces (`const.py`, `config_flow.py`, `__init__.py`, platform files), targeting ~50% error reduction.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,17 @@ repos:
         # Exclude translation files except English (en.json)
         exclude: ^custom_components/solax_modbus/translations/(?!en\.json).+
 
+      - id: mypy
+        name: mypy type checking
+        entry: uv run mypy
+        language: system
+        types: [python]
+        pass_filenames: false
+        args: ["custom_components/solax_modbus"]
+        # Set verbose to see output but don't fail
+        verbose: true
+        stages: [manual]  # Only run when explicitly called
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.14.14
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ dev = [
     "pre-commit>=4.0.0",
     "codespell>=2.2.6",
     "ruff>=0.14.14",
+    "mypy>=1.8.0",
 ]
 test = [
     "pytest>=8.0.0",
@@ -50,4 +51,44 @@ asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "function"
 markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+]
+
+[tool.mypy]
+# Python version targeting HA 2024.9.0+ (Python 3.12)
+python_version = "3.12"
+
+# Start with lenient settings - Phase 1
+strict = false
+warn_return_any = false
+warn_unused_configs = true
+
+# Ignore missing imports initially to avoid dependency issues
+ignore_missing_imports = true
+
+# Allow untyped definitions for now
+disallow_untyped_defs = false
+disallow_untyped_calls = false
+disallow_incomplete_defs = false
+
+# Basic checks that should always pass
+check_untyped_defs = false
+warn_redundant_casts = true
+warn_unused_ignores = true
+no_implicit_reexport = false
+
+# Output configuration
+show_error_codes = true
+show_column_numbers = true
+pretty = true
+
+# Files to check
+files = [
+    "custom_components/solax_modbus",
+]
+
+# Exclude patterns
+exclude = [
+    "^tests/",
+    "^scripts/",
+    "^\\.venv/",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -2491,6 +2491,7 @@ dependencies = [
 [package.dev-dependencies]
 dev = [
     { name = "codespell" },
+    { name = "mypy" },
     { name = "pre-commit" },
     { name = "ruff" },
 ]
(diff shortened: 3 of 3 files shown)


fxstein:
This PR starts the mypy type checking implementation as originally layed out in #1701

I have laid this out in 4 incremental phases to avoid too many changes at once. 

Another side benefit: Once mypy has been fully implemented we can consider adding pydantic to all our modbus communications. This will allow us to catch pretty much ALL communications errors - as all registers would be checked for validity.

infradom:
Just a remark:  I like python because it did not use static type checking. If I found static typing important, I would choose another language.

Dynamic/Flexible typing  can be very useful. I have no problem with this PR, but I fear that in the core part of this integration, we still rely a lot on flexible types (Example : the ugly solution for registers that contain 2 entities in the upper/lower byte).  Agree that we should fix this in the style bitfields solved a similar problem.
So without looking at this into detail I am worried about potential side effects of this change.

fxstein:
Without type checking this becomes a black hole of untraceable bugs. Just going through this process it identified several code bugs and execution issues. Several of the out of bound sensor readings, especially during startup are timing related when the types are NOT checked and uninitialized bad data gets passed into functions and plugins.

rdehuyss:
What is the opinion of @wills106 on this? 

I think it do makes sense to have typing support, especially in such a complex modbus integration.

wills106:
If it can help to make maintaining and development easier then I support the changes.

I'm just not 100% how these parts are all meant to work. (Phase1, Phase A etc)
Is VSCode meant to just fix things or does GitHub desktop do the fixing?

For background my "development setup" is just VSCode in Windows and either live testing files on my main HA running in Docker or when I don't mess the VM up I have a simple HA setup in a VM for testing, using a Modbus proxy. I don't have a development setup as indicated at https://developers.home-assistant.io/docs/development_environment

There was a PR #1799 to fix the line endings.
I have been working on attempting to implement `time.py` again locally and I have been greeted with:
This diff contains a change in line endings from 'CRLF' to 'LF'. In GitHub Desktop

image

Is that GitHub desktop following these various rules, or is my local setup messing with the line endings again because either I haven't got VSCode setup properly, or that I don't have a proper Dev environment?

wills106:
After making a mess of the time branch by uploading files manually. I now understand that at least .gitattributes works only when you use GitHub Desktop and presumably cli as well. So the git process cleans up the line endings.

rdehuyss:
Using Mac here so the issue with line endings are only a memory to me 😉...

wills106:
I'm keeping a keen eye out for the Studio M5's
Already picked up an M5 iPad last year.

rdehuyss:
> I'm keeping a keen eye out for the Studio M5's Already picked up an M5 iPad last year.

You and me both 🤣

fxstein:
We all are looking for the M5s to hit the stores. Gotta say even my office
M4 Mini is an amazing Machine.

fxstein:
> If it can help to make maintaining and development easier then I support the changes.
> 
> I'm just not 100% how these parts are all meant to work. (Phase1, Phase A etc) Is VSCode meant to just fix things or does GitHub desktop do the fixing?
> 
> For background my "development setup" is just VSCode in Windows and either live testing files on my main HA running in Docker or when I don't mess the VM up I have a simple HA setup in a VM for testing, using a Modbus proxy. I don't have a development setup as indicated at https://developers.home-assistant.io/docs/development_environment
> 
> There was a PR #1799 to fix the line endings. I have been working on attempting to implement `time.py` again locally and I have been greeted with: This diff contains a change in line endings from 'CRLF' to 'LF'. In GitHub Desktop
> 
> image
> Is that GitHub desktop following these various rules, or is my local setup messing with the line endings again because either I haven't got VSCode setup properly, or that I don't have a proper Dev environment?

FYI - Just change the line ending setting in VSCode and it will be a thing of the past...

wills106:
I actually noticed that setting later yesterday when I was setting up code-server to have a play with.

PR #1839, opened 2026-02-10
Fix Solax Firmware Version Number Display (Gen4/Gen5)

TCWORLD:
The firmware version numbers for the X1-IES have always shown incorrect values on the device info in HA. From reading the documentation it seems this is caused simply by incorrect string formatting and a missing major version number.

1. The minor numbers from the docs are always 2 decimal digit values (0-99). A value of say 4 should actually be rendered as 04 when displayed as a version number. This may apply to the other generation versions too.

2. Properly render the DSP version number for Gen 5 by adding `firmware_dsp_major` internal register. Note this removes the disabled-by-default `firmware_DSP_major_version` diagnostic entity to make way for cleaner rendering.

With this change, the version shown in HA now matches the version shown in SolaX Cloud:
image

image

I have a suspicion the Gen 5 version registers should also work for Gen 6 inverters, but that has not been tested. I think the same major version number registers should also work for the Gen 4 inverters too so probably worth testing.


infradom:
Tested on hybrid X3 gen4. seems to behave like before.
On my inverter, the display shows different DSP/ARM version numbers 2.xx , while the integration shows 1.xx versions.
I cannot check the cloud application, since I am disconnected from the cloud since a while. I might reconnect later to check.

TCWORLD:
@infradom Apologies, I hadn't enabled it for GEN4 as I couldn't test.

Could you try allowing the `firmware_dsp_major` and `firmware_arm_major` entities for GEN4, and change `value_function_software_version_g4` to:

```
def value_function_software_version_g5(initval, descr, datadict):
    return f"DSP {datadict.get('firmware_dsp_major')}.{datadict.get('firmware_dsp'):>02} ARM {datadict.get('firmware_arm_major')}.{datadict.get('firmware_arm'):>02}"
```

If that corrects the version number display, I'll update the commit. I believe the major number entities should exist.

infradom:
> Could you try allowing the `firmware_dsp_major` and `firmware_arm_major` entities for GEN4, and change `value_function_software_version_g4` to: .....

I enabled both entities for GEN4 and adapted the value function. I also re-enabled the cloud access to verify. The version numbers correspond to the cloud version numbers ARM 1.54; DSP 1.81. The device display still shows other numbers starting with 2.xx but this discrepancy was created after a firmware update.

Conclusion: looks fine for GEN4 after the additional changes
Thanks !

wills106:
Thanks for the last push, I was just about to start correcting to conflict.

Issue #1863, opened 2026-03-03
[Bug]: 'Solinteg' - 'MHT-10K-25' Controls not working, one shows wrong type.


sixuniform:
### Describe the bug

<img width="309" height="323" alt="Image" src="https://github.com/user-attachments/assets/e885ac0e-98bc-40d3-b248-e4f9e5f92d47" />

"number.solax_inverter_export_limit" is shown in %, but when it shows 130%, the inverter is set to 13.0kW. import limit shows kW correctly.

Changing any of:
number.solax_inverter_export_limit
number.solax_inverter_import_limit 
select.solax_inverter_export_limit_switch
select.solax_inverter_import_limit_switch

.. will not work, the changes are never reflected on the inverter, and eventually re-sets to the inverter values in the SolaX-integration.

### Integration Version

2026.03.1

### Homeassistant core version

2026.2.3

### Inverter brand

Solinteg - Not sure what plugin, I picked "Solinteg" while installing.

### Plugin used

plugin_solis.py

### Serial prefix

A11250

### Connection Method

TCP Modbus, Cabled ethernet. Solinteg default dongle.

### Detailed Error Log

I am not finding any log events related to this. No lines are added to the raw logs while I try to change these values.

user86000:
MHT-10K-25   second pv string, backup power and house load are also broken  on both new versions 2026.3.1 and 2026.3.2 with 2026.1.2 all sensors are working correct

[home-assistant_solax_modbus_2026-03-04T13-31-42.566Z.log](https://github.com/user-attachments/files/25742236/home-assistant_solax_modbus_2026-03-04T13-31-42.566Z.log)


<img width="404" height="399" alt="Image" src="https://github.com/user-attachments/assets/a281e903-fe95-4cc7-ac13-d922b45afe58" />

<img width="379" height="50" alt="Image" src="https://github.com/user-attachments/assets/c061b0fa-e584-47a9-a6b1-4a13350c9f9c" />

<img width="382" height="398" alt="Image" src="https://github.com/user-attachments/assets/ade5c2f5-9256-475d-8a8b-2552fc4ecc1e" />

gec75:
I didn't see the bug report, but it happened to me also.
#1868

sixuniform:
> I didn't see the bug report, but it happened to me also. [#1868](https://github.com/wills106/homeassistant-solax-modbus/pull/1868)

[...]

PR #1868, opened 2026-03-04
missing entities: fix immutable updates

gec75:
due to recent changes....


gec75:
Found later the bug report #1863

Issue #1858, opened 2026-03-03
[Bug]: solax_battery_voltage_cell_difference shows 0V after upgrade to 2026.03.1


nettle-m:
### Describe the bug

solax_battery_voltage_cell_difference shows 0V incorectlly all the time
solax_cell_voltage_low reads ok
solax_cell_voltage_high also reads ok

### Integration Version

2026.03.1

### Homeassistant core version

2026.2.3

### Inverter brand

solax x3 hybrid g4

### Plugin used

plugin_solax.py

### Serial prefix

H34A15

### Inverter firmware versions

DSP 1.48 ARM 1.50

### Connection Method

modbus TCP, solax wifi dongle

### Dongle firmware

3.019.01

misakm:
I have the same issue
SolaX Inverter Modbus Verze: 2026.03.1
HA 2026.2.3
X3-Hybrid-10kW
od SolaX Power
Firmware: DSP 1.55 ARM 1.51
Hardware: Gen4

itrankolov:
I have the same issue.
Solax X3 Ultra 30kW
2026.03.1

TCWORLD:
What are the values you have for `solax_cell_voltage_low` and `solax_cell_voltage_high`?

misakm:
Here is sample
<img width="1237" height="513" alt="Image" src="https://github.com/user-attachments/assets/8870cad4-408f-4de4-b400-dacb2472610d" />

<img width="422" height="159" alt="Image" src="https://github.com/user-attachments/assets/0c9e108a-f01c-4582-ac4b-9069f2381843" />

TCWORLD:
Thanks. I can see what the problem is. The calculation function seems to have been modified to accidentally cast to an integer.

misakm:
Thank you for fixing this.

wills106:
Just released [2026.03.3](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2026.03.3) with the fix

PR #1869, opened 2026-03-04
Fix SolaX Battery Voltage Cell Difference Calculation

TCWORLD:
The battery_voltage_cell_difference handles fractional values not integers.

The cast to int() appears to have been incorrectly applied during the typing clean up phase.


TCWORLD:
Fixes #1858

PR #1884, opened 2026-03-08
feat(solinteg): add Charge-Discharge working mode and power target control

IZZE2000:
## Summary

- Adds working mode `0x303` (Charge-Discharge / EMS mode) to `SELECT_TYPES` so users can select it from the Home Assistant UI
- Adds the same mode to the internal `working_mode` sensor scale dict for state tracking
- Adds `battery_charge_discharge_power_target` number entity (register 50207, `REGISTER_S16`, scale 0.01 kW) for direct charge/discharge power control
- Adds a corresponding internal sensor to poll the register value

## Context

Register `50207` on Solinteg hybrid inverters holds a signed 16-bit power target for the EMS Charge-Discharge mode (`0x303`). A value of `100` = 1 kW charge, `-100` = 1 kW discharge. This allows external systems (e.g. Home Assistant automations) to directly set the inverter charge/discharge power target when in EMS mode.

Working mode `0x303` is documented in the Solinteg Modbus protocol as an EMS-controlled Charge-Discharge mode, distinct from the standard modes (General, Economic, PeakShift, etc.).

## Testing

Tested on a Solinteg hybrid inverter connected via ModbusTCP. Both the mode selection and the power target number entity function correctly.


IZZE2000:
Closing in favor of updated PR with corrected implementation (raw register values for number entity, matching sensor unit/scale).

PR #1885, opened 2026-03-08
feat(solinteg): add Charge-Discharge working mode and power target control

IZZE2000:
## Summary

- Adds working mode `0x303` (Charge-Discharge / EMS mode) to `SELECT_TYPES` so users can select it from the Home Assistant UI
- Adds the same mode to the internal `working_mode` sensor scale dict for state tracking
- Adds `battery_charge_discharge_power_target` number entity (register 50207, `REGISTER_S16`) for direct charge/discharge power control
- Adds a matching internal sensor with consistent unit/scale for correct feedback loop

## Context

Register `50207` on Solinteg hybrid inverters holds a signed 16-bit power target for the EMS Charge-Discharge mode (`0x303`). The register unit is `x0.01 kW`, meaning register value `300` = 3 kW charge, `-200` = 2 kW discharge.

The number entity uses raw register values (`native_unit_of_measurement="x0.01 kW"`, `scale=1`) so the internal sensor feedback matches what was written. The backing sensor uses the same unit/scale for a consistent round-trip.

Working mode `0x303` is documented in the Solinteg Modbus protocol as an EMS-controlled Charge-Discharge mode, distinct from the standard modes (General, Economic, PeakShift, etc.).

## Testing

Tested on a Solinteg hybrid inverter connected via ModbusTCP. Both the mode selection and the power target number entity function correctly.

### DIFF
--- a/custom_components/solax_modbus/plugin_solinteg.py
+++ b/custom_components/solax_modbus/plugin_solinteg.py
@@ -1,6 +1,6 @@
 import logging
 from dataclasses import dataclass, field, replace
-from typing import Any, cast
+from typing import Any
 
 from homeassistant.components.number import NumberDeviceClass, NumberMode
 from homeassistant.components.sensor import SensorDeviceClass, SensorStateClass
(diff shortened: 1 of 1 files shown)


IZZE2000:
...working on the 2 failing checks...

Issue #1892, opened 2026-03-12
[Bug]: Value function exception


infradom:
### Describe the bug

There is currently an exception in value function evaluations caused by an int conversion.

> def value_function_remotecontrol_current_pushmode_power(initval: int, descr: Any, datadict: dict[str, Any]) -> int | float:
>     return int(datadict.get(descr.key, 0))
> 
> def value_function_remotecontrol_current_pv_power_limit(initval: int, descr: Any, datadict: dict[str, Any]) -> int | float:
>     return int(datadict.get(descr.key, 0))

This may  break the evaluation of other value functions.
I suggest to remove the int() function and allow None also as return value:

> def value_function_remotecontrol_current_pushmode_power(initval: int, descr: Any, datadict: dict[str, Any]) -> int | float | None
>     return datadict.get(descr.key, 0)
> 
> 
> def value_function_remotecontrol_current_pv_power_limit(initval: int, descr: Any, datadict: dict[str, Any]) -> int | float | None:
>     return datadict.get(descr.key, 0)

### Integration Version

2026.3.x

### Homeassistant core version

any

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

all

### Inverter firmware versions

all

### Connection Method

not relevant

### Dongle firmware

not relevant

infradom:
Will be solved with PR #1893

rosenrot00:
@infradom could you close this one, I guess it is fixed.

Issue #1890, opened 2026-03-11
[Bug]: Grid Export/Import intermittently shows 0W + TypeError NoneType firmware_dsp


m0nn3:
### Describe the bug

Two related issues observed with SolaX Inverter Modbus integration on HA 2026.3.1:

Issue 1 – Grid Export/Import drops to 0W intermittently
sensor.solax_grid_export and sensor.solax_grid_import regularly show 0W despite active grid activity (e.g. 6kW → 0W → 7kW → 0W). PV sensors remain stable at the same time. The raw sensor returns numeric 0 (not unavailable/unknown), so template workarounds cannot filter it.

Issue 2 – TypeError in firmware version parsing
Recurring error (929 occurrences since 2026-03-10):

Something went wrong reading from modbus: unsupported format string passed to NoneType.__format__
File "plugin_solax.py", line 1105, in value_function_software_version_g4
return f"DSP {datadict.get('firmware_dsp_major')}.{datadict.get('firmware_dsp'):>02} ...
TypeError: unsupported format string passed to NoneType.__format__

Also observed in logs (Modbus errors):

pymodbus returned isError True
device: 1 address: 22, 28, 11, 10
device: 2 address: 10

HomeAssistant Settings:
Polling: Master 30s/15s/10s, Slave 20s/20s/20s, Timeout Master 10s / Slave 5s

Gerne auch in deutsch Antworten :)

### Integration Version

Version 2026.03.3

### Homeassistant core version

2026.3.1

### Inverter brand

Q.VOLT HYB-G3

### Plugin used

plugin_solax.py

### Serial prefix

Master: SLB8CZ and Slave: SYTWEN

### Inverter firmware versions

Master: ARM 1.56-1.08 DSP 1.58 and Slave: ARM 1.56-1.12 DSP 1.58

### Connection Method

Protoss PE11 Firmware: 1.40.5

### Dongle firmware

LAN Modul 3.008.02

### Detailed Error Log

<img width="603" height="340" alt="Image" src="https://github.com/user-attachments/assets/cfb44985-afc9-4afe-bffe-a4cb7a8a3f3b" />

m0nn3:
I am seeing unstable readings with the SolaX Modbus custom integration in Home Assistant 2026.3.1.

Symptoms:
- Some PV power sensors (for example sensor.solax_pv_power_1 and sensor.solax_pv_power_2) become unavailable/unknown for long periods.
[...]

Issue #1889, opened 2026-03-11
[Bug]: Inverter stops updating grid import


wildjokerss:
### Describe the bug

Hello,

I seem to have found a issue with the grid import data.
The sensor up until now, used to update all night long, alowing me to have hourly information of the energy usage.
Now the sensor stops, and kicks on again either randomly or when I restart the integration.

<img width="2297" height="590" alt="Image" src="https://github.com/user-attachments/assets/a50713eb-dc3d-41b7-8a70-09698095cef7" />

<img width="2291" height="481" alt="Image" src="https://github.com/user-attachments/assets/ea0c4834-27c4-430f-be6a-0af9436b10a2" />
<img width="2326" height="491" alt="Image" src="https://github.com/user-attachments/assets/347adb9e-b210-4a1a-be42-a6b102f67a7d" />
<img width="2296" height="483" alt="Image" src="https://github.com/user-attachments/assets/4843b247-83bc-45b7-a57e-f85b27923050" />

The images above show the issue that results from this, the energy graphs go all bonkers :D

Additionally, I also noticed at the same time, that the sensors are taking way longer to update. I have a EVCC integration to charge my tesla on solar only and I rely on the Solax grid import / export sensors. they were previously updating fast, about 5 seconds or so, less than 10. Now I've seen the import sensor and export beeing over a minute without refreshing, this is a bit bad as it makes EVCC go wild.

I'm on the latest version of everything, HA, Solax Integration and EVCC (guess this one's irrelevant here)

I also found this in the logs:

This error originated from a custom integration.

```
Logger: custom_components.solax_modbus
Source: custom_components/solax_modbus/__init__.py:1340
integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 09:10:29 (2 occurrences)
Last logged: 09:10:58

Something went wrong reading from modbus: unsupported format string passed to NoneType.__format__
Traceback (most recent call last):
  File "/config/custom_components/solax_modbus/__init__.py", line 1340, in async_read_modbus_data
    res = await self.async_read_modbus_registers_all(group)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/solax_modbus/__init__.py", line 1588, in async_read_modbus_registers_all
    data[key] = descr.value_function(0, descr, data)
                ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
[...]

PR #1893, opened 2026-03-12
correct bug #1892: allow None when autorepeat is not active

infradom:
The int() conversion caused an exception, that may have interrupted the value_function computation of other entities also (unsure)

This is an example where the stricter type checking introduces new errors: to be compliant with the type declaration, an int() conversion was added recently; this int conversion fails if the value is None, which is a perfectly valid value for this entity.

One could add code to check the type first, but this only adds runtime load that has no additional value at all imho.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -898,10 +898,12 @@ def value_function_remotecontrol_autorepeat_remaining(initval: int, descr: Any, 
 
 
 def value_function_remotecontrol_current_pushmode_power(initval: int, descr: Any, datadict: dict[str, Any]) -> Any:
+    # do not convert to int(); None is a valid value and is returned if VPP is inactive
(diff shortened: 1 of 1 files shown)


infradom:
Might solve issue #1890 or #1889 also, as this error may have interrupted the computation of other value_functions

The error may have been a consequence of the stricter type checking introduced recently. Although this seems to be a good thing on first sight, the newly introduced convertions to int() may throw exceptions if a datadict entry is deliberately set no None. (like the ones addressed in this PR).

If a datadict entry is None, a datedict.get("entry", 0) will not return 0, but None, so a subsequent conversion to int will throw an exception. 
This exception can cause other value functions not to be evaluated .

I fear that I have only detected 2 instances of such an error, and that we may find more of these side effects.

I corrected this by allowing the value function to return any and by removing the int() call.
Other solutions would need to check the type at runtime before calling int, which causes runtime overhead (without added value in this case imho)

Issue #1871, opened 2026-03-04
[Bug]: Mode 8 Freeze Discharge not charging battery


SupraJames:
### Describe the bug

Hi All, it's possible I am misunderstanding something here but I'm trying to work through it. Nice day of sunshine today, and I have predbat doing it's thing. I had an issue this morning where the PV output was being capped at about 4kW for some time. My battery SoC was around 80%, house load 1.2kW, rest out to grid.

It's easy to see in this picture:

<img width="471" height="354" alt="Image" src="https://github.com/user-attachments/assets/c41bcc66-b746-45bc-a6f3-2f823279944b" />

The small 'blip' in the middle is me disabling Mode 8 control and seeing that the PV output immediately went to over 5kW and the battery started to charge. So it's clear that at this time the output was being limited, and my battery was not charging.

During this time, predbat was in Freeze Export mode. I am using Mode 8 control script, and this maps through to Freeze Discharge in the script, then **Export-First Battery Limit** as the Mode 8 submode.

When I read the documentation for this sub mode, it states:

_This submode prioritises feeding excess energy to the grid up to the defined export limit. Surplus power beyond that is used to charge the battery, with charging power adjustable via the Export-First Battery Charge Limit (mode 8/9) parameter._

This parameter is set to 100% in my config (I never touched it)

<img width="718" height="135" alt="Image" src="https://github.com/user-attachments/assets/00bfed0a-37bb-47fb-add2-4dcb80b4d56a" />

So, I am pretty confused as to why this mode isn't charging the battery as it's supposed to.

As an aside, I do see the new Mode 8 submode added yesterday called **Enabled Feedin Priority** was planning to try that - but honestly, I am struggling to work out the difference between the two submodes.

Everything started working properly once Predbat went to normal Demand mode and Mode 8 was disabled automatically.

### Integration Version

2026.03.2

### Homeassistant core version

2026.2.3

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H5372A

### Inverter firmware versions

DSP 025.02 ARM 020.02

### Connection Method

Waveshare Modbus

### Dongle firmware

N/A

SupraJames:
Oh damn, I have just spotted something in my screenshot!

<img width="718" height="135" alt="Image" src="https://github.com/user-attachments/assets/4ebd9d3f-8a7f-4956-a249-5d35afc082c7" />

The Export Control User Limit is set to 6000 (no idea where that figure came from)

[...]

Issue #1730, opened 2025-11-13
[Bug]: Mode 8 Export-First Battery Limit is limiting solar output


TCWORLD:
### Describe the bug

For some reason the Export-First Battery Limit mode appears to be limiting the solar output into a very odd Saw-tooth pattern.

At first I though it was the inverter doing it, but when the house load increases (deficit mode), the solar output stabilises, and similarly when the mode is set to disabled, the rollercoaster stops.

<img width="1405" height="939" alt="Image" src="https://github.com/user-attachments/assets/e29b223a-7594-4f30-9b11-74d45aee6e83" />

### Integration Version

2025.11.1

### Homeassistant core version

2025.11.1

### Inverter brand

SolaX

### Plugin used

plugin_solax.py

### Serial prefix

H5602A

### Inverter firmware versions

ARM 14.06 DSP 18.06

### Connection Method

Waveshare RS485 to RJ45 Ethernet Module with PoE

### Dongle firmware

N/A

### Detailed Error Log

To follow. Unfortunately debug logging was disabled.
Will update tomorrow as solar output has dropped too much now to test.

rosenrot00:
Thanks for reporting this. I didn't see that on my end. Are you using the CT or the smart meter in your setup. Also what looks weird to me is that there is grid export and import at the same time? How does the houseload look like?

Unfortunately, this time of the year I can't really run much tests as the conditions to test are limited to good weather and like 4h a day :)

TCWORLD:
@rosenrot00 It's using a CT on the grid connection. 

Not exporting and importing simultaneously - I think that's an optical illusion with the purple line of the solar output and green of the inverter output. The blue line is flat (apart from a tiny spike when the load jumps up in the middle).

I'll try and collect some debug logs tomorrow - I'm also in a similar boat with only a few hours window of solar.

TCWORLD:
Some hopefully clearer charts:

Export/House Load/Solar

<img width="1378" height="542" alt="Image" src="https://github.com/user-attachments/assets/6d927f52-b6be-4437-b5eb-4e217d3aba2c" />

Import:

<img width="1381" height="536" alt="Image" src="https://github.com/user-attachments/assets/7ffab650-8808-4ac8-8c09-4fd8350217db" />

Battery/Inverter:

<img width="1387" height="551" alt="Image" src="https://github.com/user-attachments/assets/29e5f8b6-9ff1-49e5-b8fa-82a49da69e03" />

[...]

PR #1895, opened 2026-03-14
Improvements to Mode 8 Export First Remote Control Loop

TCWORLD:
Co-authored by @rosenrot00.

This is linked to and fixes issues #1871 and #1730

The aim here is to improve the control loop for the Mode 8 Export-First loop to try and fix some problems that were causing the PV output to be artificially limited rather than charging the battery. So far two issues have been identified:

1. We need to account for the inverter output power rating when calculating the export limit, which includes factoring in house load. Use the new `inverter_power_type` entity to calculate the maximum possible export and use that in conjunction with the fixed limit.
2. The inverter is also trying to manage export limit itself, so if our control loop is working at the inverter set limit, the two fight each other causing the PV to slowly ramp down. We resolve this by adding a little headroom in the export limit, so we are working to a lower level.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -688,12 +688,12 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
 
         # Debug inputs
         _LOGGER.debug(
-            f"[Mode8 Export-First] inputs pv={pv}W hl={houseload}W hl_alt={houseload_alt}W (using hl_alt) exp_lim={export_limit}W imp_lim={import_limit}W "
+            f"[Mode8 Export-First] inputs pv={pv}W hl={houseload}W hl_alt={houseload_alt}W (using hl) exp_lim={export_limit}W imp_lim={import_limit}W "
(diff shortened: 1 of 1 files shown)


wills106:
Just looking at catching up on all of these PR's...

So just wondering if this one is ready to move from draft?

Also @TCWORLD & @rosenrot00 have either of you had a look at the TimeV3 branch?

rosenrot00:
For me this is ready, I'm running this as the old implementation doesn't work for me.

I haven't looked at the TimeV3 branch yet - will do.

wills106:
TimeV3 will need a bit of updating to include the other changes made. But the time entities, have been working fine for me. just unsure on the changes for Autorepeat I had to make.

rosenrot00:
> TimeV3 will need a bit of updating to include the other changes made. But the time entities, have been working fine for me. just unsure on the changes for Autorepeat I had to make.

I just screened it and I guess I understand what you try to do there. Great idea!

I think merging some PRs would help to re-gain focus. Meanwhile I'm getting confused with all the branches I have created. I would be surprised if there is nothing conflicting. Don't get me wrong, I know that there are more important things in life!

wills106:
I'll try and merge everything together. Then I'll try and update TimeV3 as it's already 14 commits behind.

PR #1896, opened 2026-03-14
Solax X1 FIT compatibility proposed changes

stuart-jjj:
This pull request is to make available the controls and sensors with a Solax X1 RetroFit inverter for the below items which were not working using the current integration because they map to register values used by the AC which are not compatible with the FIT. The first 3 are the important ones because they are needed for effective control of the system.

I would appreciate it if you could review these changes and if ok, merge into the main code as this is the only reason I am maintaining this fork. It will benefit other X1 RetroFit Gen 3 owners (admittedly a small group).

- battery minimum capacity gridtied
- forcetime period 1 max capacity
- forcetime period 2 max capacity
- today's yield
- total solar energy
- CT meter setting

### DIFF
--- a/tests/unit/test_solax_fit_inverter.py
+++ b/tests/unit/test_solax_fit_inverter.py
@@ -4,7 +4,8 @@ Covers the Copilot PR review concern about correct entity matching
 for FIT inverters, which use FIT = AC | HYBRID as a semantic alias.
 """
 
-from typing import Any, Sequence, cast
+from collections.abc import Sequence
+from typing import Any, cast
 
 import pytest
 
@@ -42,9 +43,7 @@ PV_SENSOR_KEYS = {
 # Typed view of SENSOR_TYPES so mypy can resolve allowedtypes/blacklist attributes
 # (plugin_base declares SENSOR_TYPES as Sequence[SensorEntityDescription], the HA
 # base class, which does not carry these extra fields)
-SOLAX_SENSOR_TYPES: Sequence[SolaXModbusSensorEntityDescription] = cast(
-    Sequence[SolaXModbusSensorEntityDescription], solax_plugin.SENSOR_TYPES
-)
+SOLAX_SENSOR_TYPES: Sequence[SolaXModbusSensorEntityDescription] = cast(Sequence[SolaXModbusSensorEntityDescription], solax_plugin.SENSOR_TYPES)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


stuart-jjj:
## Update: Addressed Copilot review concerns

I've pushed an additional commit ([5c425bd](https://github.com/stuart-jjj/homeassistant-solax-modbus/commit/5c425bd)) to address the two HIGH concerns from the Copilot review.

### Copilot concern: PV sensor entities incorrectly matching FIT

Because `FIT = AC | HYBRID`, the `PRI` inverter carries the `HYBRID` bit and was matching all 7 bare `allowedtypes=HYBRID` PV input sensor entities:

| Entity | Register |
|--------|----------|
| `pv_voltage_1` | 0x3 |
| `pv_voltage_2` | 0x4 |
| `pv_current_1` | 0x5 |
| `pv_current_2` | 0x6 |
| `pv_power_1` | 0xA |
| `pv_power_2` | 0xB |
| `pv_power_total` | computed |

These have no GEN qualifier, so the GEN group auto-passes for `GEN3|FIT`, causing incorrect register polling on a device that has no DC PV inputs. Fixed by adding `blacklist=["PRI"]` to each of these 7 entities.

A complete grep audit confirms there are no other bare `allowedtypes=HYBRID` entities in the file — every single one now has `blacklist=["PRI"]` directly after it.

### Copilot concern: Dedicated `FIT` type bit

The Copilot suggestion to introduce a new independent `FIT` bit would require touching every entity the FIT *should* see (currently matched via `HYBRID | GEN3`) in both directions — adding `| FIT` to dozens of allowedtypes declarations. The `blacklist` approach is proportionate: the problem set is exactly 7 entities (PV sensors FIT physically cannot have), they're all now excluded, and the scope is fully auditable via grep.

### Summary of all changes in this PR

1. `FIT = AC | HYBRID` constant — semantic alias giving `PRI` both type bits
2. `PRI` serial prefix → `FIT | GEN3 | X1`, model `"X1-FIT"`
3. `global_mppt_function` — `blacklist=["PRI"]` (FIT has no DC MPPT)
4. Registers 0x10C / 0x10D / 0x10E — AC-only duplicates restored with `blacklist=["PRI"]` (FIT uses the HYBRID-layout equivalents at 0x10F / 0x110 / 0x115)
5. `pv_voltage/current/power` entities (×7) — `blacklist=["PRI"]` (FIT has no DC PV input)
6. Comments corrected: `PRI` / `PRI5` entries in MAX_CURRENTS / MAX_EXPORT tables now read "Gen3 X1 FIT" (were incorrectly "Gen4 X1 RetroFit")

stuart-jjj:
## Update: Unit tests for FIT/PRI branch (medium concern)

Added [tests/unit/test_solax_fit_inverter.py](https://github.com/stuart-jjj/homeassistant-solax-modbus/blob/main/tests/unit/test_solax_fit_inverter.py) (commit e639dcf) — 18 tests, all passing.

**Coverage:**

| Category | Tests |
|----------|-------|
| Type constant | `FIT == AC\|HYBRID`, both bits set |
| `matchInverterWithMask` | FIT matches `HYBRID`, `AC`, `HYBRID\|GEN3`; does not match `HYBRID\|GEN4` or X3 entities |
| Blacklist | PV entities blocked for `PRI` serial; non-PRI serials unaffected |
| Regression | Genuine AC-only and HYBRID-only inverters still match/not-match correctly |
| Audit | All 7 bare `allowedtypes=HYBRID` PV sensors confirmed to carry `blacklist=["PRI"]` |
| End-to-end | PV sensors do not match PRI inverter; battery sensors do match |
| Serial detection | `async_determineInverterType` with PRI\* serial → `FIT\|GEN3\|X1` |

The audit test (`test_all_pv_sensors_blacklist_pri`) acts as a regression guard — if a new bare `allowedtypes=HYBRID` PV entity is added in future without the blacklist, the test will fail.

Issue #1897, opened 2026-03-14
[Bug]: Solax X3IES, wrong scaling "solax_export_control_user_limit"


JLinW2025:
### Describe the bug

solax_export_control_user_limit
For example, if I enter or set 100W, the inverter outputs 1000W.

It’s always a factor of 10.

<img width="777" height="276" alt="Image" src="https://github.com/user-attachments/assets/5a962c41-cf5e-475e-a599-01b77ddb5721" />

### Integration Version

2026.03.3

### Homeassistant core version

2026.3.1

### Inverter brand

Solax X3-IES

### Plugin used

plugin_solax.py

### Serial prefix

H35A15

### Inverter firmware versions

DSP 026.02 ARM 018.03

### Connection Method

WIFI+LAN

### Dongle firmware

1.009.02

TCWORLD:
There is a "Diagnostics" entity called "Config Export Control Limit Readscale" (it's disabled by default so you'll need to enable it). 

You can use this to adjust the scaling of the export user limit entity initially. Select either 10 or 0.1 depending on which way it needs scaling. Once you've found which value works, let us know and the code can be updated accordingly for future releases.

JLinW2025:
Setting to 10 works well!

Thank you.

JLinW2025:
I have another issue.
I’m not sure where this belongs, or if it even belongs here at all.
This is my first time using GitHub!

I had the Solax updated with new firmware on Thursday. Now the Solax isn’t responding to register 0x0096 (150 – SolarEnergyToday). It always shows zero.
My dealer is already aware of this. It’s probably a bug in the firmware.

<img width="239" height="359" alt="Image" src="https://github.com/user-attachments/assets/0fa063aa-96d1-4ba9-a614-abd52592a6f7" />

TCWORLD:
Unfortunately not really anything we can do about the inverter firmware returning zeros for these registers. Somebody else reported similar issues with some of the registers reading zero after their inverter firmware was updated.
The actual register address is correct as of the last version of the SolaX modbus RTU docs I've seen.

Issue #1777, opened 2026-01-07
[Bug]: Export Control User Limit multiplied by 10


rgrae81:
### Describe the bug

If I set the User Limit to e.g 400 W then it changes to 4000 W and 5000 to 50000 as shown.

<img width="459" height="209" alt="Image" src="https://github.com/user-attachments/assets/c5fc3c93-748f-40b5-a069-57192634b4f8" />

In SolaxCloud the 4000W and 50000W are confirmed in the settings!
<img width="129" height="129" alt="Image" src="https://github.com/user-attachments/assets/65f8b17f-5ff7-4343-bfb4-56d98d0a7107" />

So the reading is correct but somethings wrong with saving the value

### Integration Version

2025.11.1

### Homeassistant core version

2025.12.5

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

H3BC25

### Inverter firmware versions

ARM 18.02 DSP 17.05

### Connection Method

Solax Pocket WiFi+LAN

### Dongle firmware

1.005.04

### Detailed Error Log

No Error in Logs 
When I changed to 5000W  only this was visible:

2026-01-07 16:06:45.255 INFO (MainThread) [custom_components.solax_modbus.number] writing SolaX export_control_user_limit number register 66 value 5000 after div by readscale 1 scale 1 with mode 1
2026-01-07 16:06:49.628 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: [15s] poll started – cycle #4815
2026-01-07 16:06:49.628 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: scan_group timer initiated refresh_modbus_data call - interval 15
2026-01-07 16:06:49.628 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: device group inverter
2026-01-07 16:06:49.628 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: ** trying to read holding block 0x7d previous res:True
2026-01-07 16:06:49.628 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: READ HOLDING device_id=1 addr=0x7d cnt=91
2026-01-07 16:06:49.701 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: holding block 0x7d read done; new res: True
2026-01-07 16:06:49.701 DEBUG (MainThread) [custom_components.solax_modbus] SolaX: ** trying to read holding block 0x104 previous res:True
[...]

Issue #1636, opened 2025-10-04
[Bug]: export_control_user_limit is 1/10 of real set value


jprates:
### Describe the bug

Probably related to a recent firmware update:

I just noticed today that the setting for export_control_user_limit appears as 1/10 or real set value on inverter side.
My automation has set it to 100W, but the inverter was exporting 1000W, and when checked at Solax app it was indeed set at 1000W, while modbus integration was displaying the HASS set value of 100W.

I then fixed the value to 100W on the app, and modbus updated its displayed value to 10W, maintaining the 1/10 rule.

### Integration Version

2025.09.9

### Homeassistant core version

2025.10.1

### Inverter brand

SolaX Power

### Plugin used

plugin_solax.py

### Serial prefix

H34A10

### Inverter firmware versions

DSP v1.56 ARM v1.54

### Connection Method

SolaX Pocket WiFi 3.0

### Dongle firmware

V3.012.01

### Detailed Error Log

n/a

infradom:
Not a bug, there is a hidden setting to correct such behavior that appears on some systems: config_export_control_limit_readscale. You can set this to either 0.1, 1.0 or 10.0
One of these values should solve your problem.

I have added this in our ReadTheDocs Solax FAQ a couple of minutes ago: 

<img width="1051" height="174" alt="Image" src="https://github.com/user-attachments/assets/9be868b0-7671-4d62-abf5-2561547d325e" />

jprates:
Just for the sake of information in case others come here searching for the solution to the same problem:

I had DSP v1.48 ARM v1.45 prior to updating and the said scale was 1, never touched it, it was working fine.
The need to set scale to 10 only arrived with the update to DSP v1.56 ARM v1.54.

infradom:
Thanks for this feedback; we have indeed seen similar things we could not understand. Sorry for this rather ugly solution, but we could not find a reliable alternative.

PR #1899, opened 2026-03-15
Set Export User Limit Scaling based on X1 or X3

TCWORLD:
At the moment we have a large table of exceptions that misses newer inverters leading to issues with the scaling of this parameter.

Based on the Modbus docs, all GEN4/5/6 inverters *should* use a factor of 1 if an X1 and 10 if an X3. This largely corresponds to the table of exceptions (although for some reason all GEN4 X1 were initially set to 10 and then exceptions down to 1 added).

image

image

However looking back at past issues something may have changed between firmware versions as at least one issue reported it only because an issue after a firmware upgrade.

The docs screenshot above is from the 2025 release of the RTU for X1/X3 Gen 4/5/6 inverters.

As such I've left `config_export_control_limit_readscale` existing. This change may be breaking in that users that used to have to use the readscale "hack" will probably have to remove it. It may require people with old inverter firmwares to use the readscale "hack" where they didn't previously.

I'm open to discussion on whether this change should go ahead - my gut feeling is that we should try to match the latest documentation so that we don't have to keep adding more and more inverters into the exceptions.

---

This is related to #1897, #1777, #1636, and a few others

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1386,23 +1386,6 @@ MAX_EXPORT: list[tuple[str, int | float]] = [
     ### All known Inverters added
 ]
 
-EXPORT_LIMIT_SCALE_EXCEPTIONS = [
-    ("H4", 10),  # assuming all Gen4s
-    ("H34", 10),  # assuming all Gen4s
-    ("H3UE", 10),  # Issue #339, 922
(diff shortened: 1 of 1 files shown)


TCWORLD:
@wills106 Before merging this, please note it's worth considering carefully the comment about how this might be breaking.

wills106:
I think this makes sense, it matches the docs and those tables are getting a mess.

PR #1905, opened 2026-03-20
Solax VPP Mode1 Fixes

### LINKED ISSUE #1844: [Bug]: Solax Mode 1 - Enable No Discharge functionality issue/change
### Describe the bug

Hey folks,

Let me start by thanking you all for your hard work on the integration.  Since we've started to have a few sunny days this year (here in the UK at least) I've noticed an issue when using the "Enable No Discharge" mode 1 remote control functionality. 

Previously I have used this during periods of the day when I didn't want the battery to discharge and to use excess PV power to charge the battery during these periods.  However, this year it seems to be behaving differently.  When excess PV power is produced (above house load) and the "Enabled No Discharge" mode 1 is enabled all excess PV power seems to be being exported. I cant pinpoint the exact date of the change given my PV yield over the winter is so low.

I gave Claude code a crack at debugging the issue, which can be found below.  If you find the use of AI tools offensive please feel free to disregard it.

Is this change in functionality by design?  If so, what can I use in its place?

```
Root Cause: Mode 1 "Enabled No Discharge" exports excess PV instead of charging the battery
The problem is at plugin_solax.py:358:

elif power_control == "Enabled No Discharge":
    # Hold battery level by preventing discharge
    if battery_capacity < 98:
        ap_target = 0 - pv_power          # <-- THE BUG
        power_control = "Enabled Power Control"

Why this is wrong
The key to understanding this is the ap_target sign convention, documented clearly in solax-remote-control-redesigned.md:82-103:

ap_target = 0 means: all house load comes from the grid, no battery activity
ap_target = positive means: grid imports ap_target + house_load (charges battery)
ap_target = negative means: inverter exports power to grid
Since pv_power is always >= 0, the formula ap_target = 0 - pv_power produces a negative ap_target. A negative ap_target tells the inverter to export power to the grid.

Walkthrough with example
Say PV = 5kW, house load = 2kW:

Current code: ap_target = 0 - 5000 = -5000 (export 5kW!)
The inverter sees a -5kW target and tries to push power to the grid
The battery doesn't charge because the inverter is being told to export
What should happen
The goal of "No Discharge" is:

PV supplies house load
Excess PV charges the battery (not export to grid)
If PV is insufficient, grid makes up the deficit
Battery never discharges
The correct formula should be ap_target = 0 (not 0 - pv_power). With ap_target = 0:

House load is supplied from the grid (the implicit house_load behavior)
PV production naturally goes to the battery (since ap_target = battery_charge - pv_power, setting ap_target = 0 means battery_charge = pv_power)
The battery never discharges
This matches the documentation's own description on line 151: ap_target = 0 # Let inverter balance PV/grid naturally — but the code on line 358 says ap_target = 0 - pv_power, which contradicts the documentation.

The discrepancy
Source	Formula
solax-remote-control-redesigned.md line 151	ap_target = 0
plugin_solax.py line 358	ap_target = 0 - pv_power
The documentation says ap_target = 0 but the code says ap_target = 0 - pv_power. The code doesn't match the documentation.

Comparison with Mode 8's "No Discharge"
Mode 8's implementation (line 607-656) handles this correctly — it explicitly routes surplus PV to battery charging (pushmode_power = -desired_charge) and only exports what exceeds the BMS charge limit. Mode 1's implementation simply tells the inverter to export all PV power.

The fix
Line 358 should be:

ap_target = 0

instead of:

ap_target = 0 - pv_power

This makes ap_target = 0, which means PV will naturally charge the battery (since activepower = battery_charge - pv_power, setting activepower to 0 means battery absorbs PV), house load comes from grid, and the battery never discharges. This matches the documented intent and the doc's own formula.
```

### Integration Version

26.01.2

2026.2.2

### Inverter brand

SolaX Power

### Plugin used

plugin_solax.py

### Serial prefix

H4372A

DSP v1.29 ARM v1.27

### Connection Method

Waveshare Adapter 

### Dongle firmware

n/a

### Detailed Error Log

_No response_

### Additional context

_No response_

TCWORLD:
### Enabled Feedin Priority

The AP target should not include battery power charge. This seems to have been a mistake in the last rewrite.

The battery will only ever be charging in this mode if we've hit the inverter/export limit, so trying to increase export won't actually achieve anything.

The battery will only ever be discharging if we're in the deficit mode which is handled separately so the statement does nothing in that mode.

### Enabled No Discharge

During the last rewrite a critical behaviour was lost - the mode is supposed to put excess solar into the battery preferentially to exporting it. However the changes that were made lost that step.

Fix it by setting the AP target to the lower of house load and PV power. If the battery is fully charged then a different pathway is followed so this will still allow export when there is no need to charge the battery.

Fixes #1844.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -348,9 +348,16 @@ def autorepeat_function_remotecontrol_recompute(initval: int, descr: Any, datadi
     elif power_control == "Enabled Feedin Priority":
         # Maximize grid export by using excess PV and battery
         if pv_power > house_load:
-            ap_target = 0 - pv_power - battery_power_charge  # Export all excess
+            # If more power than house load, try to export all excess
+            # power. If this is larger than the inverter/export limit
+            # it will be internally limited and any excess PV will go
+            # to the battery.
+            ap_target = 0 - pv_power
         else:
-            ap_target = 0 - house_load  # Just supply house load
+            # Insufficient power from PV, so just cover house load. The
+            # battery will discharge to cover any deficit where possible.
+            # If the battery is drained completely, grid will cover deficit.
+            ap_target = 0 - house_load
         power_control = "Enabled Power Control"
 
     elif power_control == "Enabled No Discharge":
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -364,10 +364,15 @@ def autorepeat_function_remotecontrol_recompute(initval: int, descr: Any, datadi
         # Hold battery level by preventing discharge
         if battery_capacity < 98:
             # Use PV to supply house load, import from grid only what's needed
-            ap_target = 0 - pv_power
+            # Any excess PV above house load will go to the battery
+            ap_target = 0 - min(house_load, pv_power)
             power_control = "Enabled Power Control"
         else:
-            ap_target = 0  # Battery full, no action needed
+            # When the battery is fully charged (allowing a tolerance to prevent
+            # older inverters shutting down PV), then we can run the default work
+            # mode of the inverter. If the battery discharges during the default
+            # mode, then once below the 98% threshold we will resume VPP.
+            ap_target = 0
             power_control = "Disabled"
 
     elif power_control == "Disabled":


fxstein:
Sorry - have been super busy with work - but I can confirm the bug - seeing the same on my side and was about to put a fix in.

TCWORLD:
No worries. I wonder whether the No Discharge should have a similar check to the Mode 8 loop that works out the maximum battery charge power and allows excess above that to be exported while the battery is charging - with the current change PV will be limited until the battery is charged. Not entirely sure though how to calculate that limit for parallel mode inverters.

Issue #1668, opened 2025-10-21
[Bug]: Solax X3 Ultra inverter missing MPPT3 voltage current and power


endleistung1:
### Describe the bug

Can someone update the solax integration ? i dont have MPPT 3 values Power, Voltage, Current

### Integration Version

 2025.10.5

### Homeassistant core version

2025.10.3

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H3BC20

### Inverter firmware versions

026.01-000.10

### Connection Method

Solax WiFi+LAN

### Dongle firmware

1.007.02

TCWORLD:
Looks like the integration is set up to detect the X3-ULT-20K, while I presume you have the X3-ULT-20KP: It would help if you could confirm what hardware you have.

![Image](https://github.com/user-attachments/assets/06f4e5c6-060f-48d6-b099-c9561ed3ec6f)

As a temporary workaround, you could change line 8878 of the `plugin-solax.py` file to be the same as line 8881 (note the extra `| MPPT3`). 

https://github.com/wills106/homeassistant-solax-modbus/blob/48f5d71c549a01f5625403ffcb35cf0bf0ecf530/custom_components/solax_modbus/plugin_solax.py#L8876-L8881

 It seems that if they both have the same serial number prefix, the devs will have to find another way of distinguishing the two.

wills106:
There is an Input Register that should return the amount of MPPT's the inverter supports. So we could use that to build the logic.

<img width="664" height="25" alt="Image" src="https://github.com/user-attachments/assets/752a843c-ea6a-4b49-b7ef-842babfde434" />

endleistung1:
Hello,

i have for sure the X3-ULT-20KP becouse i have 3 MPPT tracker/PV inputs. in the solax cloud i see also 3 MPPT´s

endleistung1:
can you help? how can we go ahead with this problem?

wills106:
Could you copy over https://github.com/wills106/homeassistant-solax-modbus/blob/main/custom_components/solax_modbus/plugin_solax.py which I have just uploaded and see if "MPPT Count" returns 3 on your Inverter.

endleistung1:
[plugin_solax.py](https://github.com/user-attachments/files/23097016/plugin_solax.py)

endleistung1:
thats my actual one

wills106:
[...]

PR #1910, opened 2026-03-23
Improve inverter detection and fix initial startup refresh timing

rosenrot00:
This PR improves startup behavior in two small ways:

1. Keep reading the serial number as before, but prefer the "new" inverter identification registers for model detection when they are available. If that detection does not match, fall back to the existing serial-prefix logic.
2. Delay the one-time initial refresh until after the platforms are fully set up, so values like export_control_user_limit are available correctly right after startup.


TCWORLD:
I love how inconsistent SolaX inverters seem to be. The inverter ID register on my model returns 500 which doesn't match any of the documented IDs, though the model number string register does return correct.

Screenshot_20260323-215619_ASUS_Launcher (1)

rosenrot00:
> I love how inconsistent SolaX inverters seem to be. The inverter ID register on my model returns 500 which doesn't match any of the documented IDs, though the model number string register does return correct.
> 
> Screenshot_20260323-215619_ASUS_Launcher (1)

At least in this case my changes don't have any effect. Is this because your inverter has a lower modbus protocol version that the documentation has? Or is it just random?

TCWORLD:
Not sure why. The firmware on it was released in December, so should be using the most recent version of the protocol. I'd thought about using it for detection hence adding the diagnostics registers to test out, but was perplexed that it didn't match.

rosenrot00:
For me it fits. We could check for the modbus version, if that would even be a better indicator. have you ever read yours?

Otherwise, we just take it as a flaw. The double detection as implemented now should make things better.

TCWORLD:
Interesting, it comes back as an old protocol version. How odd. I guess that explains why it doesn't give a sensible result!

Screenshot_20260323-222546_ASUS_Launcher (1)

wills106:
Unfortunately some SolaX Inverters are only returning an mppt_count of 2 https://github.com/wills106/homeassistant-solax-modbus/issues/1668#issuecomment-3825105370 when they should be returning 3.
So we may have to set the 15kW Ultra's to always display 3MPPT's but disable the 3rd by default.
Or perhaps a toggle in the setup page to over ride the Inverter has 3 MPPTs's?

rosenrot00:
> Unfortunately some SolaX Inverters are only returning an mppt_count of 2 [#1668 (comment)](https://github.com/wills106/homeassistant-solax-modbus/issues/1668#issuecomment-3825105370) when they should be returning 3. So we may have to set the 15kW Ultra's to always display 3MPPT's but disable the 3rd by default. Or perhaps a toggle in the setup page to over ride the Inverter has 3 MPPTs's?

Thanks for the hint, I wasn't aware. I will think about a two stage config flow that let's one select batteries and MPPT after detection. I will update this PR.

PR #1911, opened 2026-03-23
Add automatic HACS preview releases for pull requests

rosenrot00:
This adds automatic HACS preview releases for pull requests so changes can be tested much faster without waiting for a full stable release.

How it works:
- Each open PR gets its own preview pre-release.
- If the PR is updated, the existing preview is replaced with a new one for the latest PR commit.
- A preview is only created after the ✅ All Tests Passed check succeeds.
- If a PR is closed, its preview release is removed.
- If a PR has already been merged, its preview release is also cleaned up when the next stable release is published.

Why:
- Makes it much easier for users to test PRs directly through HACS.
- Shortens the timespan users have to wait to get their issues solved.
- Users would have a playground to potentially solve issues by themselves. 
- Code quality would not be impacted as @wills106 will always has the last word on what is getting merged.

### DIFF
--- a/.github/workflows/hacs-preview.yml
+++ b/.github/workflows/hacs-preview.yml
@@ -1,0 +1,360 @@
+name: HACS Preview Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: Pull request number to publish as preview
+        required: true
+        type: number
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - closed
+
+permissions:
+  contents: write
+  pull-requests: read
+  checks: read
+
+concurrency:
+  group: hacs-preview-${{ github.event.pull_request.number || inputs.pr_number }}
+  cancel-in-progress: true
+
+jobs:
+  cleanup-preview:
+    name: Clean up preview release
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request_target' && github.event.action == 'closed'
+    steps:
+      - name: Delete preview releases for PR
+        uses: actions/github-script@v8
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const prNumber = process.env.PR_NUMBER;
+
+            const deleteReleaseAndTag = async (release) => {
+              core.info(`Deleting preview release ${release.tag_name}`);
+              await github.rest.repos.deleteRelease({
+                owner,
+                repo,
+                release_id: release.id,
+              });
+              try {
+                await github.rest.git.deleteRef({
+                  owner,
+                  repo,
+                  ref: `tags/${release.tag_name}`,
+                });
+              } catch (error) {
+                core.warning(`Could not delete tag ${release.tag_name}: ${error.message}`);
+              }
+            };
+
+            const releases = await github.paginate(github.rest.repos.listReleases, {
+              owner,
+              repo,
+              per_page: 100,
+            });
+
+            const matching = releases.filter((release) =>
+              release.prerelease &&
+              typeof release.body === "string" &&
+              release.body.includes("Managed-By: hacs-preview") &&
+              release.body.includes(`Preview-PR: ${prNumber}`)
+            );
+
+            for (const release of matching) {
+              await deleteReleaseAndTag(release);
+            }
+
+  publish-preview:
+    name: Publish preview release
+    runs-on: ubuntu-latest
+    if: github.event_name == 'workflow_dispatch' || github.event.action != 'closed'
+    steps:
+      - name: Resolve pull request
+        id: pr
+        uses: actions/github-script@v8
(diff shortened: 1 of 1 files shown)


rosenrot00:
I have given this a good thought and I think it would help to quickly interact with the community to provide help in case of issues. Mostly people posting issues would ever be thought to install a pre-release to try. Once there is a new release or the PR gets closed, the pre-release vanishes.

Even for me this would help to speed up my contributions instead of moving files to my Homeassistant installation manually.

@wills106 @infradom @TCWORLD could you share your thoughts?

TCWORLD:
Couple of thoughts, I haven't looked closely so this may already have been considered:
1. There should probably be some restrictions on users generating a preview release (e.g. workflow triggered by owner/maintained, maybe auto for contributors) - otherwise there's a risk of people opening PRs with malicious code and immediately generating a preview release that other users could unwittingly install.
2. I agree it's a good way of getting users to test if a change resolved an issue/get feedback. I'm not sure how useful it is as a general playground as it's a very roundabout way to copy files across to HA, but that's just me.

TCWORLD:
Just another thought, there's also the slight problem of if there are multiple PRs on the go fixing different issues, the previews would only include the one PRs fixes not the other. That doesn't mean this isn't helpful. 

Perhaps it's worth having a dev branch where PRs are merged to more frequently and then only merged into main on a release. That way you can have a "next release" PR open generating a more complete prerelase.

rosenrot00:
@TCWORLD thanks for your thoughts.

ad 1. I'm very open to any suggestions about a gate to prevent potential malicious code. As the standard users don't have update notifications enabled on pre-releases (typically) I think not a broad audience would ever be affected by such a scenario - but of course - it is very worthwhile considering. We could add a maintainer approval or restrict preview publishing to trusted authors. However, even today I could tell someone to replace a file with my malicious one.

ad2. I saw several issues in which the authors would directly provide the solution with AI. These people would more likely create PRs if they would know they get pre-releases from it. At least it would help.

rosenrot00:
> Just another thought, there's also the slight problem of if there are multiple PRs on the go fixing different issues, the previews would only include the one PRs fixes not the other. That doesn't mean this isn't helpful.
> 
> Perhaps it's worth having a dev branch where PRs are merged to more frequently and then only merged into main on a release. That way you can have a "next release" PR open generating a more complete prerelase.

I thought of this as well, having a preview branch which auto merges open PRs in case of no issues doing so. I see this as an on top help to this individual pre-releases as the beauty here is that we could help people with issues in minutes. Not every user can even copy files around...

But hey, I'm just throwing my opinion here, doesn't mean that I'm right.

TCWORLD:
I do think it's a good idea.

Gating who can generate them is my only strong opinion - a pre release dirrectly downloadable or installable through HACS seems a lot more official than instructing someone to just modify the files. Just my 2p.

rosenrot00:
> I do think it's a good idea.
> 
> Gating who can generate them is my only strong opinion - a pre release dirrectly downloadable or installable through HACS seems a lot more official than instructing someone to just modify the files. Just my 2p.

Thank you! In the end I'm happy to adjust accordingly. @wills106 has the last word on this topic I guess.

wills106:
I think it sounds good in theory, but I am worried about malicious code getting pushed out.
Even if there was a min PR's submitted before auto releases. You could still easily abuse the trust. You could easily send out 3 or 5 good PR's for minor spelling corrections, or Readme updates etc then send out a bad PR on purpose.

Perhaps limit it to plugin_*.py only?
I don't know how easy you can add in filters to block any binary type payloads or any HTML / Java etc?

rosenrot00:
> I think it sounds good in theory, but I am worried about malicious code getting pushed out. Even if there was a min PR's submitted before auto releases. You could still easily abuse the trust. You could easily send out 3 or 5 good PR's for minor spelling corrections, or Readme updates etc then send out a bad PR on purpose.
> 
> Perhaps limit it to plugin_*.py only? I don't know how easy you can add in filters to block any binary type payloads or any HTML / Java etc?

I've given some thought to the potential threat from these pre-releases now.

Actually, if I understand correctly, pre-releases don't show up as updates to users. I read that there should be an option to do so but I couldn't find it on my installation. So pre-release on my end could only be installed by going to HACS - any integration - re-downloads - select pre-release. So from this angle, not much visibility is given to these pre-builds.

Regarding your questions about binaries and so on, not sure if such a check makes sense as one could simply download binaries from any external source once the integration is running and then execute them. Usually HA instances are either running in a docker and shouldn't be running as root - so there is another safety net.

At the same time, I question whether such risks would realistically be detected today – for example, if a single line of Python code were added that downloads and executes a binary within a larger change.

We could proceed pragmatically and test it, with the option to stop at any time if the risks outweigh the benefits. However, open to add safety where it helps.

rosenrot00:
I changed this so that @wills106 can select which PR numbers should receive pre-releases. This should be more secure and maybe the way forward.
https://github.com/wills106/homeassistant-solax-modbus/pull/2054

PR #1916, opened 2026-03-25
Adapted flow to add possiblity to change MPPT and battery counts during inverter setup

rosenrot00:
This changes the integration to use the manually configured mppt_count and battery_count during setup instead of relying only on automatic detection.

What changed:

- configured MPPT count now overrides detected MPPT flags during inverter setup
- battery handling now follows the configured battery count more consistently
- topology filtering was tightened so unused MPPT and battery entities are skipped more reliably

Why:

- some inverters report the wrong MPPT count
- some installations only use part of the available MPPT or battery topology
- manual configuration should be the source of truth in those cases

@wills106 it adds a little more complexity but people with less or more mppt or without battery could now define it themselves. Let me know if it makes sense to you this way.


wills106:
Are you able to look at the failed CI/CD so I can merge this along with #1910

rosenrot00:
> Are you able to look at the failed CI/CD so I can merge this along with #1910

Please ignore this PR as well as #1910 for now. This PR will replace #1910.

Also, we need to test this PR before publishing a release as it changes the flow. In best case any plugin would provide inverter data (MPPT, batteries, max power, max export) which could be adjusted manually.

I guess this needs to be done step by step. I need to do more work here.

TCWORLD:
Worth keeping you in the loop  We should probably look at adding the ability to control sensor maps based on version numbers rather than just the GEN flags. 

There's a sensor called "Modbus Protocol Version"  now which provides a useful way of figuring out what register maps to used. 

If this value is >=100, then the latest protocol documents that have been circulating apply. Otherwise we can't know reliably whether registers can be used. The GENx can't necessarily be used to separate this because it's firmware version dependent 

I have a feeling we're going to have more and more issues going forward - like the PV power total issue that's just popped up where it works for the VAST but not G4-Pro because of different firmware flavours. 

I suspect the inverter ID number lookup in the recent protocol document applies only when the protocol version is >=100.

PR #1918, opened 2026-03-26
Correct scaling on Export User Limit readback sensor for SolaX X3 inverters

TCWORLD:
This fixes a mistake made in #1899.

For X3, the scale factor should be the same on the Sensor entity and the Number entity, but they ended up having opposing values.

This meant the export limit was being set correctly on the inverter, but when the value is read back it appears as 1/100 of the actual value.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -4410,7 +4410,7 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
         key="export_control_user_limit",
         register=0xB6,
         allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6 | X3,
-        read_scale=0.1,
+        scale=10,
         internal=True,
     ),
     SolaXModbusSensorEntityDescription(


TCWORLD:
@wills106 et al. apologies for messing up in the previous PR!

johny-mnemonic:
When do you plan to deliver this? 
Is it worth downgrading or should I wait for the fixed version?

TCWORLD:
@johny-mnemonic I'd either downgrade or manually apply the fix - i.e. modify `custom_components/solax_modbus/plugin_solax.py` to replace the one line as shown in the files tab of this PR. Again apologies for the blunder.

johny-mnemonic:
@TCWORLD no probs, fixed 👍

jprates:
For what is worth, I stumbled upon this bug yesterday and had to set the scale on integration to 1 instead of 0.1 or 10 in order for the export_control_user_limit to stick to intended value.

It's really annoying to have to keep track and try to understand why these values have to change from time to time, if anyone knows, please share that knowledge.

I'm on Solax X3 G4 with DSP 1.59 ARM 1.57.

Cheers.

PR #1922, opened 2026-03-28
Growatt: Add missing vpp register for allowing battery charging

nholmgaard:
This PR adds a missing register to the VPP (external override) section of the Modbus map. This register is needed for forcing the inverter to charge the battery from the grid. I have tested it on my MIN3600TL-XH hybrid converter. 

I am unsure exactly which flags should enable the register, it is only relevant on HYBRID. See comments in source.

Issues: My inverter sometimes limit the export power from the battery to 500W, and needs to have this set up again manually. I do not think that it is related to this register, which when enabled allows full power both in and out of the battery.


nholmgaard:
Sorry, got a bit ahead of myself there - I meant to create the PR to my own repo until it is fully tested. Please disregard for now.

brad5505:
Hi, I did not realize that  "vpp_allow_ac_charging"  register was also required to charge from the grid using VPP.  Thanks for picking this up. I look forward to seeing this PR added to the code once you have finished your testing.

nholmgaard:
Thanks for the excellent work - it made it easy to build more on top.

I have tested the addition of the register today, and it works - allowing
both export and import from the battery. The inverter seems to have a few
more settings which are reset at random times and which limit the power, so
I an not completely happy yet. I will try to track them down tomorrow.

I have a number of possible PRs as a result. The simplest is just adding
the register, which I will push shortly. I have also added the inverter as
well as my RS485 converter to the list of supported hardware. I have for my
own sake added the pdf documentation for the VPP registers, but I don't
know if that is in the spirit of the repo, or if it is better to leave it
out?

On top of that, I have had Claude Code build a scheduler for a specific
charge to buy and sell. I plan to expand this into an automation to plan
ahead based on the weather forecast and tarifs and charge at the cheapest
times if solar is insufficient. Would that be an interesting PR for this
repo, or is it out of scope?

Best regards,

        /Niels

custom_components/solax_modbus/plugin_growatt.py
```diff
@@ -1126,11 +1135,26 @@ def value_function_inverter_module(initval: int, descr: Any, datadict: dict[str,
...
+    GrowattModbusSelectEntityDescription(
+        name="VPP Allow AC charging",
+        key="vpp_allow_ac_charging",
+        register=30410,
+        register_data_type=REGISTER_U8L,
+        option_dict={
+            0: "Disabled",
+            1: "Enabled",
+        },
+        allowedtypes=GEN3 | GEN4 | HYBRID,
```
nholmgaard: Perhaps only HYBRID?

brad5505:
I added the "vpp_allow_ac_charging" register (30410) to my plugin_growatt.py file and tested drawing power from the grid to charge my battery on a poor solar day when my tariff was low. The battery charged at the rate set by  "VPP Power" register (30409) drawing power from both the grid and solar. So it all seems to work well.
This is the code I added.

   GrowattModbusSelectEntityDescription(
        name="VPP Allow AC charging",
        key="vpp_allow_ac_charging",
        register=30410,
        option_dict={
            0: "Disabled",
            1: "Enabled",
        },
        allowedtypes=GEN3 | GEN4,
        entity_category=EntityCategory.CONFIG,
        entity_registry_enabled_default=True,
        icon="mdi:dip-switch",
    ),
GrowattModbusSensorEntityDescription(
        key="vpp_allow_ac_charging",
        register=30410,
        scale={
            0: "Disabled",
            1: "Enabled",
        },
        allowedtypes=GEN3 | GEN4,
        internal=True,
    ),
    

> The inverter seems to have a few more settings which are reset at random times and which limit the power, so I an not completely happy yet. I will try to track them down tomorrow.

I have not noticed any settings reset or power being limited, but I may not be monitoring/ have enabled the settings that are resetting. Looking forward to see what you can track down.

> I have a number of possible PRs as a result. The simplest is just adding the register, which I will push shortly

I agree, a simple PR with just the additional register (30410) would be a good start, as people could just enable that register as part of their existing automations.



> I have for my own sake added the pdf documentation for the VPP registers, but I don't know if that is in the spirit of the repo, or if it is better to leave it out? On top of that, I have had Claude Code build a scheduler for a specific charge to buy and sell. I plan to expand this into an automation to plan ahead based on the weather forecast and tarifs and charge at the cheapest times if solar is insufficient. Would that be an interesting PR for this repo, or is it out of scope?

> Perhaps only HYBRID?

Maybe @wills106 can respond to these questions.

custom_components/solax_modbus/plugin_growatt.py
```diff
@@ -1126,6 +1128,7 @@ def value_function_inverter_module(initval: int, descr: Any, datadict: dict[str,
         name="VPP Remote Control",
         key="vpp_remote_control",
         register=30407,
+        register_data_type=REGISTER_U8L,
```
nholmgaard: Datasheet specifies this as an 8 bit register

datasheets/growatt/2.1.GROWATT.VPP.COMMUNICATION.PROTOCOL.OF.INVERTER_V2.01.pdf
nholmgaard, line 1: This may be a bad place for this file, but it does document all one needs to do in order to override the inverter usage via the vpp registers. I am happy to remove it from the PR if it violates normal practises.

custom_components/solax_modbus/plugin_growatt.py
nholmgaard: No harm from writing 16 bits, really a nit.

datasheets/growatt/2.1.GROWATT.VPP.COMMUNICATION.PROTOCOL.OF.INVERTER_V2.01.pdf
wills106: I think it's best to leave datasheets out of the code base.

Someone may have already shared the doc in a chat / comment somewhere on here.

custom_components/solax_modbus/plugin_growatt.py
wills106: Probably best leaving it as is, if it works. In case they have reserved the other 8bit for something.
Although it doesn't actually mention if they are using Hi8 or Low8

datasheets/growatt/2.1.GROWATT.VPP.COMMUNICATION.PROTOCOL.OF.INVERTER_V2.01.pdf
nholmgaard: Removed.

custom_components/solax_modbus/plugin_growatt.py
nholmgaard: Reverted. Since the write is now 16 bits, and since we should only be writing 0 or 1, it should be safe to use U8L, but it is unrelated to the main change and does no harm.

custom_components/solax_modbus/plugin_growatt.py
nholmgaard: @wills106 : I am still unsure here. The flag does not make sense if you do not have a battery (so Hybrid), but I am not entirely sure what generation would want this, and if GEN3 is correct. 

I know that it works for my MIN3600TL-XH, which should be a GEN4 | HYBRID, as far as I can see, but I assumed that since the rest of the VPP flags are GEN3, too, it ought to work for those as well.

brad5505: The registers do work with my Gen3 inverter with battery,
without
register_data_type=REGISTER_U8L,
and without
 | HYBRID

wills106: I think it's best to leave it HYBRID, as PV only shouldn't have this register and we haven't come across any AC Couple Battery only inverters for Growatt.

wills106:
Are you able to have a look at the failed CI/CD
Not sure if you are on older code compared to this repository?

nholmgaard:
Thanks, I will look at it tonight. I think that the change is pretty clean
now, so it should hopefully be easy to fix.

         /Niels

custom_components/solax_modbus/plugin_growatt.py
nholmgaard: Since brad5505 has confirmed that GEN3 also has this register working, GEN3 | GEN4 | HYBRID is correct.

nholmgaard:
> Are you able to have a look at the failed CI/CD Not sure if you are on older code compared to this repository?

I had a ruff change to the position of the U8L definition, which I have fixed and pushed, but I cannot restart a new run of the CI flow on this repo. When running pre-commit locally, mypy fails due to a type parameter default error in config_entries.py:374, which this commit does not touch. I am not sure what to do about that.

wills106:
I manually re-ran the test and it's passed thanks.

nholmgaard:
Thank you for the hand holding - hopefully I will be a bit less clumsy next time :-)

PR #1929, opened 2026-03-31
Revert PR #1831 to fix Growatt BMS1 Module 1 registers

TonyMathiesen:
Revert PR #1831 to fix Growatt BMS1 Module 1 registers


LiquidSnakeDM:
IMG_5009

Growatt MOD8000TL3-XH still not correct.
Can you also fix that?

PR #1931, opened 2026-04-02
Time v5

wills106:


wills106:
@rosenrot00 Just wondering if you could check the autorepeat functions still work after this change.
As I had to change a couple of `time()` to `_mtime.time()`
Or don't you use them?

Thanks

rosenrot00:
@wills106 unfortunately I have never touched the autorepeat functions. Thus, I might not the biggest help with it.

If you would do a prerelease, I could try to change times and see if they change accordingly in the SolaxApp, but I guess that is what you did already.

wills106:
I pushed a Beta out with these changes in [2026.04.2b1](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2026.04.2b1)

I have an older Gen3 so these VPP Modes that rely on the autorepeat don't exist on my Inverter.

rosenrot00:
@wills106 I just tested it on my GEN4 and it works in both directions between the SolaxApp and the integration.

Just as this change has some impact anyway, I would re-think the naming. As Homeassistant does the positioning of the elements in an alphabetical order, it might be the right time to change naming so that functional time blocks are grouped together visually. I actually never used it but it isn't very appealing to the eye currently :)

PR #1961, opened 2026-04-14
Patch 4

flopp999:
added 3 models
renamed 2 sensors


wills106:
Are you able to look into the CI/CD failing?
I think updating your local branch and resubmitting should fix it.

PR #1964, opened 2026-04-14
Rename inverter state and update inverter types

flopp999:


flopp999:
@wills106 I rename my old patch-1 to patch-11, so this patch-1 is a new patch. I used main after you change da.json
But it fails again

wills106:
It's no longer failing on da.json
I'll put the error through Roo Code and try and understand what it's not happy about.

Sorry about this.

flopp999:
> It's no longer failing on da.json I'll put the error through Roo Code and try and understand what it's not happy about.
> 
> Sorry about this.

no problem at all. I am also trying to understand what is wrong, but don't understand :)

wills106:
I think it's failing because there are too many indents on Line 9588
Might be wrong though...

PR #1972, opened 2026-04-17
Fix Export-First control with alternative house load

rosenrot00:
Adjusts the Export-First Battery Limit logic to use the alternative house load basis without double-counting current battery charging. house_load_alt includes battery charging as load, so the current charge power is subtracted before calculating available export headroom.

This also fixes a case where using the regular house load caused the inverter to keep importing from the grid when PV power was below house load, even though enough battery power was available to cover the deficit.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -711,8 +711,12 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
         # 3) Ensure that we don't exceed the inverter power limit in the calculations so that PV above
         #    this limit is allocated to the battery charge rather than being limited.
 
-        # Use the main house load, but clamp obviously invalid negative values.
-        hl = max(0, int(houseload))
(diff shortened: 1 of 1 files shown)


TCWORLD:
I don't understand this change. Maybe there is a difference between inverter versions. 

For my X1-IES (gen 5), the house load alt calculation doesn't include battery power charge - so subtracting them for the control loop would surely either double the house load or zero it out:

Screenshot_20260421-094906_ASUS_Launcher (1)

Unless I'm missing something.

rosenrot00:
> I don't understand this change. Maybe there is a difference between inverter versions.
> 
> For my X1-IES (gen 5), the house load alt calculation doesn't include battery power charge - so subtracting them for the control loop would surely either double the house load or zero it out:
> 
> Screenshot_20260421-094906_ASUS_Launcher (1)
> Unless I'm missing something.

I’ve reviewed it and realized I overlooked that. My current assumption is that subtracting it a second time may be masking another issue, likely related to the inverter limiting PV output as it approaches the grid export threshold. I’ll adjust the formula and monitor how it behaves.

PR #1976, opened 2026-04-19
Export-First Mode: Do not cap Export-First target by inverter power

rosenrot00:
Adjusts the Export-First Battery Limit control so the export target is based on the configured export limit instead of being reduced by inverter_power - house_load.

The previous logic could lower the effective export target too much when house load was high. This caused the controller to start charging the battery even though the inverter should still prioritize grid export.

This keeps the existing house_load_alt correction that removes current battery charging from the load basis, so battery charging is not double-counted when calculating the control target.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -720,14 +720,10 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
 
         # Export limit no readscale:
         export_limit = datadict.get("export_control_user_limit", 30000)
-        inverter_limit = datadict.get("inverter_power_type", 30000)
 
-        # Both house load and exported power come from the inverter and must fit within the inverter
-        # power limit. For example if we have a 6kW inverter and 4kW of house load, then we cannot
-        # export more than 2kW even if our export limit is higher. Trim the export limit so that any
-        # excess can be used for charging the battery rather than PV being clamped.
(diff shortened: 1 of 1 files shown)


TCWORLD:
The check was intended to allow the PV to charge the battery with any excess in this situation rather than reducing the PV output.

I'm curious as to why it didn't work. If the house load is high, then there is limited ability to export because the inverter output is limited.

rosenrot00:
@TCWORLD what I found is that my inverter is labeled 8kW but it does produce up to 10kW. I think that this is causing issues as the inverters can be set and are set default to do more than 100%.

It isn't easy to find this not working, I only found it because my inverter is 8kW and the maximum houseload with the headpump on is at about 8.5kW.

With this change it works great on my end. I thought as only a few people are using this mode, I could simply modify and see if it becomes better. Other people would report if anyone uses this mode.

TCWORLD:
I guess it's probably more relevant to inverters with low power rating where it's much easier to exceed the limit. Perhaps make it to a configurable option like the hysteresis setting, happy with it being disabled by default. That way of people have an issue down the line we can get them to enable the limiting to see if it helps.

rosenrot00:
100% with you. I also switched to houseload_alt as it behalves much better on my setup. At some point it makes sense to make these as settings.

TCWORLD:
Would definitely make sense to move the various factors/deadband values/etc, to a more formal configuration method at some point for sure (perhaps something like a JSON config file for advanced settings).

PR #1980, opened 2026-04-19
Add support for Growatt SPF5000ES firmware branch 113

SimoneB79:
This PR extends the inverter auto-detection logic to include Growatt SPF5000ES devices running firmware branch 113.

Previously, only some SPF firmware identifiers (e.g. 067) were recognized, which required manual changes after each update for inverters using firmware 113.

This change adds the missing identifiers to the existing matching logic, allowing automatic detection without modifying the core behavior.

Tested on:
- Growatt SPF5000ES (firmware branch 113)

No regressions observed with existing firmware mappings.

### DIFF
--- a/custom_components/solax_modbus/plugin_growatt.py
+++ b/custom_components/solax_modbus/plugin_growatt.py
@@ -9763,14 +9763,9 @@ class growatt_plugin(plugin_base):
                 invertertype = HYBRID | GEN4 | X3  # Hybrid TL3-XH (BP) 3kW - 10kW (MOD), 11kW - 30kW (MID)
             elif seriesnumber.startswith("V"):
                 invertertype = HYBRID | GEN4 | X3  # Hybrid TL3-XH 3kW - 10kW (MOD)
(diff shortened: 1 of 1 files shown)


flopp999:
@SimoneB79 it cannot detect your model with 3 first characters?

What is the 3 first character in your serial number?

SimoneB79:
Hi, I have two SPF 5000 ES inverters with firmware 113.114. I know there are other firmware branches, trac ui 069, and that's tracked in the plugin. You can check it by powering on the inverter or with ShinePhone.
If other firmware versions of the same model emerge, we could integrate them into the recognition.
Screenshot 2026-04-30 070558

flopp999:
Yes, but what is your serial number?
You can see it in the label kör nu your inverter

SimoneB79:
WhatsApp Image 2026-04-30 at 07 21 07

flopp999:
great then it much better to add your model with serial detection, then it doesn't care about your firmware version

PR #1993, opened 2026-04-26
Further improvements to Mode8 Negative Injection

TCWORLD:
Further development of the Negative Injection mode to address shortcomings in the previous changes.

1. Allow limiting max SoC in negative injection mode using the `Remotecontrol Target SoC` entity. Once this is reached, PV will be limited instead of further charging.
2. Improves the control loop further to target increasing battery charge power if possible before limiting PV. This should maximise PV output when the battery is not yet at target capacity.
3. Shift the default injection bias to ~50W instead of 20W import to reduce oscillation between import and export.
4. Speed up response to excess export when the previously set PV limit is well above the current PV output.


infradom:
thanks, I will test this. 

Today, in Belgium, we have very negative injection as wel as very negative consumption prices.
So my automations will move from "negative price" to "negative injection and consumption price". But this second step only makes sense if the battery is nearly empty. So I consider introducing another negative price mode that does not charge the battery, so that the battery remains empty until the "negative consumption price"  is triggered.

TCWORLD:
Rather than a distinct mode, maybe we could use the `Target SoC (Mode 9)` parameter in the Negative Injection to set the upper limit for charging the battery? That way it will naturally go into the PV limiting once that target has reached allowing the battery to stay part filled for such scenarios.

I've pushed an example of that in https://github.com/TCWORLD/homeassistant-solax-modbus/commit/a4a12ff1eaa7c36dec76a0af9d2044af38ff1ea5, can merge into this branch if it seems like a good idea.

infradom:
> Rather than a distinct mode, maybe we could use the `Target SoC (Mode 9)` parameter in the Negative Injection to set the upper limit for charging the battery? That way it will naturally go into the PV limiting once that target has reached allowing the battery to stay part filled for such scenarios.

yes, Indeed, I never understood the purpose of the mode9 target SOC, but I believe you are right. Thanks !

infradom:
> I've pushed an example of that in [TCWORLD@a4a12ff](https://github.com/TCWORLD/homeassistant-solax-modbus/commit/a4a12ff1eaa7c36dec76a0af9d2044af38ff1ea5), can merge into this branch if it seems like a good idea.

Great,I will certainly test, but as today has extremely negative prices (around 0.5 €/kWh) this afternoon, I will not yet test this today, as I am searching the correct approach manually first.
I believe we will see such negative prices more and more (even though the press talks about an energy crisis ;-) )

infradom:
I had a quick look at your code today, and I do not see where you specify mode9 should be used instead of mode 8. In my opinion, the return value should be modified and include somthing that maps to mode 9 in the mode 9 target SOC

```
  res = [

        (
            "remotecontrol_power_control_mode",
            "Mode 8 - PV and BAT control - Duration", # should map to 9 in case we want target SOC
        ),
     ...
```
either we use the constant 9, or we extend 
```
        name="Remotecontrol Power Control Mode (mode 8/9)",
        key="remotecontrol_power_control_mode",
        register_data_type=REGISTER_U16,
        write_method=WRITE_DATA_LOCAL,
        option_dict={
            0: "Disabled",  # not in older documentation, in recent docs added
            8: "Mode 8 - PV and BAT control - Duration",
            81: "Negative Injection Price",
            82: "Negative Injection and Consumption Price",
            83: "Export-First Battery Limit",
            84: "Enabled Grid Control",
            85: "Enabled No Discharge",
            86: "Enabled Feedin Priority",
            9:  "Mode 9 - PV and BAT control - Target SOC", # UNCOMMENT
        },

```

Maybe I looked at this too quickly ...

TCWORLD:
@infradom Apologies for the confusion - I wasn't suggesting using mode 9 specifically, just the config entity called `remotecontrol_target_soc_8_9` (previously named `Target SoC (Mode 9)`, now named `Target SoC (Mode 8/9)`).

In the future it may be worth looking at changing work modes for certain operations in a number of cases, but that's a bigger change .

pekacpekac59-design:
@TCWORLD 
Not sure if you noticed I uploaded some detailed charts yesterday. 

[closed issue #1985](https://github.com/wills106/homeassistant-solax-modbus/issues/1945#issuecomment-4322708175)

infradom:
> @infradom Apologies for the confusion - I wasn't suggesting using mode 9 specifically, just the config entity called `remotecontrol_target_soc_8_9` (previously named `Target SoC (Mode 9)`, now named `Target SoC (Mode 8/9)`).
> 
> In the future it may be worth looking at changing work modes for certain operations in a number of cases, but that's a bigger change .

Thanks, I understand, so if we dont use mode 9, target SoC will be ignored in mode 8. You probably suggest I use this parameter in my automations ?

I will have a look at your charts !  Thanks again.
These days, I have been busy with other task, hope to become more active here soon.

TCWORLD:
@infradom Yes. Previously the target soc parameter was unused in mode 8. Now it is used to set the maximum charge SoC for the Neg-Inj mode.

TCWORLD:
I'm going to mark this ready as the graphs in the linked issue show reasonable behaviour, and its in a better state than the version on main.

Some more optimisation work going forward would be useful.

infradom:
@TCWORLD, thanks, I understand now, did not study the code well enough to understand what you meant. Installed most recent version now ...

PR #2000, opened 2026-04-28
Fix firmware version display for X3-MIC GEN2, Hybrid GEN4/5/6 and EVC GEN1

Bl00d-B0b:
## What changed

**plugin_solax.py — Hybrid / MIC inverters:**

- **Fix X3-MIC GEN2 firmware display** — was showing `DSP ?.136 ARM ?.136` because `value_function_software_version_g4` was applied to all MIC X3 devices, but that function reads `firmware_dsp_major` (FC03 0x7F) which only exists on AC/HYBRID. Split the MIC X3 `software_version` sensor into two entries: GEN1 uses `value_function_software_version_air_g4` (plain `DSP 136 ARM 136`), GEN2 uses `value_function_software_version_g2` (formatted `DSP v2.136 ARM v2.136`). `MU802T` blacklist moved to the GEN1 entry where it belongs.
- **Add FC03 0x7B / 0x7C** (`firmware_version_dsp` / `firmware_version_arm`) as internal sensors for `AC | HYBRID | GEN4 | GEN5 | GEN6` — these are the PDF-documented full-version registers (`FirmwareVersionDSP = DSPMajor×100+DSPMinor`). New `value_function_software_version_full` decodes them cleanly (e.g. raw 157 → `1.57`). GEN4 and GEN5/GEN6 `software_version` now use this function instead of the 4-register split approach. Degrades gracefully to `DSP ?.?? ARM ?.??` if a register read fails.
- **Remove FC03 0x81** (`firmwareversion_modbustcp_major`) — not present in Modbus RTU V1.02 documentation; confirmed always returns 0 on live hardware (X3-Hybrid G4 PRO).
- **Rename FC03 0x82** from `firmwareversion_modbustcp_minor` → `modbus_protocol_version` — this register holds the Modbus protocol document version (e.g. 100 = V001.00), not a TCP firmware version. Extended `allowedtypes` to include GEN4/GEN5/GEN6.
- **Remove dead code** — `self.software_version = "Unknown"` in the `MC208T` branch of `async_determineInverterType` (field is never read).

**plugin_solax_ev_charger.py — EV Charger:**

- **Fix EVC GEN1 firmware sensor** — FC04 0x25 (`firmware_version`) was restricted to `allowedtypes=GEN2` only. GEN1 devices (C311/C322 series with firmware < 7.0) respond correctly to this register. Confirmed live on X3-EVC-11kW GEN1 returning raw 117 → `1.17`.

## Why

- The MIC GEN2 display bug affects all X3-MIC GEN2 users — the `?` major version was caused by a missing register, not a firmware issue.
- Registers 0x7B/0x7C are the primary firmware version registers documented in the official SolaX Modbus RTU V1.02 spec; the split approach reading four registers separately was working but unnecessarily complex for GEN4+.
- Register 0x81 was never in any official documentation and is always 0 — it caused confusion with the misnamed 0x82.
- EVC GEN1 users had no firmware version visible in HA despite the register being fully readable.

## Tested on live hardware

| Device | Serial prefix | Result |
|---|---|---|
| X3-MIC (GEN2) | MC208T | `DSP v2.136 ARM v2.136` ✓ |
| X3-Hybrid G4 PRO 15kW (GEN4) | H34A15 | `DSP 1.57 ARM 1.56` ✓ |
| X3-EVC-11kW (GEN1) | C31103 | `sensor.ev_firmware_version: 1.17` ✓ |

**Reference:** SolaX Hybrid X1&X3 ESS Modbus RTU V1.02 (Nov 2025) — registers 0x7B, 0x7C, 0x7D, 0x7F, 0x80, 0x82, 0x83

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1240,6 +1240,14 @@ def value_function_software_version_g5(initval: int, descr: Any, datadict: dict[
     )
 
 
+def value_function_software_version_full(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
+    dsp = datadict.get('firmware_version_dsp')
+    arm = datadict.get('firmware_version_arm')
+    if dsp is None or arm is None:
+        return None
+    return f"DSP {dsp // 100}.{dsp % 100:02d} ARM {arm // 100}.{arm % 100:02d}"
+
+
 def value_function_software_version_air_g3(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
     return f"DSP v2.{datadict.get('firmware_dsp')} ARM v1.{datadict.get('firmware_arm')}"
 
@@ -3878,6 +3886,18 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
         icon="mdi:dip-switch",
     ),
     SolaXModbusSensorEntityDescription(
+        key="firmware_version_dsp",
+        register=0x7B,
+        allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6,
+        internal=True,
+    ),
+    SolaXModbusSensorEntityDescription(
+        key="firmware_version_arm",
+        register=0x7C,
+        allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6,
+        internal=True,
+    ),
+    SolaXModbusSensorEntityDescription(
         key="firmware_dsp",
         register=0x7D,
         allowedtypes=AC | HYBRID,
@@ -3905,19 +3925,10 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
         internal=True,
     ),
     SolaXModbusSensorEntityDescription(
-        name="Firmware Version Modbus TCP Major",
-        key="firmwareversion_modbustcp_major",
+        name="Modbus Protocol Version",
+        key="modbus_protocol_version",
         entity_registry_enabled_default=False,
-        register=0x81,
-        allowedtypes=AC | HYBRID | GEN5,
-        entity_category=EntityCategory.DIAGNOSTIC,
-        icon="mdi:information",
-    ),
-    SolaXModbusSensorEntityDescription(
-        name="Firmware Version Modbus TCP Minor",
-        key="firmwareversion_modbustcp_minor",
-        entity_registry_enabled_default=False,
-        allowedtypes=AC | HYBRID | GEN5,
+        allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6,
         register=0x82,
         entity_category=EntityCategory.DIAGNOSTIC,
         icon="mdi:information",
@@ -8243,13 +8254,13 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
     ),
     SolaXModbusSensorEntityDescription(
         key="software_version",
-        value_function=value_function_software_version_g4,
+        value_function=value_function_software_version_full,
         allowedtypes=AC | HYBRID | GEN4,
         internal=True,
     ),
     SolaXModbusSensorEntityDescription(
         key="software_version",
-        value_function=value_function_software_version_g5,
+        value_function=value_function_software_version_full,
         allowedtypes=AC | HYBRID | GEN5 | GEN6,
         internal=True,
     ),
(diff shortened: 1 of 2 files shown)


custom_components/solax_modbus/plugin_solax.py
```diff
@@ -8243,13 +8254,13 @@ def value_function_battery_voltage_cell_difference(initval: int, descr: Any, dat
...
         allowedtypes=AC | HYBRID | GEN4,
         internal=True,
     ),
     SolaXModbusSensorEntityDescription(
         key="software_version",
-        value_function=value_function_software_version_g5,
+        value_function=value_function_software_version_full,
         allowedtypes=AC | HYBRID | GEN5 | GEN6,
         internal=True,
     ),
```
TCWORLD, line 8260: As these now use the same value function, the two definitions can be merged to a single `AC | HYBRID | GEN4 | GEN5 | GEN6` entity.

TCWORLD (changes requested):

Bl00d-B0b:
Good catch, thank you! Merged the two identical entries into a single `AC | HYBRID | GEN4 | GEN5 | GEN6` definition — pushed in commit 8a618f9.

wills106:
Are you able to look at correcting the failed CI/CD for me please?

Roo Code output:

> Why the Exit Code is 1
> 
> The key issue is in the CI output:
> 
> This happens because pre-commit returns exit code 1 when any hook modifies files. The ruff-format hook:
> Detected that the file used single quotes ('firmware_version_dsp' and 'firmware_version_arm')
> Reformatted them to double quotes ("firmware_version_dsp" and "firmware_version_arm") to match the project's quote-style = "double" setting
> Wrote the changes back to the file
> Returned exit code 1 to signal "files were modified"
> The CI pipeline interprets this non-zero exit code as a failure.
> The Fix
> 
> The PR author needs to:
> Commit the reformatted changes that the hook made (shown in the diff)
> Push the updated commit to the PR branch
> The diff shows exactly what the hook changed:
> Line 1243: 'firmware_version_dsp' → "firmware_version_dsp"
> Line 1244: 'firmware_version_arm' → "firmware_version_arm"
> After committing these changes, the hooks will run again but this time won't modify any files (since they're already correctly formatted), and the exit code will be 0 (success).

wills106:
I’ll correct the changes locally, as I want to merge these changes so I can do a release.

TCWORLD:
Unfortunately while `0x7B` and `0x7C` exist in the latest protocol docs, they don't work for lots of older inverters (X1-IES, X3-G4-Pro, etc), so this change basically results in lots of older inverters reporting back a version of 0.00. The fail back to ?.?? doesn't happen because they do return a value, but one of 0.

I'd suggest we revert the GEN4/5/6 version number change as while slightly more complex, at least it appears to work for all inverters.

Or have a hybrid approach using the single register, but failing back to the split registers if a value of 0. However unless there is an inverter where the old split registers don't work, that seems like extra unnecessary complexity.

Bl00d-B0b:
I notice that. Here's the full summary of my devices:

**X3-MIC/PRO-G2 — simple combined encoding**

| Addr | Raw | Decoded | Key | Cloud |
|---|---|---|---|---|
| `0x352` | 136 | `1.36` | `firmware_version_dsp` | MDSP 1.36 ✓ |
| `0x353` | 136 | `1.36` | `firmware_version_arm` | ARM 1.36 ✓ |
| `0x354` | 100 | `1.00` | `firmware_version_arm_sub` | -1.00 ✓ |

**X3-Hybrid-G4 — split major/minor encoding**

| Addr | Raw | Notes | Cloud |
|---|---|---|---|
| `0x7B` | 160 | `firmware_version_dsp` → `1.60` ✓ | MDSP 1.60 ✓ |
| `0x7C` | 158 | `firmware_version_arm` → `1.58` ✓ | ARM 1.58 ✓ |
| `0x7D` | 60 | `firmware_dsp` (minor only) | |
| `0x7F` | 1 | `firmware_dsp_major` | |
| `0x80` | 1 | `firmware_arm_major` | |
| `0x83` | 58 | ARM minor (unassigned) — matches 0x7C minor | |
| `0x84` | 15 | **ARM sub-minor** — with major from `0x80`=1 → `1.15` | -1.15 ✓ |

maybe you can provide your registers? then i can update a code

Bl00d-B0b:
please be sure, if you revert back, it also remove a fix for EV Charged, which is correct. it's better to have a new code.

TCWORLD:
@Bl00d-B0b I wasn't suggesting reverting the whole PR at all, just specifically the version number parsing for Gen4/5/6.

PR #2001, opened 2026-04-28
fix(evc): correct control_command idle state and rename ECO/Green mode selects

Bl00d-B0b:
## What changed

**plugin_solax_ev_charger.py — EV Charger:**

- **Fix \`start_charge_mode\` always showing unknown** — register 0x610 was only declared in the select entity. The hub builds Modbus scan blocks exclusively from sensor descriptions, so without a backing sensor the register was never polled and \`hub.data\` was never populated. Added an \`internal=True\` sensor (no HA entity created) to guarantee the register is scanned every cycle.
- **Fix \`control_command\` showing unknown in normal operation** — register 0x627 returns 0 when no command is active. Added \`0: "No Command"\` to both the sensor scale dict and the select option dict. These must stay in sync: once a sensor scale dict converts a raw integer to a string, that string must exist in the matching select's option dict or the select stays unknown.
- **Add Overload Limit / Undervoltage Limit** (FC03 0x611 / 0x612) — confirmed from live register scan (0x611=265, 0x612=160). Added as number entities (configurable, disabled by default) with matching diagnostic sensors. Identified as unknown registers in earlier community work.
- **Add Modbus Address / Baud Rate** (FC03 0x640 / 0x641) — read-only diagnostic sensors. 0x640=70 (slave address), 0x641=2 (baud rate code, 2=9600). Matches the known EVC communication settings.
- **Rename "ECO Gear" → "ECO Mode" / "Green Gear" → "Green Mode"** — matches SolaX app terminology. Entity keys unchanged.
- **Rename "Meter Setting" → "Grid Data Source"** — matches SolaX app terminology. Entity key unchanged.

## Why

- \`start_charge_mode\` was the only select in the EVC plugin with no backing sensor — all others have a matching sensor that gets the register into the poll blocks.
- Overload/Undervoltage limits and Modbus config registers were present in the hardware but not mapped. Identified via direct FC03 scan of a live X3-EVC-11kW GEN1 device.
- "ECO Gear", "Green Gear", and "Meter Setting" are internal register names; the SolaX app uses "ECO Mode", "Green Mode", and "Grid Data Source".

## Tested on live hardware

| Device | Serial prefix | Result |
|---|---|---|
| X3-EVC-11kW (GEN1) | C31103 | \`start_charge_mode\`: Plug & Charge ✓, \`control_command\`: No Command ✓, Overload Limit: 265V ✓, Undervoltage Limit: 160V ✓, Modbus Address: 70 ✓, Baud Rate: 9600 ✓ |


Bl00d-B0b:
Closing to reopen as a cleaner, updated PR from scratch.

PR #2019, opened 2026-05-02
Add Damping to Mode8 Negative Injection Control Loop

TCWORLD:
Further improvements to improve stability in the Mode 8 Negative Injection mode. This aims to remove oscillation behaviour seen due to there being no damping on the new set-points values being calculated.

Based on discussions in https://github.com/wills106/homeassistant-solax-modbus/discussions/2017

Example execution:

image

Close-up of grid port - largely hovering around the target ~50W import bias. Some short export periods caused by damped response, but that is to be expected.

image

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -690,10 +690,13 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
 
             # Battery gets surplus up to BMS limit
             desired_charge, bms_cap_w, pct_cap_w = autorepeat_bms_charge(datadict, battery_capacity, max_charge_soc, surplus)
-            pushmode_power = -desired_charge
+
+            # Simple moving average filter to slow down changes to battery
+            selected_charge = (cur_push * 4 + desired_charge) / 5
+            pushmode_power = -selected_charge
 
             # Any surplus not able to feed into the battery needs to be corrected through PV limiting
-            error = surplus - desired_charge
+            error = surplus - selected_charge
 
             if abs(error) <= export_deadband_w:
                 target_pvlimit = cur_pvlimit
@@ -713,12 +716,13 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
                 control_reason = "increase-pv"
 
             # Simple moving average filter to slow down changes to PV limit
-            pvlimit = (cur_pvlimit * 7 + target_pvlimit) / 8
+            pvlimit = (cur_pvlimit * 4 + target_pvlimit) / 5
 
             _LOGGER.debug(
                 f"[Mode8 Negative Injection] {control_state}: surplus={surplus}W measured_power={measured_power}W "
                 f"export_target={export_target}W error={error}W pvtarget={target_pvlimit}W reason={control_reason} "
-                f"bms_cap≈{bms_cap_w}W pct_cap={pct_cap_w}W -> charge={desired_charge}W pvlimit={pvlimit}W hl={hl}W"
+                f"bms_cap≈{bms_cap_w}W pct_cap={pct_cap_w}W desired_charge={desired_charge}W -> charge={selected_charge}W "
+                f"pvlimit={pvlimit}W hl={hl}W"
             )
 
         else:
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -553,7 +553,7 @@ def autorepeat_function_remotecontrol_recompute(initval: int, descr: Any, datadi
     return {"action": WRITE_MULTI_MODBUS, "data": res}
 
 
-def autorepeat_bms_charge(datadict: dict[str, Any], battery_capacity: float, max_charge_soc: float, available: float) -> tuple[int, int, int]:
+def autorepeat_bms_charge(datadict: dict[str, Any], battery_capacity: float, max_charge_soc: float, available: float) -> tuple[int, int, int, int]:
     # Determines max rate for charging battery
 
     # User cap (% of BMS max charge power).
@@ -590,12 +590,14 @@ def autorepeat_bms_charge(datadict: dict[str, Any], battery_capacity: float, max
     if battery_capacity < max_charge_soc:
         # Limit to charge rate to lesser of the available
         # power and the %age capped charge limit.
+        max_charge = int(max(0, pct_cap_w))
         desired_charge = int(max(0, min(available, pct_cap_w)))
     else:
         # Can't charge the battery
+        max_charge = 0
         desired_charge = 0
 
-    return desired_charge, bms_cap_w, pct_cap_w
+    return desired_charge, max_charge, bms_cap_w, pct_cap_w
 
 
 def autorepeat_setpoint_filter(current_value: float, desired_value: float, steps: int = 5) -> float:
@@ -695,7 +697,7 @@ def autorepeat_function_powercontrolmode8_recompute(initval: int, descr: Any, da
             control_state = "surplus" if pv >= hl else "clipping"
 
             # Battery gets surplus up to BMS limit
-            desired_charge, bms_cap_w, pct_cap_w = autorepeat_bms_charge(datadict, battery_capacity, max_charge_soc, surplus)
+            desired_charge, max_charge, bms_cap_w, pct_cap_w = autorepeat_bms_charge(datadict, battery_capacity, max_charge_soc, surplus)
 
             # Setpoint filter to slow down changes to battery
             selected_charge = autorepeat_setpoint_filter(current_charge, desired_charge)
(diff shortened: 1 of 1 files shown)


infradom:
One remark: I do not understand 
```
cur_push = max(0, battery_charge if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else (0 - cur_push))
```

I would expect cur_push instead of (0 - cur_push)
I would rather inverse battery_charge

should not this be:
```
cur_push = min(0, (0 - battery_charge) if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else cur_push)
```

infradom:
> One remark: I do not understand
> 
> ```
> cur_push = max(0, battery_charge if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else (0 - cur_push))
> ```
> 
> I would expect cur_push instead of (0 - cur_push) I would rather inverse battery_charge
> 
> should not this be:
> 
> ```
> cur_push = min(0, (0 - battery_charge) if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else cur_push)
> ```

Sorry, the name was misleading, cur_charge would  be a better name
Please ignore my remark for now
EDIT: But I remain confused about this

TCWORLD:
Yeah, the naming might be off. The logic is correct, but its the weirdness that battery charge and push power have opposite signs. `current_charge` would make more sense.

infradom:
Thanks, I am extending your model to the full negative section:
```
    elif power_control == "Negative Injection and Consumption Price":  # disable PV, charge from grid
        pvlimit = 0
        cur_push = max(0, (-battery_charge) if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else cur_push)
        current_charge = -cur_push
        desired_charge = import_limit - house_load
        # Simple moving average filter to slow down changes to battery
        selected_charge = (current_charge * 4 + desired_charge) / 5
        pushmode_power = -selected_charge
```

To be tested ...

TCWORLD:
I've added the filter to the deficit pathway of NI as well just for belt and braces.
Also moved the filter into its own function to keep things neat.

infradom:
This latest version seems to limit PV immediately, before the SOC target is reached. (NI mode)
My soc target is set to 98% today, and it starts limiting PV immediately

TCWORLD:
Interesting, could you upload debug logs

infradom:
> Interesting, could you upload debug logs
Currently I am analyzing what goes wrong in the logs, will send debug logs once I understand better which part causes the problem. (debug logs can be overwhelming, wo I want to send only an interesting part)

infradom:
Please find a log (I changed the original statements temporarily to info instead of debug, and added some extra info statements). The full debug log would have contained more irrelevant entries.
[home-assistant_2026-05-03T09-19-38.657Z.log](https://github.com/user-attachments/files/27315265/home-assistant_2026-05-03T09-19-38.657Z.log)

note: With this rainy weather here, during the logging, uncapped PV was somewhere between 1200 and 3500 . House load is around 350W.

TCWORLD:
Ahhh. Try it now.

I swapped the sign of `cur_push` when tidying it up, but forgot to change the `max(0, cur_push)` to `min(0, cur_push)`.

Edit: In fact, it shouldn't be clamping it at all. Removed the min entirely now.

infradom:
yes, I already tried with min(), but that did not help. Will retest soon with your latest version

TCWORLD:
The other thing would be to try changing `error = surplus - selected_charge` to `error = surplus - desired_charge` on line 705 and see if that helps. The two filters may end up with excessive limiting of the PV, so using desired rather than the filtered battery charge for calculating the error in PV may avoid that issue.

infradom:
The first change (remove clamping) did not help: see attached logfile
[home-assistant_2026-05-03T10-41-56.224Z.log](https://github.com/user-attachments/files/27316267/home-assistant_2026-05-03T10-41-56.224Z.log)

I will now test your latest suggestion:

> The other thing would be to try changing `error = surplus - selected_charge` to `error = surplus - desired_charge` on line 705 and see if that helps. The two filters may end up with excessive limiting of the PV, so using desired rather than the filtered battery charge for calculating the error in PV may avoid that issue.

infradom:
That latest suggestion did not help either (might even oscillate a bit more)
See log:
[home-assistant_2026-05-03T10-50-57.678Z.log](https://github.com/user-attachments/files/27316355/home-assistant_2026-05-03T10-50-57.678Z.log)

TCWORLD:
Yeah I was afraid of that. It's very tricky to figure out a nice way of controlling both limits without knowing what went on previously.

I think I also ended up moving us back to the doom spiral we saw when playing with export first - where PV limit decays, but there is nothing to allow it to increase.

I've tried to add an extra check in now that basically says - "once we're stable, see if the battery can absorb any more energy, and if so bump up the PV limit.". Whether that is going to work or cause more problems I don't know.

infradom:
Still, at 57%,  I am far from a battery full condition, so there must be something more fundamentally wrong. It is just the PV limit that is completely wrong; why is desired_charge so low when returned from autorepeat_bms_charge ? I would expect it should charge at full capacity if battery is only 57 % , but I must admit that I did not study autorepeat_bms_charge . It looks too complicated for my simple mind ;-)

EDIT: I will of course test your latest upload

TCWORLD:
The BMS charge function correctly works out that your battery can charge at 8kW based on current SoC so that's ok.

The reason the desired charge is so low is because PV has been limited to the point that it can't charge any faster otherwise it would start charging off the grid. It therefore needs to respond by trying to increase the PV output, but there wasn't anything doing that. I hope the new change will, but it may not be enough.

infradom:
Tested the latest changes, they are still limiting PV (to more or less the house load)

infradom:
Correction: a second test seems to work better ... dont know what is happening ...
Edit: Third test: still behaving as expected ..

TCWORLD:
If possible, let it run for 10-15 minutes and then upload a log. Will be interesting to know what it does over time.

infradom:
> If possible, let it run for 10-15 minutes and then upload a log. Will be interesting to know what it does over time

[home-assistant_2026-05-03T12-19-39.379Z.log](https://github.com/user-attachments/files/27317574/home-assistant_2026-05-03T12-19-39.379Z.log)

Still behaving well !

TCWORLD:
Interesting response. You can see it stabilising on the no-export target and then periodically kicking up the PV limit to try and boost the output.

image

TCWORLD:
Could potentially either increase PV filter steps or decrease battery filter steps so that the battery response is faster than the PV response. That might help with the slightly weird PV dips.

infradom:
In the past, I had the impression that the system does not react quickly to PV target changes, I have the feeling that it can adapt faster to battery charge changes.
But these are just impressions, without solid proof.

Thanks again for your work on this !

infradom:
> That might help with the slightly weird PV dips.

The weather here is very strange today, so PV fluctuates a lot, so the numbers are difficult to interprete.

When there is a real PV dip during a rain shower, the target PV goes up to the configured maximum, which looks fine to me, the system tries to reach the maximum sunlight when there is insufficient.

TCWORLD:
I'm looking around say 14:10 where it bumps up the PV, and then you see the battery charge (yellow) going up but the pv limit (grey) decaying slowly. If we make the battery slightly faster it may reduce the amount the PV limit drops, but probably not really an issue.

infradom:
UPDATED:

For the 'Full Negative Scenario' (Negative Injection and Consumption Price), I modified my development fork with:
```
    elif power_control == "Negative Injection and Consumption Price":  # disable PV, charge from grid
        pvlimit = 0
        battery_charge = max(0, int(datadict.get("battery_power_charge", 0) or 0))
        max_charge_soc = min(target_soc, datadict.get("battery_charge_upper_soc", 100))
        cur_push = (-battery_charge) if (cur_push := datadict.get("remotecontrol_current_pushmode_power", None)) is None else cur_push
        pushmode_power = 0  # + = discharge, - = charge
        current_charge = -cur_push
        available = import_limit - house_load # or ( current_charge + measured_power + import_limit  )
        desired_charge, max_charge, bms_cap_w, pct_cap_w = autorepeat_bms_charge(datadict, battery_capacity, max_charge_soc, available)
        selected_charge = autorepeat_setpoint_filter(current_charge, desired_charge)
        pushmode_power = -selected_charge
```

The idea of this model is to put pv input = 0, while loading from the grid ad the maximum specified import_limit.

NOT YET TESTED - waiting for a full-negative prices day to test (probably next weekend)

(Feel free to add this or an enhanced version to your PR, otherwise I will issue a separate PR after sufficient testing in a full negative price scenario.)

TCWORLD:
I've added the changes for the full negative case with a bit of tidying and a bug-fix (`house_load` variable doesn't exist). Changed it to use `houseload_alt` for consistency with the other updated loops.

TCWORLD:
I've tested the full-negative case (albeit its night-time so no solar), and it seems to work as expected. Forces PV limit to 0, then charges battery at max rate until target SoC is reached (nice ramp down of power as the target SoC is approached), then simply runs house load off the grid.

(Please ignore the sudden change of mode half way through, predbat decided to try and be helpful!)

image

infradom:
Nice, thanks !

PR #2030, opened 2026-05-11
Fix Pre-commit Errors in SolaX EV Charger Py.

TCWORLD:
The last round of changes introduced code that yeilds errors in pre-commit check script.

Fix errors in missing return typing, and also a potential call-method-on-None error.


wills106:
Thanks, I was working through the fixes but struggling with code-server.

PR #2033, opened 2026-05-12
Fix EV Charger breaking changes from PR #2012

KIvosak:
This PR fixes compatibility regressions introduced in #2012.

### Background
PR #2012 renamed several SolaX EV Charger phase entities from the integration-wide L1/L2/L3 convention to vendor-style A/B/C.

Across this integration, SolaX phase entities consistently use L1, L2, and L3. Changing EV Charger entity keys from *_l1, *_l2, *_l3 to *_a, *_b, *_c is a breaking change because Home Assistant unique IDs are based on entity keys.

For existing users, this can break:
- dashboards
- automations
- scripts
- history/statistics references
- custom cards and templates

### Changes
Restore SolaX EV Charger phase names and keys back to L1/L2/L3.
Keep vendor A/B/C wording only in code comments as L1 (A) / L2 (B) / L3 (C).
Add developer documentation for the SolaX phase naming convention.
Add a regression test to prevent SolaX plugin phase naming from drifting back to A/B/C.
Fix charge_duration Gen2 decoding by removing the incorrect order32="big" override, so the sensor uses the plugin default little-endian word order.

### Why
This keeps EV Charger entities compatible with existing Home Assistant installations and aligned with the rest of the SolaX integration.

### Tested on 
X3-HAC 11kW
Firmware: ARM v9.05

### DIFF
--- a/custom_components/solax_modbus/plugin_solax_ev_charger.py
+++ b/custom_components/solax_modbus/plugin_solax_ev_charger.py
@@ -519,9 +519,9 @@ SELECT_TYPES = [
         register=0x625,
         option_dict={
             0: "Three Phase",
-            1: "Phase A",
-            2: "Phase B",
-            3: "Phase C",
+            1: "L1 Phase",
+            2: "L2 Phase",
+            3: "L3 Phase",
         },
         entity_category=EntityCategory.CONFIG,
         icon="mdi:dip-switch",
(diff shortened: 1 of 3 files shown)


KIvosak:
@wills106  could you please consider merging and releasing this soon if the fix looks OK?

The phase key rename from #2012 changes HA unique IDs for existing EV Charger users and can break automations, dashboards and scripts. Releasing this quickly should help prevent duplicate issues from users upgrading into the breaking change.

wills106:
Thanks, I should have been more thorough in checking what had been changed.

Issue #2046, opened 2026-05-15
[Bug]: solax_pv_power_total stopped reporting power for Solax GEN6 inverter after upgrade to 2026.05.2


mpi:
### Describe the bug

After upgrade to 2026.05.2 plugin stopped reporting `pv_power_total` for my Solax X3 Hybird G4 inverter:

<img width="351" height="213" alt="Image" src="https://github.com/user-attachments/assets/93dd87e8-1e7d-445c-9820-824e616bf8cb" />

After rollback total power is back:
<img width="1088" height="449" alt="Image" src="https://github.com/user-attachments/assets/187afee1-f6db-444b-a7a0-b39a248fe448" />

I suspect this might be connected to this change:
https://github.com/wills106/homeassistant-solax-modbus/pull/2031

### Integration Version

2026.05.2

### Homeassistant core version

2026.5.1

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

10K0A1

### Inverter firmware versions

DSP 0.00 ARM 0.00

### Connection Method

Solax WiFi-LAN Dongle (via LAN)

### Dongle firmware

1.005.04

TCWORLD:
I was afraid of that.

Testing showed it worked fine on the VAST inverters but not on older ones hence the change was limited to GEN6. But clearly it is more problematic than expected.

I'll roll the change back to use the internally calculated register for PV limit.

Out of curiosity, does the `sensor.solax_battery_power_charge` sensor give readings for you?

Also, do you know what firmware version you have? I wonder if they only became available in newer firmware versions.

mpi:
> Out of curiosity, does the sensor.solax_battery_power_charge sensor give readings for you?

I don't have such sensor. I have `sensor.solax_inverter_solax_total_battery_power_charge`.
It reported `0 W` in `2026.05.2` version, currently it reports `Unavailable`:

<img width="981" height="456" alt="Image" src="https://github.com/user-attachments/assets/670a9ca2-f5bb-443b-a5ea-52694f3ac757" />



> Also, do you know what firmware version you have? I wonder if they only became available in newer firmware versions.

Solax Modbus reports DSP 0.00 ARM 0.00:

<img width="320" height="240" alt="Image" src="https://github.com/user-attachments/assets/4a8cd818-49ea-44ab-afb9-84598ac6f971" />


At the Inverter display I see: 
DSP 2.07 ARM 2.03

[...]

Issue #2028, opened 2026-05-08
[Bug]: SolaX X1-VAST (GEN6) PV3 Register Mapping Update for Firmware 018.04/016.05


heliveth:
### Describe the bug

Following a recent remote firmware upgrade from SolaX Support (to DSP 018.04 / ARM 016.05), the PV3/MPPT3 sensors stopped reporting data on the X1-VAST-10K.

SolaX Support provided an updated Modbus document confirming that for this firmware version, the registers for PV3 have been relocated. I have manually tested these addresses in plugin_solax.py and can confirm they restore full functionality for the third string.

Proposed Code Changes:
In plugin_solax.py, SolaX X1-VAST (Identified as GEN6 in the plugin), under the GEN6 category, the following registers should be updated:
pv_voltage_3: Changed to 0x122 (Previously 0x2A)
pv_current_3: Changed to 0x123 (Previously 0x2B)
pv_power_3: Changed to 0x124 (Previously 0x2C)

For instance:
`    SolaXModbusSensorEntityDescription(
        name="PV Power 3",
        key="pv_power_3",
        native_unit_of_measurement=UnitOfPower.WATT,
        device_class=SensorDeviceClass.POWER,
        state_class=SensorStateClass.MEASUREMENT,
        register=0x2C,
        register_type=REG_INPUT,
        allowedtypes=HYBRID | GEN6 | MPPT3 | MPPT4,
        icon="mdi:solar-power-variant",`
register=0x2C should be changed to 0x124 to restore the reading.

Evidence: Verified via manual file edit confirmed working, and official SolaX documentation (attached).

### Integration Version

2026.04.5

### Homeassistant core version

2026.5.0

### Inverter brand

SolaX

### Plugin used

plugin_solax.py

### Serial prefix

10M0A1

### Inverter firmware versions

DSP 018.04 ARM 016.05

### Connection Method

WiFi+LAN

### Dongle firmware

1.009.02

### Additional context

[(Solax) Hybrid X1&X3 ESS Modbus RTU V1.02-English_251126(1).pdf](https://github.com/user-attachments/files/27502279/Solax.Hybrid.X1.X3.ESS.Modbus.RTU.V1.02-English_251126.1.pdf)

gordanbanjac-hue:
I have the exact same bug since the firmware upgrade.

TCWORLD:
[...]

PR #2051, opened 2026-05-15
Revert SolaX Hybrid version number reporting to split registers

TCWORLD:
In #2000 version number reporting was changed to use the combined major/minor registers 0x7B and 0x7C which now exist in the latest protocol docs. However this doesn't work for older inverters and ones with outdated firmware (X1-IES, X3-G4-Pro, etc), meaning the version number reports back as 0.00.

For GEN4 revert to the old split registers approach.

For GEN5/6 use the new combined register if its value is non-zero, else use the old split register approach.

(#2046 shows example of incorrect version number decoding)

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1276,6 +1276,9 @@ def value_function_software_version_g3(initval: int, descr: Any, datadict: dict[
 
 
 def value_function_software_version_g4(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
+    if datadict.get("modbus_protocol_version", 0) >= 100:
+        # Use modern combined register if protocol version high enough
+        return value_function_software_version_full(initval, descr, datadict)
     return (
         f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '?')}."
         f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
@@ -1285,10 +1288,9 @@ def value_function_software_version_g4(initval: int, descr: Any, datadict: dict[
 
 
 def value_function_software_version_g5(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
-    if datadict.get("firmware_version_dsp"):
-        # Use modern combined register if it gives a sensible value
+    if datadict.get("modbus_protocol_version", 0) >= 100:
+        # Use modern combined register if protocol version high enough
         return value_function_software_version_full(initval, descr, datadict)
-    # Use the split approach if combined is unavailable.
     return (
         f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '???'):>03}."
         f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1285,6 +1285,10 @@ def value_function_software_version_g4(initval: int, descr: Any, datadict: dict[
 
 
 def value_function_software_version_g5(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
+    if datadict.get("firmware_version_dsp"):
+        # Use modern combined register if it gives a sensible value
+        return value_function_software_version_full(initval, descr, datadict)
+    # Use the split approach if combined is unavailable.
     return (
         f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '???'):>03}."
         f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
@@ -8290,8 +8294,14 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
     ),
     SolaXModbusSensorEntityDescription(
         key="software_version",
-        value_function=value_function_software_version_full,
-        allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6,
+        value_function=value_function_software_version_g4,
+        allowedtypes=AC | HYBRID | GEN4,
+        internal=True,
+    ),
+    SolaXModbusSensorEntityDescription(
+        key="software_version",
+        value_function=value_function_software_version_g5,
+        allowedtypes=AC | HYBRID | GEN5 | GEN6,
         internal=True,
     ),
     SolaXModbusSensorEntityDescription(
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -3987,7 +3987,6 @@ SENSOR_TYPES_MAIN: list[SolaXModbusSensorEntityDescription] = [
     SolaXModbusSensorEntityDescription(
         name="Modbus Protocol Version",
         key="modbus_protocol_version",
-        entity_registry_enabled_default=False,
         allowedtypes=AC | HYBRID | GEN4 | GEN5 | GEN6,
         register=0x82,
         entity_category=EntityCategory.DIAGNOSTIC,
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -1278,33 +1278,17 @@ def value_function_software_version_g3(initval: int, descr: Any, datadict: dict[
 def value_function_software_version_g4(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
     if datadict.get("modbus_protocol_version", 0) >= 100:
         # Use modern combined register if protocol version high enough
-        return value_function_software_version_full(initval, descr, datadict)
+        dsp = datadict.get("firmware_version_dsp")
+        arm = datadict.get("firmware_version_arm")
+        dsp_str = f"{dsp // 100}.{dsp % 100:02d}" if dsp is not None else "?.??"
+        arm_str = f"{arm // 100}.{arm % 100:02d}" if arm is not None else "?.??"
+        return f"DSP {dsp_str} ARM {arm_str}"
     return (
         f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '?')}."
         f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
         f"ARM {value_str_default(datadict.get('firmware_arm_major'), '?')}."
         f"{value_str_default(datadict.get('firmware_arm'), '??'):>02}"
     )
-
-
-def value_function_software_version_g5(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
-    if datadict.get("modbus_protocol_version", 0) >= 100:
-        # Use modern combined register if protocol version high enough
-        return value_function_software_version_full(initval, descr, datadict)
-    return (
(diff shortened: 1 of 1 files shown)


Bl00d-B0b:
I am not sure it's a good idea to revert it back.
I assume there different places available for versions, and it's possible to add condition.
maybe you can dump your modbus registers, and i can redevelop?

my registers are:

**X3-MIC/PRO-G2 — simple combined encoding**

| Addr | Raw | Decoded | Key | Cloud |
|---|---|---|---|---|
| `0x352` | 136 | `1.36` | `firmware_version_dsp` | MDSP 1.36 ✓ |
| `0x353` | 136 | `1.36` | `firmware_version_arm` | ARM 1.36 ✓ |
| `0x354` | 100 | `1.00` | `firmware_version_arm_sub` | -1.00 ✓ |

**X3-Hybrid-G4 — split major/minor encoding**

| Addr | Raw | Notes | Cloud |
|---|---|---|---|
| `0x7B` | 160 | `firmware_version_dsp` → `1.60` ✓ | MDSP 1.60 ✓ |
| `0x7C` | 158 | `firmware_version_arm` → `1.58` ✓ | ARM 1.58 ✓ |
| `0x7D` | 60 | `firmware_dsp` (minor only) | |
| `0x7F` | 1 | `firmware_dsp_major` | |
| `0x80` | 1 | `firmware_arm_major` | |
| `0x83` | 58 | ARM minor (unassigned) — matches 0x7C minor | |
| `0x84` | 15 | **ARM sub-minor** — with major from `0x80`=1 → `1.15` | -1.15 ✓ |

maybe you can provide your registers? then i can update a code

TCWORLD:
The combined registers 0x7B/0x7C on mine read as zero, in other words they don't exist in the inverter firmware so are just yeilding a default 0x0 response. This is a firmware released Dec 2025, so relatively modern.

In your table you say 0x83 is unassigned, but for all generations this is assigned to the `firmware_arm` entity. The combination of `(0x80).(0x83)` gives version 1.58 in your table which matches the combined register value. The same goes for `(0x7F).(0x7D)` which gives 1.60 in your table, again matching the combined register value. So the old approach should work absolutely fine for you. I could add the same check of whether `firmware_version_dsp` is non-zero to the g4 function too I suppose.

The `MIC` versioning is unaffected by this PR, it only targets `AC | HYBRID`.

Bl00d-B0b:
i can add condition, to check if registers  0x7B/0x7C are 0, then use split registers. Who knows, maybe in feature SolaX make decision to use only split registers:) let me fix it and push PR
P.S. 
Good findings from the doc. Key points relevant to our implementation:

**Confirmed correct:**
- `0x7B` = `FirmwareVersionDSP` (combined, added in spec V001.00) ✅
- `0x7C` = `FirmwareVersionARM` (combined, **only valid when `0x82 >= 100`**) ✅
- `0x7F` = `FirmwareVersionDSPMajor` ✅
- `0x7D` = `FirmwareVersionDSPMinor` ✅
- `0x80` = `FirmwareVersionARMMajor` ✅
- `0x83` = `FirmwareVersionARMMinor` ✅
- `0x84` = `FirmwareVersionARMBoot` ✅

**One improvement:** the doc says to use `0x82` (`FirmwareVersionModbus`) as the feature gate (`< 100` → use split registers), rather than checking if `0x7B == 0`. We already read `0x82` as `modbus_protocol_version` — let's use it properly:

TCWORLD:
What does your value for `modbus_protocol_version` (0x82) read? Mine is 0.

Bl00d-B0b:
mine is 100, this is exactly like in documentation. let me fix it and i send you my PR for review

TCWORLD:
Wait. No. Mine is 7, sorry.

So if yours is 1, then that check is not going to work...

---

I've go an old version installed because of the issues. (You got rid of Modbus Major, and replaced Modbus Minor Protocol Version. I mixed the two up when checking)

Bl00d-B0b:
could you dump your registers 0x7b-0x84
mine are:
Now I can do a proper side-by-side. Here's doc vs plugin vs live device:

| Addr | Doc name | Live GEN4 | Plugin key |
|---|---|---|---|
| `0x7B` | `FirmwareVersionDSP` | 160 | `firmware_version_dsp` |
| `0x7C` | `FirmwareVersionARM` | 158 | `firmware_version_arm` |
| `0x7D` | `FirmwareVersionDSPMinor` | 60 | `firmware_dsp` |
| `0x7E` | `HardwareVersion` | 0 | `firmware_DSP_hardware_version` |
| `0x7F` | `FirmwareVersionDSPMajor` | 1 | `firmware_dsp_major` |
| `0x80` | `FirmwareVersionARMMajor` | 1 | `firmware_arm_major` |
| `0x82` | `FirmwareVersionModbus` | 100 | `modbus_protocol_version` |
| `0x83` | `FirmwareVersionARMMinor` | 58 | `firmware_arm` |
| `0x84` | `FirmwareVersionARMBoot` | 15 | `firmware_arm_boot` |

TCWORLD:
Okay, that's fine, yours is 100. You said it was 1. We're all good. 

This PR should work as you suggest, checking the protocol version.

Bl00d-B0b:
and this valid for all HYBRID's I assume, does not require to filer by GEN's, etc. can be simplified. as now is too complex, let me push my PR for review.

TCWORLD:
The current restriction to GEN4/5/6 makes sense. The latest protocol doc doesn't mention any of the inverters older than a GEN4, so we shouldn't be making assumptions.

The only reason GEN5 was separated from GEN4 is because in the app shows the major version padded to 3 characters with leading zeros, so we were matching that. The two could be merged together I suppose, but it was nice to match the app format.

Bl00d-B0b:
yep. mine is math app format, but i do not have gen5 to test. let's me finalize it and push. i also cleanup unneeded functions.

TCWORLD:
I have a GEN5, so that is why it has the format it does - the app shows `024.99` not `24.99` as the DSP version and `016.04` as ARM version number.

Bl00d-B0b:
actually it should show for all invertors the same, like in documentation, 

ARM Firmware Full Version = ARM Major * 100 + ARM Minor.
So the current ARM firmware version is V001.09
FirmwareVersionARM is a full version with a content of 109 in the register.
Note:
FirmwareVersionARM is only supported for FirmwareVersionModbus ≥ V001.00 Version.

but I prefer to see all devices in a standard way, like v1.09, looks like much better :) standardized, and does not add additional code.

Bl00d-B0b:
one more question, in your device, which is looking for legacy registers, how does it show version in app:
DSP 1.60 1.58-1.15 or
DSP 1.60 1.58

my question does legacy way shows bootloader?

TCWORLD:
Screenshot_20260515-193618_ASUS_Launcher (1)

Bl00d-B0b:
ok, thanks, could you also show me your values of registers: 
0x7B | FirmwareVersionDSP | x| firmware_version_dsp
0x7C | FirmwareVersionARM | x| firmware_version_arm
0x7D | FirmwareVersionDSPMinor | x| firmware_dsp
0x7E | HardwareVersion | x| firmware_DSP_hardware_version
0x7F | FirmwareVersionDSPMajor | x| firmware_dsp_major
0x80 | FirmwareVersionARMMajor | x | firmware_arm_major
0x82 | FirmwareVersionModbus | x | modbus_protocol_version
0x83 | FirmwareVersionARMMinor | x | firmware_arm
0x84 | FirmwareVersionARMBoot | x | firmware_arm_boot

wills106:
When I first started to add in the GEN6 it used a different set of registers for some parts.

Am I correct in thinking that in newer Firmware versions they are aligning the GEN5 & GEN6 (Possibly GEN4?) to use the same register set?
So there could be a cutover point in Firmwares where the Inverter can / won’t work with the Integration?

Also is Solax.Hybrid.X1X3.ESS.Modbus.RTU.V1.02-English_251126.1 the latest version for the newer Inverters?

TCWORLD:
@wills106 The 1.02 doc is the latest I've seen. From what I can tell, going by model name, I think the common GEN4 through GEN6 seem to be included in that document. But only if they indicate a Modbus protocol version of 100 or higher. Technically 102 for the 1.02 doc.

I think we're potentially going to have more and more issues going forward with diverging firmware versions. For example the changes I made to PV to fix issues with the VAST inverter based on that latest document unfortunately broke the G4-Pro ones despite both being designated GEN6.

From what I can tell the people with VAST that it worked for had 100 for the proto, whilst the G4-Pro ones that broke are e.g. 005.

My own X1-IES (Gen5) has a proto num of 007, which probably explains why the inverter ID number doesn't match the ID table listed in that latest doc.

---

Fortunately so far there seems to be workarounds for each of the problematic registers, but we probably need to take extreme care going forward trying to add new registers.

rosenrot00:
@TCWORLD and @Bl00d-B0b I'm coming late to the game but couldn't we use the modbus version as the only differentiator? Or isn't that precise enough? Older software versions that don't or not correctly report the modbus version would use the current logic. All newer software versions get a new logic in which we set the registers accordingly?

@TCWORLD is this your suggestion from the last post?

TCWORLD:
@rosenrot00 Yeah, that's what I was wondering/thinking.

Perhaps add a GEN7 flag which is used for any inverter reporting a modern protocol version? 

My only concern with that is there are some differences between GEN5 and GEN6 that nobody has reported issues with - particularly around registers like Total Yield scaling which you'd think would have changed if the firmware updated to a unified API.

rosenrot00:
@TCWORLD and @wills106 I did some work previously when I played with the setup flow. Maybe this comes in handy? Instead of using GEN7 I would use a modbus_min/modbus_max property.

Back then I thought it might be overkill, now we are finally running into this issue.

https://github.com/rosenrot00/homeassistant-solax-modbus/tree/identity

It is bigger change but should be future proof.

TCWORLD:
@rosenrot00 That sounds like a better way of doing it. I guess we need to have the option for min version with unlimited max. Otherwise we'd be forever be having to update the max number with each firmware update even if unrelated. Only add the max when it becomes known a register has changed.

rosenrot00:
@TCWORLD this is exactly what I implemented.

I defined some of the registers with the attributes. I'm not sure about latest changes so maybe you could have a look if the registers that recently caused troubles are implemented correctly.

Maybe we then could do a pre-release to test.

Bl00d-B0b:
Yes, but when I ask Solax Support to update Modbus document, they told me it's a latest version they have. Then I write them several mistakes in document and ask to update, also to know which registers are available in which version. I get an answer. We forward you request to developer team, but we not sure answer will be so fast :)

TCWORLD:
Interesting development. It seems to totals registers (the ones that oddly have different scaling for GEN6) are actually giving incorrect readings on the VAST with proto>=100 according to #2028, so that's reassuring!

rosenrot00:
I created this PR https://github.com/wills106/homeassistant-solax-modbus/pull/2054 with the new modbus version logic.

Maybe with the action mentioned we could enable pre-releases for this PR so we could test rather fast.

TCWORLD:
Thanks, I'll have a look over it when I get a moment.

We're definitely going to need it as we now have some Gen6 VAST inverters with older firmware having problems with the fixes I made for the newer firmware Gen6 VAST inverters.

TCWORLD:
Closing as resolved by #2054

PR #2054, opened 2026-05-16
Introduce Solax registers depending on the inverters modbus version

rosenrot00:
Registers now support `modbus_min` and `modbus_max` properties, allowing compatibility across Modbus versions where newer releases introduce additional registers or change existing register addresses.

@wills106 could you create a pre-release as certainly something will break.

You could use this action and allow this PR to receive pre-releases. https://github.com/rosenrot00/homeassistant-solax-modbus/blob/main/.github/workflows/hacs-preview.yml

Settings → Secrets and variables → Actions → Variables

Create this variable:
HACS_PREVIEW_PR_NUMBERS

and set the PR number

2054,20XX

This will create a pre-release every 10 minutes when all test are ok. Pre-releases are deleted once the PR is closed or merged.

This will help us to move much faster and test with people.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -566,6 +566,7 @@ class SolaXModbusHub:
         self._validate_register_func = getattr(plugin, "validate_register_data", None)  # Cache function reference
         self.wakeupButton: Any = None
         self._invertertype: int | None = None
+        self.modbus_protocol_version: int | None = None
         self.localsUpdated: bool = False
         self.localsLoaded: bool = False
         self.config: Any = config  # MappingProxyType from entry.options
@@ -675,6 +676,8 @@ class SolaXModbusHub:
             model=getattr(self.plugin, "inverter_model", None),
             name=plugin_name,
             serial_number=self.seriesnumber,
+            sw_version=self.plugin.getSoftwareVersion(self.data),
+            hw_version=self.plugin.getHardwareVersion(self.data),
         )
 
         if getattr(self, "_stopping", False):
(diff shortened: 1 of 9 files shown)


wills106:
Am I merging this PR for now and releasing as a Beta and Ignoring #2051 & #2064 for now?
Just want to make sure I am following along correctly.

rosenrot00:
@wills106 for me this is the way forward. 

However, one remark about merging this PR, I would strongly suggest to try out the automatic pre-release build in order to become much faster in fixing and testing. At least test it and see if it is worthwhile. Such bigger PRs need testing with users in the field and having 4 days of delay between each update cycle is killing it a bit as I have to come back here and read again what I did the week before.

wills106:
> You could use this action and allow this PR to receive pre-releases. https://github.com/rosenrot00/homeassistant-solax-modbus/blob/main/.github/workflows/hacs-preview.yml
> 
> Settings → Secrets and variables → Actions → Variables
> 
> Create this variable: HACS_PREVIEW_PR_NUMBERS
> 
> and set the PR number
> 
> 2054,20XX

Does this go under:
Environment variables
Or
Repository variables

I guess I also need to pull in hacs-preview.yml into this repository?

rosenrot00:
@wills106

Path:
Repository Settings → Secrets and variables → Actions → Variables → New repository variable

Set:
HACS_PREVIEW_PR_NUMBERS=2068,124,130
Use comma-separated PR numbers. These are the PRs for which preview pre-releases may be built.

Yes, you also need .github/workflows/hacs-preview.yml in that repository, ideally already merged into the default branch, usually main.

Actually for the pre-releases to be generated you need to keep the PR open, so as long as it is open it generates pre-releases on changes whenever the checks are good.

I created https://github.com/wills106/homeassistant-solax-modbus/pull/2068 as a placeholder to test.

rosenrot00:
Actually there was an error in the`__init__.py` which I overlooked. But it is now corrected with #2068

johny-mnemonic:
Not sure if this is related, but since updating to latest release I see no versions for my X3 Gen4: `Firmware: DSP 0.00 ARM 0.00`

PR #2061, opened 2026-05-18
fix: key today_s_solar_energy renamed to today_s_yield

jorritsmit:
Renamed key in energy dashboard for solar production energy from "today_s_solar_energy" renamed to "today_s_yield"

This was needed for my X3-MIC-G2 and tested to work

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -10590,14 +10590,21 @@ ENERGY_DASHBOARD_MAPPING = EnergyDashboardMapping(
             needs_aggregation=True,
             allowedtypes=GEN3 | GEN4 | GEN5 | GEN6,
         ),
-        # Solar Production Energy (GEN2-6 today)
+        # Solar Production Energy (AC and Hybrid GEN2-6 today)
         # Aggregate energy totals across Primary + Secondary in parallel mode.
         EnergyDashboardSensorMapping(
             source_key="today_s_solar_energy",
             target_key="solar_energy_production",
             name="Solar Production Energy",
             needs_aggregation=True,
-            allowedtypes=GEN2 | GEN3 | GEN4 | GEN5 | GEN6,
+            allowedtypes=AC | HYBRID | GEN2 | GEN3 | GEN4 | GEN5 | GEN6,
+        ),
+        # Solar Production Energy ( MIC today)
+        EnergyDashboardSensorMapping(
+            source_key="today_s_yield",
+            target_key="solar_energy_production",
+            name="Solar Production Energy",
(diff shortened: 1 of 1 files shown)


TCWORLD:
This change is not correct for most of the supported inverters.

The `Todays Yield` and `Todays Solar Energy` sensors report different things on Hybrid inverters.

---

Edit: You should instead change to:

```
        EnergyDashboardSensorMapping(
            source_key="today_s_solar_energy",
            target_key="solar_energy_production",
            name="Solar Production Energy",
            needs_aggregation=True,
            allowedtypes=AC | HYBRID | GEN2 | GEN3 | GEN4 | GEN5 | GEN6,
        ),
        EnergyDashboardSensorMapping(
            source_key="today_s_yield",
            target_key="solar_energy_production",
            name="Solar Production Energy",
            allowedtypes=MIC | GEN | GEN2 | GEN4,
        ),
```

jorritsmit:
dont have information on other devices so thanks for investigating. two questions though:

1. Why no aggregation for the yield source key?
2. Should remove the riemann sum sensormapping for the GEN? --> no i guess not because it will cover the non-mic devices

TCWORLD:
1. Aggregation flag is only needed for inverters supporting parallel mode which the MIC series don't AFAIK.
2. If there are others which don't apply to the MIC ones, then you should be able to apply `AC | HYBRID |` to the start of the allowed types to filter them out (I'd guess that applies to all the battery ones too).

TCWORLD:
Re point 2, I see what you mean now. For the riemann sum Solar Production Energy below, you'll need to change (line 10618) to `allowedtypes=AC | HYBRID | GEN` too.

jorritsmit:
> Re point 2, I see what you mean now. For the riemann sum Solar Production Energy below, you'll need to change (line 10618) to `allowedtypes=AC | HYBRID | GEN` too.

Updated

TCWORLD:
@wills106 This looks fine to merge now.

custom_components/solax_modbus/plugin_solax.py
```diff
@@ -10608,7 +10615,7 @@ def localDataCallback(self, hub: Any) -> bool:
             use_riemann_sum=True,
             filter_function=lambda v: max(0, v),
             needs_aggregation=True,
-            allowedtypes=GEN,
+            allowedtypes= AC | HYBRID | GEN,
```
TCWORLD:
```suggestion
            allowedtypes=AC | HYBRID | GEN,
```
Extra whitespace causing CI to fail.

jorritsmit: implemented

PR #2064, opened 2026-05-18
fix(solax): unify firmware version display and key naming across all inverter types

Bl00d-B0b:
## Related PRs

- #1839 — established the split-register approach for older protocol (Gen4/Gen5)
- #2014 — extended split-register pattern to GEN3
- #2000 — introduced combined registers 0x7B/0x7C for GEN4/5/6, but was too aggressive (see #2051)
- #2051 — reported that unconditional combined registers break older GEN4/X1-IES firmware
- #2054 — proposed `modbus_min`/`modbus_max` read-time register gating — **this has since landed in 2026.06.7**, and this PR now builds directly on it

## Problem

Firmware version sensors didn't match SolaX Cloud, and firmware key naming was inconsistent across inverter families:

1. **MIC GEN2** (X1 & X3): registers 0x352/0x353 hold the full version ÷100 encoded (136 = 1.36), but the display functions treated them as raw minors with hardcoded prefixes → `DSP v2.136 ARM v2.136`
2. **AC/HYBRID GEN4+**: the ARM bootloader suffix shown by SolaX Cloud (e.g. `1.58-1.15`) was missing even though both source registers were already polled
3. **Key naming**: `firmware_dsp` meant "minor version" for AC/HYBRID (0x7D) but "full version" for MIC (0x352) — no rule to know which was which

## Solution (aligned with 2026.06.7)

This PR is rebased onto 2026.06.7 and **builds on the infrastructure that release introduced** rather than duplicating it:

- **Read-time gating stays**: 0x7B/0x7C keep `modbus_min=100`, so the wrong register set is filtered out before polling — exactly the approach discussed with @rosenrot00 above
- **Early firmware read stays**: `async_read_inverter_firmware_info` is kept, only its output keys are renamed to follow the new naming rule

**Naming rule established**: `firmware_dsp` / `firmware_arm` always hold the full ÷100 value, everywhere. Split components are `firmware_dsp_minor` / `firmware_arm_minor` (0x7D/0x83). The name now states exactly what the register holds.

**One unified `value_function_software_version`** replaces `g4`/`g5`/`full`/`protocol_aware`/`air_g3`/`air_g4` for AC/HYBRID GEN4+ and all MIC:

- AC/HYBRID, proto ≥ 100: full values from 0x7B/0x7C
- AC/HYBRID, proto < 100 or 0x7B unavailable: major × 100 + minor (split registers) — no regression for older firmware (#2051 scenario)
- ARM bootloader suffix from `firmware_arm_major` (0x80) + `bootloader_version` (0x84) → `1.58-1.15`, matching SolaX Cloud
- MIC: full ÷100 values directly; GEN2 adds the boot suffix from new internal sensor 0x354 (÷100) → `1.36-1.00`, matching SolaX Cloud
- `MU802T` blacklist preserved on the collapsed MIC entry

**Deliberately conservative**: AC/HYBRID GEN2/GEN3 keep their legacy per-gen display functions (adapted to the renamed keys). Their split registers hold raw minors with implicit majors ("v2.x"/"v3.x") and no device data confirms the unified format for them — changing their output without evidence would be guesswork.

## Register Map

**AC/HYBRID** (gated on `modbus_protocol_version`, register 0x82):
```
0x7B  firmware_dsp          full ÷100 value      modbus_min=100
0x7C  firmware_arm          full ÷100 value      modbus_min=100
0x7D  firmware_dsp_minor    split minor          (renamed from firmware_dsp)
0x7F  firmware_dsp_major    split major
0x80  firmware_arm_major    split major / boot major
0x83  firmware_arm_minor    split minor          (renamed from firmware_arm)
0x84  bootloader_version    boot minor           (existing sensor, now used in display)
```

**MIC** (no 0x82; registers hold full ÷100 values):
```
GEN1  0x33D/0x33E   firmware_dsp / firmware_arm
GEN2  0x352/0x353   firmware_dsp / firmware_arm   + 0x354 firmware_arm_boot (new, ÷100)
GEN4  0x394/0x390   firmware_dsp / firmware_arm
```

Note: 0x354 is not in the MIC-G2/PRO-G2 Modbus V3.1 document, but a real X3-MIC/PRO-G2 returns the boot version there, matching the "ARM Version 1.36-1.00" shown in SolaX Cloud.

## Verified on real devices (SolaX Cloud vs HA side by side)

| Device | Before | After | SolaX Cloud |
|---|---|---|---|
| X3-MIC/PRO-G2 8kW | `DSP v2.136 ARM v2.136` | `DSP 1.36 ARM 1.36-1.00` | MDSP 1.36 / ARM 1.36-1.00 |
| X3-Hybrid-G4 15kW (proto=100) | `DSP 1.60 ARM 1.58` | `DSP 1.60 ARM 1.58-1.15` | MDSP 1.60 / ARM 1.58-1.15 |

Function-level tests additionally cover: legacy protocol path (proto < 100), proto ≥ 100 with failed 0x7B read (falls back to split), MIC GEN4, and partial/empty data (`?.??` placeholders, no exceptions).

## Reference Documents

| Document | Version | Relevance |
|---|---|---|
| Solax Hybrid X1/X3 ESS Modbus RTU | V1.02 (Nov 2025) | ✅ Primary source — confirms 0x7B–0x84 and the 0x82 gate |
| Solax MIC-G2/PRO-G2 Modbus RTU | V3.1 | ✅ Confirms MIC GEN2 registers 0x352/0x353 |
| Solax X1-MINI/BOOST-G4 Modbus RTU | V2.3 | ✅ Confirms MIC GEN4 registers 0x390/0x394 |

## Testing wanted

Confirmation welcome from: GEN3, GEN5, GEN6, older GEN4 with proto < 100 (X1-IES), MIC GEN1, MIC GEN4. Please report your 0x82 value and whether the displayed version matches your SolaX Cloud "Device Info" page.

### DIFF
--- a/custom_components/solax_modbus/plugin_solax.py
+++ b/custom_components/solax_modbus/plugin_solax.py
@@ -267,12 +267,12 @@ async def async_read_inverter_firmware_info(hub: Any) -> int:
         inverter_data = await hub.async_read_holding_registers(unit=hub._modbus_addr, address=0x7D, count=8)
         if inverter_data is not None and not inverter_data.isError():
             registers = inverter_data.registers
-            data["firmware_dsp"] = int(registers[0])
+            data["firmware_dsp_minor"] = int(registers[0])
             data["firmware_DSP_hardware_version"] = int(registers[1])
             data["firmware_dsp_major"] = int(registers[2])
             data["firmware_arm_major"] = int(registers[3])
             version = int(registers[5])
-            data["firmware_arm"] = int(registers[6])
+            data["firmware_arm_minor"] = int(registers[6])
             data["bootloader_version"] = int(registers[7])
     except Exception:
         _LOGGER.debug(f"{hub.name}: attempt to read inverter firmware info failed data: {inverter_data}", exc_info=True)
@@ -284,8 +284,8 @@ async def async_read_inverter_firmware_info(hub: Any) -> int:
     else:
         hub.modbus_protocol_version = None
         data.pop("modbus_protocol_version", None)
-        data.pop("firmware_version_dsp", None)
-        data.pop("firmware_version_arm", None)
+        data.pop("firmware_dsp", None)
+        data.pop("firmware_arm", None)
         _LOGGER.debug(f"{hub.name}: Modbus protocol document version unavailable")
         return 0
 
@@ -294,13 +294,13 @@ async def async_read_inverter_firmware_info(hub: Any) -> int:
         try:
             full_version_data = await hub.async_read_holding_registers(unit=hub._modbus_addr, address=0x7B, count=2)
             if full_version_data is not None and not full_version_data.isError():
-                data["firmware_version_dsp"] = int(full_version_data.registers[0])
-                data["firmware_version_arm"] = int(full_version_data.registers[1])
+                data["firmware_dsp"] = int(full_version_data.registers[0])
+                data["firmware_arm"] = int(full_version_data.registers[1])
         except Exception:
             _LOGGER.debug(f"{hub.name}: attempt to read full firmware version failed data: {full_version_data}", exc_info=True)
     else:
-        data.pop("firmware_version_dsp", None)
-        data.pop("firmware_version_arm", None)
+        data.pop("firmware_dsp", None)
+        data.pop("firmware_arm", None)
 
     return version
 
@@ -1351,7 +1351,8 @@ def value_function_battery_capacity_gen5(initval: int, descr: Any, datadict: dic
 
 
 def value_function_software_version_g2(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
-    return f"DSP v2.{datadict.get('firmware_dsp')} ARM v2.{datadict.get('firmware_arm')}"
+    # AC/HYBRID GEN2: split registers hold raw minor versions with a fixed major of 2
+    return f"DSP v2.{datadict.get('firmware_dsp_minor')} ARM v2.{datadict.get('firmware_arm_minor')}"
 
 
 def value_function_firmware_major_default(val: Any, default: int) -> Any:
@@ -1361,27 +1362,9 @@ def value_function_firmware_major_default(val: Any, default: int) -> Any:
 def value_function_software_version_g3(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
     return (
         f"DSP v{value_function_firmware_major_default(datadict.get('firmware_dsp_major'), 3)}."
-        f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
+        f"{value_str_default(datadict.get('firmware_dsp_minor'), '??'):>02} "
         f"ARM v{value_function_firmware_major_default(datadict.get('firmware_arm_major'), 3)}."
-        f"{value_str_default(datadict.get('firmware_arm'), '??'):>02}"
-    )
-
-
-def value_function_software_version_g4(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
-    return (
-        f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '?')}."
-        f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
-        f"ARM {value_str_default(datadict.get('firmware_arm_major'), '?')}."
-        f"{value_str_default(datadict.get('firmware_arm'), '??'):>02}"
-    )
-
-
-def value_function_software_version_g5(initval: int, descr: Any, datadict: dict[str, Any]) -> str | None:
-    return (
-        f"DSP {value_str_default(datadict.get('firmware_dsp_major'), '???'):>03}."
-        f"{value_str_default(datadict.get('firmware_dsp'), '??'):>02} "
-        f"ARM {value_str_default(datadict.get('firmware_arm_major'), '???'):>03}."
-        f"{value_str_default(datadict.get('firmware_arm'), '??'):>02}"
+        f"{value_str_default(datadict.get('firmware_arm_minor'), '??'):>02}"
     )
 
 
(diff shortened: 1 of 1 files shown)


rosenrot00:
I think the firmware key cleanup makes sense, but I wonder if we should solve it on top of the Modbus protocol version gating from #2054 instead of adding a separate display-time switch.

With #2054 we can make the register semantics explicit:

- protocol >= 100:
  - 0x7B -> firmware_dsp
  - 0x7C -> firmware_arm
  - both are full div100 firmware values

- protocol < 100:
  - 0x7D -> firmware_dsp_minor
  - 0x7F -> firmware_dsp_major
  - 0x83 -> firmware_arm_minor
  - 0x80 -> firmware_arm_major

Then the software_version value function can simply prefer the full values when present and otherwise compose the version from major/minor.

That would keep the naming cleanup from this PR, but avoids reading registers that do not exist on older firmware, because the wrong register set is filtered out before polling.

Would that cover the same issue in a cleaner way, or is there a case where this gating is still needed?

Bl00d-B0b:
Thanks for the suggestion — a few thoughts on why I'd keep this PR as-is rather than building on top of #2054, plus a counterproposal for the read-time gating idea.

**Naming consistency**

Keeping `firmware_dsp` as the minor component (0x7D) while adding `firmware_version_dsp` for the combined value (0x7B) leaves two inconsistently-named keys with no clear rule for which is which. This PR establishes one rule: `firmware_dsp` / `firmware_arm` always means the full ÷100 combined value everywhere — AC/HYBRID and MIC. Minor components are renamed `firmware_dsp_minor` / `firmware_arm_minor` so the name tells you exactly what it is. If #2054 landed first it would need to adopt this naming anyway, so merging this first avoids introducing the inconsistency and then fixing it.

**`modbus_min` / `modbus_max` — documentation and version confusion**

#2054 assigns `modbus_min=101` to 25 registers and `modbus_min=102` to 9 registers. We only have one published spec (Modbus RTU V1.02 PDF) — there is no V1.00 or V1.01 document to confirm which register was introduced at which version. Those values appear inferred from device testing, not from a published spec.

There is also a fundamental confusion between the PDF revision and what register `0x82` actually returns. From real devices reported in this thread:

| Device | proto (0x82) |
|---|---|
| X3-Hybrid G4 PRO | 100 |
| X1-IES GEN5 | 7 |
| G4-Pro older firmware | 5 |

No real device is known to return 101 or 102 from `0x82`. The assumption that the V1.02 PDF → proto=102 is not confirmed by any device data. Assigning `modbus_min=102` today means those registers would be hidden for everyone, since no device in this thread returns that value.

The only threshold explicitly documented in the spec is: *"FirmwareVersionARM is only valid for FirmwareVersionModbus ≥ V001.00"* — i.e. `proto >= 100`. That is the only gate anchored to real documentation and confirmed with real device data.

**A simpler approach for read-time gating**

The read-time filtering idea from #2054 is useful, but `modbus_min`/`modbus_max` is more infrastructure than needed. We already read `modbus_protocol_version` and already filter entities using `allowedtypes` — the same pattern can be extended with a single `proto` field on the entity descriptor:

```python
    key="firmware_dsp",
    register=0x7B,
    allowedtypes=AC | HYBRID,
    proto=100,   # only read when device proto >= 100
    internal=True,
)
```

The hub checks at read time: if `entity.proto` is set and `device_proto < entity.proto`, skip the register — same logic as `matchInverterWithMask` already does for GEN flags. One field, no max needed (registers don't get removed, only added), and the only value we can confidently assign today is `100`.

**PR recommendation**

- **Merge this PR (#2064)** — solves the immediate firmware display problem with consistent naming and the documented threshold
- **Close #2051** — superseded by this PR
- **Rework #2054** — adopt the naming from this PR, replace `modbus_min`/`modbus_max` with a single `proto` field, and start only with `proto=100` (the one value confirmed from real devices and documentation). Can be built cleanly on top of this PR once it lands.

Happy to coordinate on a follow-up for the `proto` field once this lands.

TCWORLD:
Min-max makes more sense than a single field. Otherwise if a register is removed we have no way of indicating that, which is a large part of the problem it's trying to resolve - registers do get removed, and moved. The battery total capacity moved.

Let's get #2054 tested and merged.

As suggested by @rosenrot00 this one (and mine) becomes largely redundant as the protocol version limits will already select which registers to read. Some of the cleanup can be reapplied afterwards.

Bl00d-B0b:
Rebased onto 2026.06.7 (single commit, conflicts resolved). Notable changes versus the previous revision:

- Now **builds on** the `modbus_min` register gating and early firmware read that landed in 2026.06.7, instead of conflicting with them — this addresses the read-time gating discussion above: 0x7B/0x7C keep their `modbus_min=100` so the wrong register set is filtered out before polling.
- Key naming rule unchanged: `firmware_dsp`/`firmware_arm` = full ÷100 value everywhere; split components renamed to `firmware_dsp_minor`/`firmware_arm_minor`.
- One unified `value_function_software_version` replaces `g4`/`g5`/`full`/`protocol_aware`/`air_g3`/`air_g4` for AC/HYBRID GEN4+ and all MIC. ARM bootloader suffix now included, matching SolaX Cloud.
- AC/HYBRID **GEN2/GEN3 keep their legacy display functions** (adapted to the renamed keys) — their split registers hold raw minors with implicit majors and I have no device data to confirm the unified format there. Safer to leave them.
- `MU802T` blacklist preserved on the collapsed MIC entry.
- Dropped the 0x7E → `hardware_version` rename from the previous revision: upstream now computes `hardware_version` per-gen, so the rename would collide.

Live-verified on two devices (SolaX Cloud vs HA side by side):

| Device | Before | After | SolaX Cloud |
|---|---|---|---|
| X3-MIC/PRO-G2 | `DSP v2.136 ARM v2.136` | `DSP 1.36 ARM 1.36-1.00` | MDSP 1.36 / ARM 1.36-1.00 |
| X3-Hybrid-G4 15kW | `DSP 1.60 ARM 1.58` | `DSP 1.60 ARM 1.58-1.15` | MDSP 1.60 / ARM 1.58-1.15 |

Bl00d-B0b:
@TCWORLD @rosenrot00 — updating now that the protocol-gating work has landed in 2026.06.7:

- **The min/max gating is in and this PR builds on it.** `modbus_min`/`modbus_max` now exist upstream, and the firmware registers here carry `modbus_min=100` (0x7B/0x7C) so the wrong register set is filtered out before polling — exactly the read-time approach you both asked for. This PR no longer adds a competing mechanism; it sits on top of the merged one.

- **Gating alone does not fix the reported bug, though.** The protocol version selects *which registers are read*, but the remaining defect is *how the value is formatted*: MIC GEN2 read `0x352`/`0x353` = `136` and displayed `DSP v2.136 ARM v2.136` (a hardcoded prefix on a ÷100-encoded value). That is a display-function bug, orthogonal to register selection, and is what this PR fixes → `DSP 1.36 ARM 1.36-1.00`, matching SolaX Cloud. Verified on two real devices (X3-MIC/PRO-G2 and X3-Hybrid-G4).

- **What survives from the original PR:** the single unified `value_function_software_version` (replacing g4/g5/full/protocol_aware/air_g3/air_g4) plus the key-naming rule (`firmware_dsp`/`firmware_arm` = full ÷100 value everywhere; split parts → `_minor`). AC/HYBRID GEN2/GEN3 keep their legacy per-gen display functions since no device data confirms the unified format for them.

So the "largely redundant after gating" concern applies to the register-selection half (now dropped/rebased away), but the display-format fix and naming cleanup are still needed. Happy to split the naming cleanup out if you'd prefer the pure bugfix on its own.

custom_components/solax_modbus/plugin_solax.py
```diff
@@ -1392,31 +1375,51 @@ def value_function_modbus_protocol_version(datadict: dict[str, Any]) -> int:
...
+    AC/HYBRID — gated on modbus_protocol_version (0x82, only read for AC/HYBRID):
+      >= 100 — full values from the combined registers 0x7B (DSP) / 0x7C (ARM)
+      <  100 — composed as major × 100 + minor (0x7F/0x7D for DSP, 0x80/0x83 for ARM)
+      ARM bootloader suffix: firmware_arm_major (0x80) + bootloader_version (0x84, raw
+      minor) — matches SolaX Cloud "ARM Version 1.58-1.15".
 
+    MIC — 0x82 absent (proto is None); registers hold the full ÷100 value directly:
+      GEN1 — 0x33D/0x33E   GEN2 — 0x352/0x353 (+ 0x354 boot, ÷100)   GEN4 — 0x394/0x390
+      Matches SolaX Cloud "ARM Version 1.36-1.00".
+    """
```
TCWORLD: I'm not entirely convinced of the benefit of merging this all into a single function.

Going through it, there are lots of if-else branches used to split everything back up into their own groupings (MIC/AC-Hybrid), and the only common bit is the two formatting lines.

Similarly for the proto<100, you are basically now just multiplying the register values together to form the combined version, only to split them apart again when rendering. That seems unnecessary compared to just having a separate function to handle that case.

I agree the MIC needs fixing and the GEN4+ ones can be simplified (we don't need one for each generation), however I suspect we are much better off having three functions:

1. MIC specifically including your display fixes,
2. AC/Hybrid proto<100
3. AC/Hybrid proto>=100

TCWORLD (changes requested):

Bl00d-B0b:
@TCWORLD — done, split into the three functions you suggested (force-pushed, single commit):

1. **`value_function_software_version_mic`** — MIC (all gens), including the display fix + 0x354 ARM bootloader suffix.
2. **`value_function_software_version_hybrid_legacy`** — AC/HYBRID protocol < 100. Renders `major.minor` **directly** from the split registers now — no more multiplying to a combined value only to divide it back apart, exactly the round-trip you flagged.
3. **`value_function_software_version_hybrid_full`** — AC/HYBRID protocol >= 100, full ÷100 value from 0x7B/0x7C.

A thin `value_function_software_version` dispatches AC/HYBRID to full vs legacy on the protocol version (this is the only remaining branch — a 3-line router, matching the previous `protocol_aware` pattern). GEN2/GEN3 keep their existing per-gen functions.

No shared if-else grouping left; each inverter type has its own focused formatter. Behaviour re-verified against both real devices:

| Device | Result |
|---|---|
| X3-MIC/PRO-G2 (proto absent) | `DSP 1.36 ARM 1.36-1.00` |
| X3-Hybrid-G4 (proto 100) | `DSP 1.60 ARM 1.58-1.15` |
| AC/HYBRID legacy (proto < 100) | `major.minor` rendered directly |

Agreed the register-selection concern is handled by the merged `modbus_min` gating; this PR is now purely the display fix + per-type split.

TCWORLD:
Looks fine to me now.

wills106:
I probably should have another look to see if the GEN3 actually return a sensible mayor version now. Probably 4 or 5 years since I last checked.

PR #2071, opened 2026-05-21
Additional Hybrid EPS Entities

TCWORLD:
Add controls for EPS Min SoC and EPS Restart SoC registers. These are essentially overrides for the Self-Use Min SoC when operating in EPS mode.

Update sensor entity protocol limits for EPS Min SoC/Set Frequency. These are supported (albeit undocumented) on older protocol versions. 

The Set Frequency is technically also a control, but I've left it as sensor-only because writing to this is potentially dangerous and should never really be changed.


wiltet:
First I have to thank for your work!
I only want to say something to not forget about perhaps by accident.

"The Set Frequency is technically also a control, but I've left it as sensor-only because writing to this is potentially dangerous and should never really be changed."

And always keep in mind that in MicroGrid-Mode, with a connected extra On-Grid-Inverter, that this one gets connected to the EPS circuit and it is controlled by EPS-Over-Frequency. It would be dangerous to change freq for that functionality too.

PR #2073, opened 2026-05-21
Fix SolaX Battery Input/Output Energy (protocol<100) and Total Energy Sensors (protocol≥100)

TCWORLD:
PR #2054 stopped the battery energy totals entities appearing for GEN4/GEN5 inverters with old firmware. However these do exist for GEN4/5 on older protocol versions, so revert the protocol gating for these registers.

Additionally split these sensors for GEN6 into pre/post-100 protocol versions as the scaling is incorrect in the newer version. The same applies for Total Yield and Total Solar Energy. This was confirmed in https://github.com/wills106/homeassistant-solax-modbus/issues/2028#issuecomment-4468998499.


rosenrot00:
@TCWORLD thanks for cleaning up. I wasn't following the register discussions closely, so any help here to get the right registers for the right modbus version going is highly appreciated.

PR #2050, opened 2026-05-15
Revert changes to PV Power Total for GEN6

TCWORLD:
The new totals register doesn't appear to work on the G4-Pro systems which are also marked as GEN6, yet works fine on the VAST systems.

Revert to use the old approach of summing all individual PV power sensors for all systems as this seems more stable across firmware versions.

Fixes #2046 and fixes #2049.

PR #2031, opened 2026-05-12
Fix MPPT Register Mappings for SolaX GEN6 (e.g. X1-VAST)

TCWORLD:
This PR updates some of the register maps for GEN6 inverters based on the latest protocol docs

1. Correct the location of MPPT3 and MPPT4 registers based on latest protocol docs. Fixes #2028. Possibly #1991.
2. Correct location of `battery_total_capacity_charge`. This applies to GEN5 as well. The previously used register used does not exist in the current docs (and seems to have always returned 0). The new register does exist in the docs.
3. Add missing `battery_power_charge` sensor for GEN6. Rather than using the sum of batteries approach used for GEN5, investigate a new register in the protocol docs which should give the correct sum value.
4. Switch `pv_power_total` to use the actual totals register from the inverter rather than summing them all individually.

PR #2077, opened 2026-05-21
Reinstate PV Total Register for newer GEN6 (protocol≥100)

### LINKED ISSUE #2080: [Bug]: Solax Battery Power sensor not working
### Describe the bug

Hey,

the "SolaX Battery Power" Energy Dashboard senor does not work, also the "SolaX Total Battery Power Charge" sensor directly from the inverter device shows nothing.

Both should show something, also when the battery is discharging.

The "Solax Battery X Power Charge" sensor seems, from the inverter device seems to work fine.

Image

Image

Image

Image

### Integration Version

2026.05.3

2026.5.2

### Inverter brand

SolaX Power X3-HYB-G4 PRO 

### Plugin used

plugin_solax.py

### Serial prefix

10K0F1

ARM: 018.06-000.06 / MDSP 018.04 / DSP 000.00 / ARC 000.00

### Connection Method

Pocket WiFi+LAN (connected via LAN)

### Dongle firmware

Firmware: 1.002.09 / Internal Code: 1.008.05

### Detailed Error Log

_No response_

### Additional context

_No response_

TCWORLD:
Reinstate the changes for `PV Power Total` register.

This is one of those whack-a-mole PRs. It was originally added in #2031, but then removed in #2050 due to problems with lack of support on older firmware versions.

Now that we have protocol aware register mappings, this change can be made in a more targetted way. Based on reviewing all the recent issues around the change:
* All those that reported it as breaking have `protocol<100`, so we do not apply the change for those. 
* People indicating they have `protocol≥100` reported this register works,
* There was one case where it was not working when `protocol<100` then started working after SolaX updated the firmware to `protocol≥100` which indicates this should play nicely as inverters are upgraded over time.

Furthermore, we also add a missing `battery_power_charge` for older GEN6 inverters (protocol<100) which fixes #2080.


wills106:
Was going to merge this and come across the conflict, that needs resolving.
Why would there be a conflict for PV Power 4, when the PR doesn’t alter that?

TCWORLD:
Fixed the conflict. It was because two of my PRs touched just either side of the same line - the change here was actually just the comment being deleted.

PR #2084, opened 2026-05-22
Handle Modbus task cancellation cleanly during shutdown

rosenrot00:
Collect cancelled in-flight Modbus tasks during shutdown and downgrade expected pymodbus cancellation errors to debug logging. This avoids noisy unhandled shutdown exceptions when Home Assistant stops while a Modbus read is still in progress.


wills106:
Just looking at the conflict.

Am I correct in thinking we need both:
INFLIGHT_CANCEL_TIMEOUT & CONNECT_RETRY_DELAY

I can't see this PR removing any CONNECT_RETRY_DELAY parts.

PR #2086, opened 2026-05-23
Add support for Solax 3-NEO-10K-LV

rosenrot00:
Solves https://github.com/wills106/homeassistant-solax-modbus/discussions/2083


wills106:
I think this Inverter should use solax_lv need to check.

wills106:
NEO uses a different set of registers to the Hybrids and the current plugin_solav_lv
So will most likely need it's own plugin https://github.com/wills106/homeassistant-solax-modbus/issues/2129

PR #2095, opened 2026-05-30
Fix local entity state restore after number polling was disabled

rosenrot00:
This fixes local WRITE_DATA_LOCAL entities staying unknown after Home Assistant restart.

The regression was introduced by my commit 4770711, which added should_poll = False to number entities. That change is correct for Modbus-backed entities because their state is pushed by the hub, but local entities such as remotecontrol_timeout have no Modbus register and are intentionally skipped from hub polling. As a result, their persisted values were loaded into hub.data but never pushed to Home Assistant.

This PR keeps should_poll = False and keeps local entities out of Modbus polling. Instead, local number/select/time/switch entities now refresh their HA state when local data has been loaded.

@wills106 please push another release with this fix.


wills106:
Are you able to look at the failed CI/CD. Then I can do a release.

PR #2096, opened 2026-05-30
Handle in-flight Modbus reads during Home Assistant shutdown

rosenrot00:
Stopping SolaX Modbus hubs on the Home Assistant stop event and collect expected pymodbus cancellation results from in-flight read tasks during shutdown. This prevents "Request cancelled outside pymodbus" from leaking as an unhandled shutdown exception while preserving normal Modbus error handling during runtime.


wills106:
Are you planning to include this in 2026.05.5? Or leave it for a future release?

rosenrot00:
You can add this one to 2026.05.05 as well. Not a dramatic one but currently we cause errors when HA restarts.

PR #2097, opened 2026-05-30
Initialize computed Energy Dashboard sensors on add

rosenrot00:
Computes register=-1 sensors once when they are added to Home Assistant and registers them in hub.computedSensors, so Energy Dashboard diagnostics like Mode, Parallel Setting, and Inverter Count do not remain unknown.


rosenrot00:
@wills106 also add this one as well as there where unknowns in the energy board. This fixes it.

Quickly tested all three PRs on my end and now it looks good.

Sorry for the inconvenience, sometimes changing a small thing causes more errors than expected.

wills106:
I didn't even think to check the dashboard. I just checked the main part worked my end.

Issue #2085, opened 2026-05-22
[Bug]: Multiple sensors are missing for GEN6 inverter


gabriel31337:
### Describe the bug

I am constantly running into missing sensors for GEN6 inverter, previously it was inverter temperature, now I have noticed State of Health was not available at all. Some of the sensors were disabled in the entities list, I had to manually enable them in the HA menu to get the values.

I have manually enabled SOH in the python script by adding GEN6 as an option and it shows the correct value. Would it be possible to go through the sensors enabled for the X3 Hybrid G4 and match it for the X3 Hybrid G4 Pro (GEN6) version as well?

Thank you in advance.

### Integration Version

2026.05.3

### Homeassistant core version

2026.5.3

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

10K0F1

### Inverter firmware versions

ARM 18.06-0.06, DSP 18.04

### Connection Method

Wifi+LAN dongle

### Dongle firmware

1.002.09

wills106:
Have you looked at the Beta release to try and address the old vs newer firmware versions?

 #2079

gabriel31337:
That was the first thing I have checked. It doesn't seem to enable more sensors, just fix the values of actual sensors between different modbus versions.

rosenrot00:
@gabriel31337 could you tell exactly which ones are missing?

Also please tell your modbus version.

gabriel31337:
How do I know what else is missing? That's why I requested to do a compare, ideally take all the sensors available and try to make them work with GEN6.

I accidentally come across some information in FB groups where people state they have this and that information visible with this integration. I look inside my HA integration, the info is not there. I check the source code, the GEN6 condition is missing. I add it there, it works. So would be nice for the author(s) to check all the available sensors if these can be used with GEN6 as well.

Modbus protocol version 6.

TCWORLD:
For modbus protocol <100 we have no reliable documentation, so the register support is largely guesswork as to whether they exist - there is nothing we can check against. Without having a GEN6 Inverter ourselves, we also can't test things out.

GEN6 register have been added incrementally as based on user feedback as to what exists and what doesn't. In particular for the older protocol version there are definite differences between X3 Hybrid GEN4s and GEN6s so just blindly enabling everything is not sensible and could even cause issues if e.g. we write to an address that does something different from what we expected.

If you have tried additional registers and found them to work, by all means either open a PR, or let us know what you have tested.

[...]

Issue #2079, opened 2026-05-21
2026.05.4 / 2026.05.5 / 2026.05.6 Release - Issue Reporting Thread


TCWORLD:
## Summary

I'm opening this issue to track the work we've been doing to hande differences in Modbus register maps based on inverter firmware version.

Recent versions of SolaX GEN6 firmware upgrades and (potentially GEN4/GEN5 as they are updated) led to incompatabilities in the register maps causing problems.

Thanks to @rosenrot00's for PR #2054 which allows the SolaX plugin to distiguish between the firmware versions based on an entity called "Modbus Protocol Version". Specifically:
 * Inverters with a value <100 (seems to be usually 5, 6 or 7) are the old protocol version where we don't have full documentation.
 * Inverters with a value >=100 are using the new protocol version where we have a largely documented register map.

## Related PRs

The following PRs are related to this register map handling:

- #2054 (base PR, released as 2026.05.4b1)
- #2069
- #2070
- #2072
- #2073
- #2076 
- #2077 
- #2089
- #2093

## Testing

Users with a GEN6 inverter in particular (X3-HYB-G4 PRO, X1-VAST, X3-VAST, etc,) if you are willing to give feedback on the changes that would be great.

If you wish to help us test this release, you'll need to have 2026.05.5 installed.

If you find an issue, please post here, and include:
1. The value of "Modbus Protocol Version" entity (under Diagnostics)
2. Software version number from "SolaX Software Version" entity (under Diagnostics)
3. Your inverter serial number prefix (first 6 digits, e.g. 10K0F1).
4. Details of any registers which have inaccurate or unexpected values (e.g. unknown, incorrect scale factor, always 0, etc.) compared to what they used to be, please let us know so that we can make sure they are fixed.

rosenrot00:
@TCWORLD great, this will help. I created a Github action that would generate pre-releases from an open PR. This would help to test much faster with the users without them editing files manually.

wills106:
Just released [2026.05.4b2 ](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2026.05.4b2) with the latest changes.

rosenrot00:
With the great work of @TCWORLD I think we are ready to leave beta?

Several issues were closed as solved. No flaws on my system. Due to the massive changes I made to introduce the modbus version I thought it to cause more issues.

@TCWORLD what do you think - can @wills106 pull it off?

TCWORLD:
[...]

PR #2099, opened 2026-05-30
Further protocol specific register mapping improvements.

### LINKED ISSUE #2100: [Bug]: SoC reports 16% when Solax Cloud reports 10%
### Describe the bug

Upgraded to 2026.05.5 today and observed some odd SoC behaviour after a discharge (SoC creeping up by 6%). On the final discharge of the day it stopped at 16% and couldn't figure out why it wouldn't go to 10%. Logged into solax cloud and it was reporting SoC of 10%, looking at the cloud SoC where I saw odd SoC behaviour of SoC creeping up, the cloud trace has non of that. Rolled back to 2026.05.03 and SoC dropped back to 10%.

### Integration Version

2026.05.5 

2026.5.4

### Inverter brand

X1-Hybrid-5.0kW by SolaX Power Firmware: DSP 1.58 ARM 1.53 Hardware: Gen4

### Plugin used

plugin_solax.py

### Serial prefix

H4502A

DSP 1.58 ARM 1.53

### Connection Method

Solax pocket wifi 3

### Dongle firmware

1.006.01

### Detailed Error Log

_No response_

### Additional context

Image

### LINKED ISSUE #2102: [Bug]: Battery SoC stuck at 0% - 2026.05.5 & 2026.05.4 (broken by #2069)
### Describe the bug

I updated the integration yesterday and immediately my reported battery SoC went to 0%, when the real value was 96%.
I've removed the inverter, re-downloaded the integration and recreated the inverter, still the same issue.

### Integration Version

2026.05.5, possibly also .04.5 as I skipped a version.

2026.5.4

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H4502A

DSP 1.58 ARM 1.53

### Connection Method

Pocket wifi 2.0, modbus/TCP

### Dongle firmware

-

### Detailed Error Log

The only errors/warnings in the log related to solax are these:

Image

Full log attached.
[home-assistant_2026-05-31T12-53-24.474Z.log](https://github.com/user-attachments/files/28437479/home-assistant_2026-05-31T12-53-24.474Z.log)

### Additional context

Modbus version reports 100.

TCWORLD:
Several improvements to the protocol dependent register mappings:

### Fix GEN6 EPS Yield Total scaling for Proto>=100

Wrong scaling for GEN6 on newer protocol version.

### Tidy Grid Meter/CT Registers

1. There were some duplicate registers (no enabled by default hence didn't show up in testing)
2. The registers were out of order which made it hard to see what was going on.
3. Rename some from Meter to Grid for consistency with the docs and the rest of that register set. 
    As these are disabled by default and were largely added in the most recent release, I doubt the rename will cause problems.
4. Add some of these registers for proto < 100 based on the V3.36 docs.

**NOTE:** For the `grid_reactive_power_l1`/`2`/`3` registers, we should probably test on a protocol 100 inverter and see if they have the correct values. Between the old versions, to V1.00. and then in V1.02 docs these changed address in a frankly strange ways. I wonder if it was actually a documentation error in V1.00 compared with V1.02 rather than registers actually moving, but could be wrong. I've left this discrepancy with gating in for now with protocol gating.

### Enable extra Meter2 registers for Proto < 100

These exist for GEN3, but were disabled for 4/5/6 except in protocol >=101. However these are present in the G4 V3.36 docs, so I suspect as with the other meter 2 registers these should just be always available.

### Exclude GEN4 from battery SoH/SoC total protocol gating

For some reason, even on proto>=100 the GEN4s are reporting incorrect information in the battery totals registers.

Fortunately GEN4 are physically limited to 1 battery, so we can always use the battery base register rather than needing the total registers.

Fixes #2100. Fixes #2102.

### Enable runtime registers for X1

These were added for X3 but not X1. The docs don't explictly list them as X3 only.

### Change Grid Export Total to U32 

According to both old and new docs this should indeed be unsigned. Convert the type and remove remark.

### Re-enable registers for < 100.

(Re-)enable some registers for older versions as they exist in the old V3.36 G4 docs.

1. Pgrid Bias used to be enabled (requested in #2079)
2. VPP Exit Idle Enable used to be enabled
3. Main Breaker Limit now enabled. Add different native ranges for X1 vs X3 per docs.
4. Battery State of Health enabled for GEN6 (requested in #2085)


TCWORLD:
Regarding "Rename some entities from Meter to Grid for consistency" - I suspect based on testing this should have been the other way around, renaming them from Grid to Meter (or Meter 1). 
These registers don't appear to work for the CT case. In fact in the old docs they are labelled simply "Meter", while in the new >=100 docs they are labelled as being Meter/CT, so don't really know the best option there. Happy to change to Meter.

PR #2111, opened 2026-06-05
Respect Home Assistant entity lifecycle for sensor hub references

rosenrot00:
This refactors sensor registration so the hub only stores live sensor entity references after Home Assistant has actually added the entity via async_added_to_hass. Sensor metadata is kept separately in sensorDescriptions for register planning and Energy Dashboard mapping, which avoids keeping entity objects for disabled entities.


wills106:
Which Issue does this PR relate to? Or is it just a general clean up?

Just so I know how to address it in the next release.

rosenrot00:
This solves the issue that was mentioned in the collective thread regarding reloading loops on multiple inverters.

wills106:
Ahh ok.

PR #2122, opened 2026-06-10
Replace time selects with native time entities

### LINKED ISSUE #2119: Display lag
When opening a page with multiple Solax-modbus entities there is lag on mainly my android devices.
First some icons delay then if you scroll on the page there is a delay before it will actually scroll.
Rolled back to 2026.3.4 works perfectly.

HASS: 2026.6.1
SolaX Inverter MODBUS: From 2026.4.on
Inverter: X1-Hybrid Gen4

rosenrot00:
Replace legacy charger/discharger time dropdowns with native Home Assistant TimeEntity entities.

On SolaX Gen4, the device page previously exposed 16 time settings as select dropdowns with 1440 minute options each. That produced 23000 time options just for these controls, and the measured HA device page rendered 23000 dropdown items overall, with about 71k <slot> and 70k <span> elements. This caused noticeable lag when opening the device UI.

This change removes the legacy time selects, reuses the existing Modbus payload mappings for native time.* entities, and adds the missing Modbus read/default handling so migrated time entities can decode raw register values correctly instead of staying unavailable.

@wills106 this is your baby as you recently changed it. This is just a suggestion to fix https://github.com/wills106/homeassistant-solax-modbus/issues/2119

But be aware this changes the affected entities from select.* to time.*. Existing dashboards, automations, scripts, or templates that reference the old select entity IDs will need to be updated to the new time entities.


wills106:
I think this is a bit messy trying to map in selects as time entities.
I'll look at going over the GEN4-6 entities and add them in as per https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2026.04.3
I'll probably simplify the names as well like I did for the GEN3 if it makes sense.
As this will be a breaking change anyway, although I don't know how often people actually change time slots using the drop down selects?

rosenrot00:
@wills106 you are right about that there might be a better way to do it. What do you think about my recent changes? I updated the files from this PR to a cleaner version.

wills106:
Thanks

PR #2123, opened 2026-06-10
Increase Gen6 X1-VAST battery max current from 30A to 50A

rosenrot00:
I know it might be that this depends on the attached battery, but does it hurt if we just set the limit to 50A in general?

Other installations that can't do 50A won't allow it anyway.


TCWORLD:
There's an odd note in the datasheet which could explain why 30A was chosen (or it could have just been chosen from copy and paste of the GEN5 inverters):

> If each of the two battery ports is connected to a separate battery, it's 25A per port. If one port is connected to a single battery, it's 30A. If both ports are connected to a single battery using a 2-in-1 splitter cable(sold separately), it's 50A

But to be honest I would hope the inverter is sensible enough to internally limit to 30A based on how it is configured, so allowing 50A in the integration makes sense.

rosenrot00:
I would say the inverter wouldn't allow values outside its limits. I'm not concerned.

Issue #1815, opened 2026-01-28
[FR]: Multi-Battery String Support for SofarSolar HYD Inverters


paveltresnak:
### Description

# Bug Fix: Multi-Battery String Support for SofarSolar HYD Inverters

## Issue
The integration creates entities for all battery strings but only reads data from Battery 1. All other batteries (2-7) display identical values copied from Battery 1 instead of their actual readings.

## Hardware
- **Inverter:** SofarSolar HYD 10KTL-3P (Serial: SP1ES110N3T121)
- **Batteries:** 7× AMASS GTX3000H (17.5 kWh total)

## Root Cause
Line ~152 in `sensor.py`:
```python
async def readPreparation(old_data, key_prefix=key_prefix, batt_nr=0, batt_pack_nr=batt_pack_nr):
```
The hardcoded `batt_nr=0` causes all batteries to read from battery 0.

## Fix
**Line ~152:** Change `batt_nr=0` to `batt_nr=batt_nr`

**Line ~762:** Add exception for BMS registers (0x9000-0x9FFF) to prevent duplicate register warnings, since these registers are dynamically switched via BMS_Inquire.

## Test Results
✅ All 7 batteries now display unique, correctly updating values  
✅ Each battery shows different cell voltages (3-6mV differences)  
✅ History graphs confirm proper data updates  
✅ No breaking changes to single-battery systems

## Files Modified
- `custom_components/solax_modbus/sensor.py` (~20 lines, 2 changes)

I can provide a detailed pull request with the complete patch if needed.

### Suggested Solution

# Multi-Battery String Support for SofarSolar HYD Inverters

## Summary
Add support for reading all battery strings (not just the first one) on SofarSolar HYD inverters with multiple battery packs. This enhancement enables proper monitoring of all connected battery modules.

## Tested Hardware
- **Inverter:** SofarSolar HYD 10KTL-3P
- **Serial Number:** SP1ES110N3T121
- **Battery System:** 7× AMASS GTX3000H (2.5 kWh each, 17.5 kWh total)
- **Configuration:** 7 battery strings, 1 pack per string, 16 cells per pack

## Problem Description
Currently, the integration only reads data from the first battery string (battery 0), even when multiple battery strings are connected. All battery entities (Battery 2/1 through Battery 7/1) are created but display identical values copied from Battery 1/1, rather than reading their actual cell voltages and status.

### Root Cause
[...]

PR #2132, opened 2026-06-15
Fix SofarSolar multi-battery-string support (#1815)

paveltresnak:
## Problem

Refs #1815. On SofarSolar HYD inverters with a **multi-string** battery stack, only the **first** string (`batt_nr=0`) ever shows up in Home Assistant.

In `sensor.py`, `batt_quantity` (number of strings, read from the device's BaPack-number register) is computed but **never used** — the battery setup loop iterates only over `batt_pack_nr` with `batt_nr` hardcoded to `0`.

## Fix

Scoped to Sofar (this loop runs only when `plugin.BATTERY_CONFIG` is set, which is exclusive to `plugin_sofar.py`). For single-string inverters `batt_quantity == 1`, so behaviour is unchanged.

- Wrap the per-pack loop in an outer `for batt_nr in range(0, batt_quantity or 1, 1)`. The `or 1` guards a `None` read of the BaPack-number register (preserves the previous single-iteration behaviour when the string count is unknown).
- `batt_pack_id` now includes the string index: `f"battery_{batt_nr + 1}_{batt_pack_nr + 1}"`.
- `readPreparation` closure default `batt_nr: int = batt_nr` (was `0`) so the per-iteration string is captured (`readFollowUpBattery` already did this).
- Suppress the `holding register already used` warning for BMS registers (`0x9000-0x9FFF`), which are intentionally re-registered per string via `BMS_Inquire`.

## Testing

Tested on a **HYDxxKTL-3P** inverter with a **7-string** battery stack (AMASS GTX3000H): all 7 strings now appear with correct per-cell voltage/SOC/SOH data. `python -m py_compile` clean. Single-string setups unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/solax_modbus/sensor.py
+++ b/custom_components/solax_modbus/sensor.py
@@ -207,73 +207,79 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, async_add_e
         batt_quantity = await battery_config.get_batt_quantity(hub)
         _LOGGER.info(f"batt_pack_quantity: {batt_pack_quantity}, batt_quantity: {batt_quantity}")
 
-        batt_nr = 0
-        for batt_pack_nr in range(0, batt_pack_quantity, 1):
-            if not await battery_config.select_battery(hub, batt_nr, batt_pack_nr):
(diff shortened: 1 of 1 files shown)


wills106:
Thanks for fixing the CI/CD

Issue #2130, opened 2026-06-14
Feature Request/Bug: Enable Off-Grid / EPS sensors for single-phase hybrid inverters (ZCS HYD 6000HP / Sofar HYD)


vinceg0267:
### Describe the bug / feature request
The off-grid (EPS) output sensors are currently unavailable for single-phase hybrid inverters like the **ZCS Azzurro HYD 6000HP** (which is a rebranded Sofar HYD inverter). 

These sensors are hidden because their `allowedtypes` parameter requires the `X3` flag (three-phase), which is not active on single-phase models. However, single-phase hybrid inverters do have full EPS (Off-Grid) functionality, and these registers are valid for them too.

### Expected behavior
All sensors belonging to the off-grid output group should be visible on single-phase hybrid models when they are in EPS mode.

### Suggested fix
In the sensor definitions file, the `allowedtypes` for the off-grid/EPS sensor entities should be changed from:
`allowedtypes=HYBRID | X3 | EPS,`
to:
`allowedtypes=HYBRID | PV,` 
(or modified so that the `X3` flag is not mandatory for single-phase hybrid units).

### Examples of affected sensors:
- `offgrid_voltage_l1` (Register `0x50A`)
- All other sensors belonging to the **Off-Grid Output** group.

I tested this manually by removing the `X3` flag from `allowedtypes` in the python file, and the sensors immediately became available and read the data perfectly from my ZCS HYD 6000HP.

### Integration Version

2026.06.5

### Homeassistant core version

2026.6.2

### Inverter brand

ZCS

### Plugin used

plugin_sofar.py

### Serial prefix

0

### Connection Method

RS485-USB

wills106:
As you have done most of the work locally, are you able submit your file as a PR?

vinceg0267:
I have submitted the Pull Request here: #2135. It updates the allowedtypes for the Sofar off-grid sensors as discussed.

TCWORLD:
(Edited)

I'm not convinced the change you've made is the correct way to go about this.

The "PV" flag is IIRC intended for inverters which are PV-only. These do not have a battery and therefore the off-grid registers do not apply.

As your inverter is a hybrid, it should have the "HYBRID" flag which is already present. If this is not the case, please let us know as it would imply your inverter has been mis-identified. 


[...]

PR #2135, opened 2026-06-16
Update plugin_sofar.py

vinceg0267:
Fix Off-Grid sensors for single-phase Sofar/ZCS HYD inverters

As requested in issue #2130 , this layout enables off-grid sensors for single-phase hybrid models.


custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1865,7 +1865,7 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
         register=0x50F,
         scale=0.01,
         rounding=2,
-        allowedtypes=HYBRID | X3 | EPS,
+        allowedtypes=HYBRID | PV,
     ),
     SofarModbusSensorEntityDescription(
         name="Off-Grid Voltage L2",
```
wills106, line 1936: Does your Inverter actually return anything for these L2 and L3 sensors? Or is it just unknown / 0.
You are probably best keeping these:
`allowedtypes=HYBRID | X3 | EPS,`

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1765,7 +1765,7 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
         register_data_type=REGISTER_S16,
         scale=0.01,
         rounding=2,
-        allowedtypes=HYBRID | X3 | EPS,
+        allowedtypes=HYBRID | PV,
```
wills106: These should be:
allowedtypes=HYBRID | EPS,

THE PV only Inverters shouldn't have the ability to function when the Grid is down.
Also keeping the EPS flag means that only people who set the EPS flag during setup get to see these Off-Grid / EPS settings. As not everyone has their Inverter and house wired up to use EPS functions.

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1799,7 +1799,7 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
         register=0x507,
         scale=0.01,
         rounding=2,
-        allowedtypes=HYBRID | X3 | EPS,
+        allowedtypes=HYBRID | PV,
     ),
     SofarModbusSensorEntityDescription(
         name="Off-Grid Voltage L1",
```
wills106, line 1815: As you wouldn't normally refer to these as L1 on a single phase normally you are best having a Single and 3Phase option like below:

```
        name="Off-Grid Voltage",
        key="offgrid_voltage",
        native_unit_of_measurement=UnitOfElectricPotential.VOLT,
        device_class=SensorDeviceClass.VOLTAGE,
        register=0x50A,
        scan_group=SCAN_GROUP_FAST,
        scale=0.1,
        rounding=1,
        allowedtypes=HYBRID | X1 | EPS,
    ),
        name="Off-Grid Voltage L1",
        key="offgrid_voltage_l1",
        native_unit_of_measurement=UnitOfElectricPotential.VOLT,
        device_class=SensorDeviceClass.VOLTAGE,
        register=0x50A,
        scan_group=SCAN_GROUP_FAST,
        scale=0.1,
        rounding=1,
        allowedtypes=HYBRID | X3 | EPS,
    ),
```

TCWORLD:
I've added some comments in the linked issue thread about this, not sure the change is correct as-is.

wills106:
> I've added some comments in the linked issue thread about this, not sure the change is correct as-is.

Not sure if you can see my review comments, but I have basically said the same regarding the EPS flag & L1 etc.
Perhaps only @vinceg0267 can see them.

TCWORLD:
@wills106 Ah ok, no I can't see the review comments.

wills106 (changes requested):

wills106:
> @wills106 Ah ok, no I can't see the review comments.

I don't think I finished the process properly. I think they should be showing now.

vinceg0267:
Hi @TCWORLD, @wills106 

Thank you for the detailed explanation! It makes total sense now, and I apologize for changing the whole Off-Grid group. Since I have a single-phase inverter, you are completely right: I don't need L2 and L3 sensors, and I realize now that my bulk modification was incorrect.

To answer your questions:
- My inverter **Serial Number** starts with: **ZM2ES0**
- The inverter does have batteries and the EPS function is physically active on my system, but I now realize that I missed checking the "EPS" option during the initial integration setup in Home Assistant.

I completely agree with your proposal to add specific sensors for single-phase models using `allowedtypes=HYBRID | X1 | EPS` without the "L1" suffix. It sounds like the perfect and cleanest solution for everyone.

wills106:
> Thank you for the detailed explanation! It makes total sense now, and I apologize for changing the whole Off-Grid group. 
>
> I completely agree with your proposal to add specific sensors for single-phase models using `allowedtypes=HYBRID | X1 | EPS` without the "L1" suffix. It sounds like the perfect and cleanest solution for everyone.

Don't worry about it.

Thanks for taking your time to submit your first PR on here.

Edit:
Once you have submitted your next push we can have another look and see if any further refinement is needed or not.

vinceg0267:
Thank you @wills106, I really appreciate your kindness! 

Feel free to close this PR since you and @TCWORLD have the perfect implementation in mind. I will properly enable the EPS option in my Home Assistant setup and look forward to testing your official update as soon as it's released!

wills106:
@vinceg0267 you can edit your local branch to incorporate the suggested changes and then push them to this PR as an update, so it updates what you have already submitted. Then I can review and merge in your PR.

vinceg0267:
Hi @wills106, 

I have updated the file in my fork following the suggestions. I split the off-grid voltage sensor into two separate definitions: one for single-phase (`X1`) without the "L1" suffix, and one for three-phase (`X3`) keeping the "L1" suffix.

The PR should now be updated and ready for your review. I hope everything looks good and the implementation is correct!

vinceg0267:
Hi @wills106, 

I did a thorough review of the code and noticed a few syntax mistakes and duplicate keys in my previous draft. 

I have just pushed a new commit to fix everything:
- Fixed the duplicate keys conflict on `active_power_offgrid_total` and `reactive_power_offgrid_total`.
- Removed the temporary `1pH` suffixes and properly formatted the `X1` entity names and keys according to your guidelines (removing the `_l1` suffix for single-phase without adding `_x1` where not needed).
- Swapped the `PV` flag back to `EPS` as requested.

The PR should now be clean, fully compliant, and ready for your final review. Thank you again for your patience and guidance!

TCWORLD:
@vinceg0267 I don't see any changes yet - you'll need to push your local commits before they appear here.

Edit: seems you pushed to the "main" branch on your fork rather than the one called "patch-1" that this PR uses. You'll need to merge that commit onto the "patch-1" branch for this PR to update.

vinceg0267:
Hi @TCWORLD,

Apologies for that! I had accidentally pushed the changes to my main branch instead of the PR branch. 

I have just successfully pushed the correct code to this PR branch. It now includes the clean split between single-phase (`X1`) and three-phase (`X3`) off-grid sensors, without any of the temporary suffixes and using `HYBRID | X1 | EPS` / `HYBRID | X3 | EPS` accordingly.

Everything should be visible and ready for review now!

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1754,19 +1754,42 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
...
         native_unit_of_measurement=UnitOfPower.KILO_WATT,
         device_class=SensorDeviceClass.POWER,
         state_class=SensorStateClass.MEASUREMENT,
         register=0x504,
-        # newblock = True,
         register_data_type=REGISTER_S16,
         scale=0.01,
         rounding=2,
         allowedtypes=HYBRID | X3 | EPS,
     ),
```
TCWORLD: Where both X1 and X3 share the same register and sensor name, you can just remove the X3 flag and both will use the same definition. No need to add a second one just for the X1.

Same applies to `apparent_power_offgrid_total_x1`, `reactive Power_offgrid_total_x1` and `offgrid_frequency_x1` below, these can also use the same entity for X1/X3 flags.

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1876,7 +1986,7 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
         scan_group=SCAN_GROUP_FAST,
         scale=0.1,
         rounding=1,
-        allowedtypes=HYBRID | X3 | EPS,
+        allowedtypes=HYBRID | PV,
```
TCWORLD: Please change this back to:

```suggestion
        allowedtypes=HYBRID | X3 | EPS,
```

Same for the other X3 specific registers below that were changed.

(Refer to the diff for what changes this PR is proposing: https://github.com/wills106/homeassistant-solax-modbus/pull/2135/files)

TCWORLD (changes requested):
Getting there, but some more tidying up still required.

custom_components/solax_modbus/plugin_sofar.py
TCWORLD: Just to explain this further, we basically try to keep commonality grouped to make maintenance easier, and only when differences arise do we split them up.

When creating the same entity using mutually exclusive flags (e.g. one entity with X3 and another with X1), the key for both matching entities should be the same as they are never loaded together - so if these are left as separate entities, the `_x1` suffix should be removed.

vinceg0267:
Hi @TCWORLD,

Thank you for the guidance! I completely understand your point about keeping the code clean and avoiding unnecessary duplication. 

I have just pushed a new commit to clean everything up based on your suggestions:
- Removed the separate `_x1` clones for the global/total sensors (`Active Power Total`, `Reactive Power Total`, `Apparent Power Total`, and `Off-Grid Frequency`).
- Unified those global entities into single definitions using `allowedtypes=HYBRID | EPS`.
- Kept the split layout only for the phase-specific sensors where names actually differ between `X1` and `X3`.
- Reverted the `L2/L3` specific registers back to `HYBRID | X3 | EPS`.

The file should now be much cleaner and fully compliant with your standard!
Please let me know if everything is okay now!

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1866,7 +1931,7 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
         scale=0.01,
         rounding=2,
         allowedtypes=HYBRID | X3 | EPS,
-    ),
+    ),
```
TCWORLD:
```suggestion
    ),
```

custom_components/solax_modbus/plugin_sofar.py
```diff
@@ -1812,6 +1822,17 @@ def value_function_sync_rtc_ymd_sofar(initval: Any, descr: Any, datadict: dict[s
...
+        name="Off-Grid Current Output",
+        key="offgrid_current_output",
+        native_unit_of_measurement=UnitOfElectricCurrent.AMPERE,
+        device_class=SensorDeviceClass.CURRENT,
+        register=0x50B,
+        register_data_type=REGISTER_S16,
+        scale=0.01,
+        rounding=2,
+        allowedtypes=HYBRID | X1 | EPS,
+    ),
```
TCWORLD:
```suggestion
    ),
```

TCWORLD (approved):
Whitespace formatting to fix CI failure. There should be an option to commit these suggestions.

wills106:
Thanks @vinceg0267 for working on this.

vinceg0267:
Hi @wills106 Thanks to you and @TCWORLD for making this possible and for your much-appreciated support.
I will try, with your permission, to collaborate to improve, if possible, the plugin_sofar for single-phase inverters.

wills106:
Further improvements are always appreciated.

PR #2138, opened 2026-06-16
Revamp "Mode 8 Export First" control loop filtering

TCWORLD:
Following on from the redevelopment of the Negative Injection control loop, this PR updates the Export-First control loop to use a similar filtered setpoint approach.

The intended functionality of the mode should be unchanged, however the aim here is to make the response more stable by filtering the battery control setpoint to reduce oscilatory behaviour. Rather than trying to select between large/small step changes, we use the `autorepeat_setpoint_filter` helper to select the new control output value using an exponential weighted filter.

The logic also reintroduces the inverter limit to the control loop in a way which is hopefully stable unlike the previous attempt. This is designed to ensure that the control loop understands that it can't export more than the inverter power rating to the combined grid and home, and so allow it to optimise feeding extra energy to the battery when it is above what can be exported or used by the home.


TCWORLD:
@rosenrot00 Would you be able to test this out when you get a chance. 
I've been running it for a couple of days and seems to be giving a reasonable response.

rosenrot00:
@TCWORLD happy to test. One remark without having looked in detail. My inverter seems to go beyond its 8kW rating at times. Up to 9kW (house+export+battery).

TCWORLD:
@rosenrot00 Seems like some of the inverters have a 110% output rating, so 8kW giving 8.8kW would make sense. 

To be honest that shouldn't matter too much in this implementation, it just means the excess would go into the battery sooner if the target SoC has not yet been reached, but shouldn't actually limit the overall generation. 

The aim is to stop the inverter internally clipping the PV - for example if you had an 8.8kW real output limit, but 10kW of PV, without checking the limit, it may end up setting the battery charge to 0kW, at which point the inverter would have to clip the PV down to 8.8kW to respect its limits. Whereas with the limit check it could be putting 8kW to the grid/home and 2kW to the battery and allow the full 10kW of PV.

We can probably add the 110% factor, either as an entity or an override option.

I guess the only possible issue would be if you were drawing 8.8kW of house load, at which point it would start importing 800W while charging the battery, so that might need investigation.

rosenrot00:
Today was cloudy, I need a bit more time to test.

The issue with the remote control is the limited information the inverter provides. I'm still wondering why they don't just provide one setting called "Direct Control" or whatever, in which settings are not stored to the eeprom. No remote controls would be needed, one could just change all the registers on the fly without any issue.

With the current remote control we struggle even with understanding if the inverter is throttling its PV output.

rosenrot00:
Seems to work. I can't tell if it works better or how it tackles special cases however.

TCWORLD:
I've had it running the last couple days and seems to work nicely. Going to switch back to the current release version for tomorrow and compare the new with old to see if that is enlightening.

TCWORLD:
Some screenshots for comparison, of original and new filtering.

The conditions for each are obviously not the same as solar is variable, but I've tried to screenshot sections of similar conditions to try and compare tem.

I think overall the new one is a more damped as expected by the filtering. Whether it is better or not is hard to tell.
The advantage is that the logic is more closely matched with the NI mode so might be easier to maintain. I'm intending to try applying the same to the No-Discharge mode as well.

### Load step change
 
Original:

image

New:
 
image

### Export Limit – charges battery with excess
  
Original:

image

New:

image

### Low-level deficit response

Note: these two are not really comparable as different conditions. I don't have an example of the old one under low PV and deficit conditions.
  
Original:

image

New:

image

PR #2139, opened 2026-06-16
feat(growatt): add SPA (RH/WPD) support with strict entity matching

danmiles86:
his PR adds proper support for Growatt SPA series inverters (e.g. SPA3000).

### Changes
- Added `SPA` bitmask (0x800000) and `ALL_SPA_GROUP`
- Added RH/WPD series detection in both primary and fallback (addr 9) paths
- Made `spmatch` strict so SPA inverters only receive entities that explicitly include the SPA flag
- Updated 70 registers (validated against real SPA3000 test data) to `GEN3 | SPA`

### Testing
- Tested on real SPA3000 unit
- All 6 time slots (1–6) for Grid/Battery/Load First now appear as config entities
- Battery SOC, priority, rates, and energy counters all working correctly

This resolves the "unknown error" when adding SPA inverters and ensures only relevant entities are created.


TCWORLD:
There already appears to be an entry for the SPA3000, with a prefix check for RH1 - not sure why that doesn't match for you. 

There shouldn't be a need for a specific SPA flag - there is already an AC flag which indicates a grid connected battery (without PV), a flag used on the existing SPA3000 check. So I think the first thing to resolve is why that existing check doesn't match for your inverter.

PR #2160, opened 2026-07-03
fix: standardize entity and device naming across all platforms

Bl00d-B0b:
## Problem

Entity naming is built three different ways depending on the platform, so the same integration produces inconsistent friendly names and entity_ids — and with an inverter name suffix configured, doubled words:

| Platform | Current convention | `has_entity_name` |
|---|---|---|
| sensor | `"{hub_name} {description name}"` (legacy composition, no suffix) | ❌ |
| select, number, button, time | `"{hub_name} {suffix} {description name}"` | ❌ |
| switch | `"{suffix} {description name}"` — hub missing entirely | ❌ |
| device name | `plugin_name + suffix` | — |

### Real-world example (hub name `EV`, suffix `Charger`, X3 EVC device)

| Layer | Current | Problem |
|---|---|---|
| Device name | `SolaX EV Charger Charger` | plugin_name already ends with "Charger", suffix appends it again |
| Switch friendly name | `Charger Electronic Lock` | hub name missing — and its entity_id (`switch.charger_electronic_lock`) is not even scoped to the hub |
| Number friendly name | `EV Charger Overload Limit` | looks right **only by accident** — "EV" + "Charger" happens to spell the device name; any other hub-name/suffix combination renders doubled or mismatched |
| Two hubs, same plugin | `SolaX Inverter` + `SolaX Inverter` | device names collide, users must rename manually |

## Change

One convention, applied identically to all six platforms:

```
device name   = hub name + " " + suffix        →  "EV Charger"
entity name   = description name               →  "Electronic Lock"
                (fallback: entity key, on all six platforms)
friendly name = composed by HA (has_entity_name=True)
              →  "EV Charger Electronic Lock"
```

- `_attr_has_entity_name = True` on sensor, select, switch, number, button, time
- every platform exposes the **same** `name` property: `str(description name or key)` — no hub-name composition, no setup-time suffix injection (`replace(descr, name=suffix + name)` removed everywhere)
- device name = `hub_name + suffix` instead of `plugin_name + suffix` — the suffix now exists in exactly **one** place, and device names are unique per config entry
- **battery pack sub-devices follow the same rule**: device = `"{hub} Battery {n}/{m}"`, entity names no longer carry the pack prefix (keys keep it — identity unchanged). Previously the pack identity lived in both the device name and the entity name, which would have doubled under `has_entity_name`. `battery_sensor_name_prefix` stays in `base_battery_config` for API compatibility but is no longer applied to names.

### Before/after (hub `EV`, suffix `Charger`)

| | Before | After |
|---|---|---|
| Device | SolaX EV Charger **Charger** | EV Charger |
| Switch | Charger Electronic Lock (no hub) | EV Charger Electronic Lock |
| Sensor | EV Firmware Version (no suffix) | EV Charger Firmware Version |
| Select | EV Charger ECO Gear | EV Charger ECO Gear (unchanged) |

## Compatibility

- **`unique_id` untouched** (`{hub_name}_{key}`, byte-identical on all six platforms before and after) — no entity identity changes, nothing breaks
- Existing **entity_ids and history are preserved**: HA resolves stored entity_ids by unique_id before ever invoking its generator, so the new naming applies to friendly names immediately and to entity_ids only for newly created entities (or via HA's "Recreate entity IDs" dialog, opt-in)
- Entities group properly under their device in the UI (standard `has_entity_name` behaviour per HA developer docs)
- Applies uniformly to all 17 brand plugins — naming lives entirely in the platform files
- None-name handling strictly improves: previously a nameless description rendered the literal string `"None"` inside composed names; now it falls back to the entity key (generalizing the pattern upstream already used for its communication sensors). A repo-wide scan confirms zero non-internal descriptions currently lack a name, so this is defensive only.

## Testing

- Verified live on three hubs (X3-Hybrid-G4 15kW, X3-MIC/PRO-G2, X3 EVC 11kW): friendly names consistent across all six platforms, no doubled words, device names unique
- Adversarial review pass performed; both findings (battery pack doubling, None-name fallback) fixed — see review comment below
- Automated uniformity check: `has_entity_name`, name rule, unique_id rule, no suffix injection, no manual entity_id — identical across all six platform files
- `test_device_info_regressions.py` updated for the renamed device-name variable (same structural guarantee, both init paths verified identical)

## Entity IDs & areas (clarification)

This PR never sets `entity_id` — HA core generates it. The PR only supplies the two clean inputs (device name, entity name); HA composes:

```
entity_id (first registration) = slug(device name) + "_" + slug(entity name)
                               = switch.ev_charger_electronic_lock
```

- **No area is involved** in normal registration — a device's area assignment does **not** change entity IDs (verified: assigning an area leaves every existing ID untouched).
- The area only ever enters an entity_id if a user explicitly clicks HA's **"Recreate entity IDs"** dialog while the device is in an area — that is HA core behaviour and is outside this integration's control.
- Because `unique_id` is unchanged, existing installs keep their current entity IDs regardless; only the friendly names update on upgrade.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -675,14 +675,16 @@ class SolaXModbusHub:
             return
 
         # Prepare device_info (inverter detected during initial window)
-        plugin_name = self.plugin.plugin_name
+        # Device name = hub name + optional suffix (e.g. "EV" + "Charger" -> "EV Charger").
+        # Unique per config entry; entity names never repeat it, HA composes the friendly name.
+        device_name = self._name
         if self.inverterNameSuffix is not None and self.inverterNameSuffix != "":
-            plugin_name = plugin_name + " " + self.inverterNameSuffix
+            device_name = device_name + " " + self.inverterNameSuffix
         self.device_info = DeviceInfo(
             identifiers=cast(set[tuple[str, str]], {(DOMAIN, self._name, INVERTER_IDENT)}),
             manufacturer=self.plugin.plugin_manufacturer,
             model=self._get_inverter_model(),
-            name=plugin_name,
+            name=device_name,
             serial_number=self.seriesnumber,
             sw_version=self.plugin.getSoftwareVersion(self.data),
             hw_version=self.plugin.getHardwareVersion(self.data),
@@ -732,14 +734,14 @@ class SolaXModbusHub:
                     self._invertertype = inv
                     _LOGGER.debug(f"{self._name}: inverter detected during deferred setup (type={inv}) – forwarding platforms")
                     # Prepare/refresh device_info in case it wasn't set
-                    plugin_name = self.plugin.plugin_name
+                    device_name = self._name
                     if self.inverterNameSuffix:
-                        plugin_name = plugin_name + " " + self.inverterNameSuffix
+                        device_name = device_name + " " + self.inverterNameSuffix
                     self.device_info = DeviceInfo(
                         identifiers=cast(set[tuple[str, str]], {(DOMAIN, self._name, INVERTER_IDENT)}),
                         manufacturer=self.plugin.plugin_manufacturer,
                         model=self._get_inverter_model(),
-                        name=plugin_name,
+                        name=device_name,
                         serial_number=self.seriesnumber,
                         sw_version=self.plugin.getSoftwareVersion(self.data),
                         hw_version=self.plugin.getHardwareVersion(self.data),
--- a/custom_components/solax_modbus/button.py
+++ b/custom_components/solax_modbus/button.py
@@ -1,5 +1,4 @@
 import logging
-from dataclasses import replace
 from time import time
 from typing import Any
 
@@ -36,17 +35,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, async_add_e
     hub = hass.data[DOMAIN][hub_name]["hub"]
 
     plugin = hub.plugin
-    inverter_name_suffix = ""
-    if hub.inverterNameSuffix is not None and hub.inverterNameSuffix != "":
-        inverter_name_suffix = hub.inverterNameSuffix + " "
-
     entities = []
     for button_info in plugin.BUTTON_TYPES:
         if plugin.matchInverterWithMask(
             hub._invertertype, button_info.allowedtypes, hub.seriesnumber, button_info.blacklist
         ) and matches_modbus_protocol(hub, button_info):
-            if not (button_info.name.startswith(inverter_name_suffix)):
-                button_info = replace(button_info, name=inverter_name_suffix + button_info.name)
             button = SolaXModbusButton(hub_name, hub, modbus_addr, hub.device_info, button_info)
             entities.append(button)
             if button_info.key == plugin.wakeupButton():
@@ -80,6 +73,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, async_add_e
 class SolaXModbusButton(ButtonEntity):
     """Representation of an SolaX Modbus button."""
 
+    _attr_has_entity_name = True
+
     def __init__(
         self,
         platform_name: str,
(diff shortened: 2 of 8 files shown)


Bl00d-B0b:
**Review finding (now fixed in the squashed commit): battery pack naming had the same doubling bug.**

An adversarial review pass caught that battery pack entities (`read_battery`, e.g. Sofar) reproduced exactly the bug this PR fixes for the inverter suffix: the pack identity lived in **both** the battery sub-device name (`"SOFAR Battery 1/1"`) and the entity name prefix (`"Battery 1/1 SOC"`), so with `has_entity_name` the friendly name would have rendered as *"SOFAR Battery 1/1 Battery 1/1 SOC"*.

Fix applied, following the same rule as everywhere else in this PR — context lives in the device name only:

- battery sub-device name is now `"{hub} Battery {n}/{m}"` (hub-based → unique per config entry, consistent with the inverter device naming)
- entity **names** no longer get the `"Battery {n}/{m} "` prefix; entity **keys** keep it, so `unique_id`s are unchanged
- `battery_sensor_name_prefix` remains in `base_battery_config` for API compatibility but is no longer applied to names

Also hardened `button`/`select` name properties to fall back to the entity key instead of rendering a literal `"None"` when a description has no name.

Bl00d-B0b:
## Naming & identity reference — before/after formulas

For reviewers: a complete map of how every name/ID is generated, what this PR changes, and what it deliberately leaves alone. Example inputs used throughout: hub name = `EV`, inverter name suffix = `Charger`, description name = `X`.

### Entity name (what the entity calls itself)

| Platform | BEFORE | Result | AFTER | Result |
|---|---|---|---|---|
| sensor | `hub + " " + desc` | `EV X` | `desc` (fallback: key) | `X` |
| select | `hub + " " + suffix + " " + desc` | `EV Charger X` | `desc` (fallback: key) | `X` |
| switch | `suffix + " " + desc` ⚠️ no hub | `Charger X` | `desc` (fallback: key) | `X` |
| number / button / time | `hub + " " + suffix + " " + desc` | `EV Charger X` | `desc` (fallback: key) | `X` |

Before: three different shapes in one integration — sensor dropped the suffix, switch dropped the hub, the rest had both. After: **all six platforms expose the identical `name` property** — `str(description name or key)`.

### Friendly name

| | BEFORE | AFTER |
|---|---|---|
| Formula | `= entity name` verbatim (`has_entity_name` was False everywhere) | `device name + " " + entity name`, composed by HA core (`has_entity_name = True`) |
| switch example | `Charger X` ⚠️ | `EV Charger X` |

### Device name

| Device | BEFORE | AFTER |
|---|---|---|
| Main | `plugin_name + " " + suffix` → `SolaX EV Charger Charger` | `hub + " " + suffix` → `EV Charger` |
| Battery pack | `plugin_name + " Battery n/m"` (collides across hubs) | `hub + " Battery n/m"` (unique per entry) |

### entity_id — no code changed, only the ingredients

This integration never sets `entity_id` (only commented-out remnants exist in all six platforms); HA core composes it at **first registration**:

```
has_entity_name = False  →  entity_id = slug(entity name)
has_entity_name = True   →  entity_id = slug(device name + " " + entity name)
```

| Platform | BEFORE | AFTER (new registrations only) |
|---|---|---|
| sensor | `sensor.ev_x` | `sensor.ev_charger_x` |
| number | `number.ev_charger_x` | `number.ev_charger_x` (same) |
| switch | `switch.charger_x` ⚠️ not scoped to hub | `switch.ev_charger_x` |

### Unchanged — the compatibility guarantee

| Artifact | Formula (identical before/after, all 6 platforms) |
|---|---|
| `unique_id` | `hub + "_" + key` → `EV_x` |
| Device identifiers | `(solax_modbus, hub, "inverter")` |

Because `unique_id` is untouched, **every existing entity keeps its registered entity_id** — HA resolves stored IDs by unique_id before ever invoking the generator. Existing users see improved friendly names immediately; entity_ids change only for newly created entities or if a user explicitly runs HA's "Recreate entity IDs" dialog. No automations or dashboards break on upgrade.

### One-sentence summary

Before: each platform glued hub/suffix/desc into one string its own way, and friendly names / entity_ids inherited that inconsistency. After: the plugin provides exactly two clean inputs — device name (`hub + suffix`) and entity name (description, key fallback) — and HA core composes everything downstream the documented way.

wills106:
As long as this doesn't break automations, which I don't think it will. I think it's a good idea.

PR #2161, opened 2026-07-03
feat: display all number entities as input boxes by default

Bl00d-B0b:
## Problem

Number entities render as **sliders** by default, which is a poor fit for this integration: inverter and EV charger numbers are precise configuration values — charge current limits, SOC thresholds, export power limits — often with wide ranges (e.g. 0–30000 Wh, 0–6000 W). On a slider:

- exact values are hard to hit (try setting 4600 W on a 0–6000 W slider),
- a stray click/drag silently writes a wrong value to the inverter,
- mobile UX is worse still.

Box mode was already being adopted entity-by-entity — 10 scattered `display_as_box=True` overrides across `plugin_solax.py` and `plugin_solax_ev_charger.py` — while every other number stayed a slider. The result is inconsistent: `Adjust Current` renders as a box while `Export Control User Limit` next to it is a slider.

## Change

- `display_as_box` on `BaseModbusNumberEntityDescription` now defaults to **`True`** → every number entity in **all plugins** renders as an input box
- Removed all 10 per-entity `display_as_box=True` overrides — now redundant
- Verified no plugin forces a mode anywhere: no `display_as_box=False`, no direct `NumberMode` usage remains in any of the 17 plugin files

The escape hatch is preserved: an individual entity can still opt back into a slider with `display_as_box=False` (falls through to HA's default mode).

## Scope

| | Before | After |
|---|---|---|
| Default mode | slider | **input box** |
| Per-entity overrides | 10 × `display_as_box=True` | none |
| Opt-out | n/a | `display_as_box=False` |
| Affected plugins | — | all (mechanism lives in `const.py` + `number.py`) |

Diff is intentionally tiny: 1 line changed in `const.py`, 10 lines removed from plugins.

### DIFF
--- a/custom_components/solax_modbus/const.py
+++ b/custom_components/solax_modbus/const.py
@@ -354,7 +354,7 @@ class BaseModbusNumberEntityDescription(NumberEntityDescription):
     # update only when read value changes
     sensor_key: str | None = None  # only specify this if corresponding sensor has a different key name
     depends_on: list[str] | None = None  # list of modbus register keys that must be read
-    display_as_box: bool = False  # if true, displays the entity as a box rather than a slider.
+    display_as_box: bool = True  # display numbers as an input box (default); set False for a slider.
     suggested_display_precision: int | None = None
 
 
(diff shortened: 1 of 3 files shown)


wills106:
I think this is why some of the export limits got capped, as the sliders got out of hand with +/- 60,000w
Maybe sliders were the only option, and box input was added latter to HA, or I didn't know how to set them? Can't remember.

wills106:
I also know some users don't like the sliders on mobile devices, as you can accidentally change a value too easy just from scrolling.

PR #2162, opened 2026-07-03
feat(evcharger): convert binary SELECT entities to SWITCH

Bl00d-B0b:
## Problem

The EV charger exposes boolean device states as 2-option **select dropdowns**:

- `Device Lock` (0x615): *Unlock / Lock*
- `RFID Card Activation` (0x616): *Disabled / Enabled*

A dropdown is the wrong entity type for a boolean. Compared to a switch it:

- **takes two interactions** (open dropdown, pick option) instead of one tap,
- **wastes UI space** — a select renders as a full-width dropdown row, a switch is a compact toggle,
- **reads worse at a glance** — a toggle's on/off position is instantly visible; a dropdown shows a word you have to parse,
- **integrates worse with HA** — switches work directly with `switch.turn_on/off` services, Google/Alexa exposure, dashboards' toggle cards and quick-action buttons; selects need `select.select_option` with a magic string,
- **is harder in automations** — `state: 'on'` vs matching the exact option string ("Enabled"? "Enable"? "On"?), which differs per entity.

## Change

| Before | After |
|---|---|
| `select.device_lock` — Unlock / Lock | `switch.electronic_lock` — toggle |
| `select.rfid_program` — Disabled / Enabled | `switch.rfid_card_activation` — toggle |
| — (not exposed) | `switch.main_breaker_limit_switch` — **new**, GEN2, register 0x664 |

- Uses the existing `BaseModbusSwitchEntityDescription` infrastructure (register + `sensor_key` + `value_function`) — no platform changes needed
- Adds a generic `value_function_enable_disable` helper to `const.py` (write 1/0), reusable by any plugin that wants the same conversion
- `Main Breaker Limit Switch` (0x664) has inverted register polarity (0 = Closed/Active, 1 = Open/Inactive); a local inverted value function + inverted sensor scale make the switch read naturally: **on = limit active**
- Backing internal sensors deliver 0/1 ints so switch state tracks the actual register on every poll — external changes (via the charger's app/display) are reflected in HA

## Naming

Keys renamed to match switch semantics (`device_lock` → `electronic_lock` per SolaX's own terminology, `rfid_program` → `rfid_card_activation`). The entities change domain anyway (`select.*` → `switch.*`), so there is no unique_id continuity to preserve.

## Testing

Verified on a real **X3 EVC 11kW (GEN2)**: all three switches track register state correctly (including changes made from the charger side) and writes take effect. This pattern could later be applied to the ~20 binary Enabled/Disabled selects in `plugin_solax.py` (MPPT, Battery Heating, HotStandBy, …) — kept out of scope here to keep the change reviewable and because inverter users' automations reference the existing select entities.

### DIFF
--- a/custom_components/solax_modbus/const.py
+++ b/custom_components/solax_modbus/const.py
@@ -483,6 +483,11 @@ def value_function_disabled_enabled(initval: Any, descr: Any, datadict: dict[str
     return scale.get(initval, str(initval) + " Unknown Status")
 
 
+def value_function_enable_disable(bit: int | None, state: bool | None, sensor_key: str | None, datadict: dict[str, Any]) -> int:
+    """Switch value function: write 1 for on, 0 for off (whole-register enable/disable)."""
+    return 1 if state else 0
+
+
 def value_function_gain_offset(initval: Any, descr: Any, datadict: dict[str, Any]) -> float:
     """Apply gain and offset calibration to power measurement."""
     # Simple offset (unit) and gain (%) calibration of the measured power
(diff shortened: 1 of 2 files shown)


Bl00d-B0b:
Closing — consolidating my local work; this conversion may return later as part of a broader binary-select cleanup.

PR #2163, opened 2026-07-03
Respect battery minimum SoC in SolaX remote control

### LINKED ISSUE #2148: [Bug]: Battery Min SOC is not respected when in Enabled Feedin Priority
### Describe the bug

Greetings,

As part of my Home Assistant automation, I set `Remotecontrol Power Control (mode 1)` to `Enabled Feed-in Priority` in the mornings and evenings to ensure excess energy is exported to the grid.

However, when this mode is enabled and household consumption exceeds PV generation, the inverter drains the battery below the configured Min SOC value. In some cases, it has even discharged the battery below 10%, which could negatively impact long-term battery health.
Once the mode is switched back to `Disabled`, the battery behavior returns to normal.

Is this the expected behavior of `Enabled Feed-in Priority`, or is there an additional setting required to prevent the battery from discharging below Min SOC?

### Integration Version

2026.06.5

2026.6.4

### Inverter brand

Solax X3-Hybrid-10kW Gen4

### Plugin used

plugin_solax.py

### Serial prefix

H34A10

DSP 1.58 ARM 1.56

### Connection Method

WiFi 3.0+

### Dongle firmware

3.041.04

### Detailed Error Log

Image
The mode was enabled at 5:40, and battery drain can be seen.

### Additional context

_No response_

### LINKED ISSUE #2152: [Bug]: Enabled no discharge mode sometimes switches off PV generation
### Describe the bug

I'm using solax-modbus with a X1 Hybrid G4 and Predbat in HA. Predbat uses the 'Enabled no discharge' mode for a charge hold - allow the battery to charge, but not to discharge.
Sometimes, this works fine. Other times, typically with the battery at 97-98%, it causes PV generation to pretty much stop, with the PV power down to <20W vs 2-6kW seconds earlier. The battery might be full (not quite), but the PV should be going to the house and export, not stopped. If I override the mode to something like 'Enabled self-use', PV power returns to normal and there's even a trickle going into to the battery (40-150W). If I don't change the mode, it stays with near-zero PV forever.

When I change the mode, predbat will then put it back to 'no discharge' a few minutes later and it's fine for a while but the problem can return after 20 minutes, sometimes longer.

I've not yet found a clear pattern other than the battery being almost full and the mode being 'Enabled no discharge'. I've just enabled debug logging for the integration, so hopefully will get some more clues soon. If there's anything else I can check, please let me know!

Also interesting, 'no discharge' sometimes shows the battery discharging nonetheless. Only small amounts, typically around 140W, so possibly the inverter balancing but it's not really respecting the definition of 'no discharge'.

Working normally:

Image

No PV:
<< screenshot of parameters to follow when it next fails! >>

### Integration Version

2026.06.7

2026.6.1 (container)

### Inverter brand

Solax

### Plugin used

plugin_solax.py

### Serial prefix

H4502A

DSP 1.58 ARM 1.53

### Connection Method

Pocket 3.0 / ethernet

### Dongle firmware

1.006.01  (Communication Module Version, from solaxcloud.com)

### Detailed Error Log

_No response_

### Additional context

_No response_

TCWORLD:
The remote control modes on some (if not all) SolaX inverters don't respect the minimum SoC for the current work mode. For example if the self-use minimum SoC is set to 20%, the remote control mode appears to simply bypass this limit allowing further discharge of the battery.

To combat this, we now perform bounds checking on the push requests such that we do not request a battery discharge if the current capacity has reached the minimum SoC. For Mode 1 we use the self-use minimum SoC sensor for this limit. For Mode 8, we add a new minimum SoC entity to complement the existing mode 8 target SoC entity.

Fixes #2148.

---

Furthermore, add an additional headroom to the `Mode 1 Enabled No Discharge` mode when the battery is full to prevent PV output being clamped to 0. We had a similar issue in the Mode 8 loop which uses a far more complex solution. For the Mode 1 case we'll use a simple extra allowance as the full control loop would be hard to implement with parallel inverters. This should fix #2152.


TCWORLD:
I've tested this so far in Mode 8 PV/BAT control and it works as expected - requested power drops to zero once the minimum SoC sensor is reached.

image

Testing in Mode 1 is still required.

TCWORLD:
Drafting this for now as I <strike> messed something up when resolving the conflict. Also</strike> still need to test mode 1.
(Fixed the merge issue).

TCWORLD:
Mode 1 limiting works as expected:

image

PR #2167, opened 2026-07-05
Add missing MPPT3 flag to EGR and JQG prefix inverters

### LINKED ISSUE #2165: [Bug]: Growatt - MOD 15KTL3-HU Third MPPT Tracker / String is not available as sensor
### Describe the bug

Added the SolaX integration to HA and configured it according to the documentation. Everything works like a charm (read & write to/from the inverter) except for one thing: The values "PV Power 1" and "PV Power 2" are shown, but not the value/entity for "PV Power 3", also not under the disabled entities. My inverter (Growatt MOD 15KTL3-HU) has 3 seperate MPPT tracker, therefore it would be great if I could somehow get those values as well.

### Integration Version

2026.06.7

2026.6.4

### Inverter brand

Growatt

### Plugin used

plugin_growatt.py

### Serial prefix

000000

### Connection Method

Modbus TCP via WiFi Datalogger / Stick

### Detailed Error Log

_No response_

### Additional context

_No response_

TCWORLD:
They are labelled in the comments as 3x MPPT, however were missing the flag to actually enable sensors for the 3rd solar input.

Fixes #2165.

### DIFF
--- a/custom_components/solax_modbus/plugin_growatt.py
+++ b/custom_components/solax_modbus/plugin_growatt.py
@@ -9663,7 +9663,7 @@ class growatt_plugin(plugin_base):
         elif seriesnumber.startswith("SMN"):
             invertertype = HYBRID | GEN4 | X1 | MPPT4  # MIN TL-XH-US Hybrid, 4 MPPT
         elif seriesnumber.startswith("JGQ"):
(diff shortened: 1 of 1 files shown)


pdomcke:
Amazing, works now. Thanks for the super fast fix

PR #2175, opened 2026-07-09
Add scale parameter to PeakShaving Charge Limit

empx92:
Solax X3 G4 Pro:
If entering 4000W register writes 40000 in Peakshaving Charge Limit, so there should be a scale of "10" set.


TCWORLD:
Please can you confirm the modbus protocol version number? 

We shouldn't necessarily make this change without further investigation, as based on the protocol documents this scaling should not be required.

Furthermore when scaling these number parameters you also need to find the matching sensor entity used for readback.

empx92:
> Please can you confirm the modbus protocol version number?
> 
> We shouldn't necessarily make this change without further investigation, as based on the protocol documents this scaling should not be required.
> 
> Furthermore when scaling these number parameters you also need to find the matching sensor entity used for readback.

When editing the Code my inverter was:
X3-HYB-G4 Pro Module IC 1.011.01 / ARM IC 022.06-0
Communication WiFi+LAN but only using LAN


Now i updated everything to the newest Inverter Firmware
Module IC 1.011.01
ARM 023.05-000.06
DSP IC 023.05

without scale=10 there is always too high number written to the system, as said if you type 4000 the inverter saves 40000

PR #2177, opened 2026-07-09
feat: convert binary Enabled/Disabled selects to switches (SolaX inverter + EV charger)

Bl00d-B0b:
## Problem

Boolean device settings are exposed as **2-option select dropdowns** — `Enabled/Disabled`, `On/Off`, `Unlock/Lock`. A dropdown is the wrong control for a boolean:

- two interactions (open dropdown, pick option) vs one tap
- a full-width dropdown row vs a compact toggle
- automations must match a magic option string (`"Enabled"` vs `"Enable"` vs `"On"`, varies per entity) instead of `state: 'on'`
- worse integration with HA services (`switch.turn_on/off`), Google/Alexa exposure, and dashboard toggle cards

The EV charger switch idea (previously #2162) is here generalised to the whole SolaX inverter.

## Change

Converts **all 27 binary Enabled/Disabled/On-Off/Lock selects** to switch entities: 25 in `plugin_solax.py` + 2 in `plugin_solax_ev_charger.py`. Multi-option selects (Charger Use Mode, Shadow Fix levels, Manual Mode Select, …) are left untouched.

Per entity:
- the select becomes a `SolaXModbusSwitchEntityDescription` on the **same write register**, with a `value_function` matching the register's write polarity — `value_function_enable_disable`, or `value_function_disable_enable` for inverted-polarity registers (e.g. HotStandBy, whose `0 = Enabled`)
- the switch reads its state via `sensor_key` from the entity's **existing internal readback sensor**, which now yields an int `0/1` instead of a display string (read polarity derived independently — on the inverter the readback register differs from the write register)
- shared `value_function_enable_disable` / `value_function_disable_enable` helpers added to `const.py`
- **entity keys are unchanged**; only the domain changes (`select.x` → `switch.x`)

### Example (HotStandBy — inverted polarity, handled correctly)
| | write reg 0x99 | readback reg 0xA1 |
|---|---|---|
| Select (before) | `{0: Enabled, 1: Disabled}` | string via readback |
| Switch (after) | on → writes `0` (`disable_enable`) | `scale={0: 1, 1: 0}` → int |

## ⚠️ Breaking change

`select.<hub>_mppt` → `switch.<hub>_mppt`, etc. Automations, scripts and dashboards referencing the old `select.*` entities must be updated to the new `switch.*` entities, and use `state: 'on'` / `switch.turn_on` instead of `select.select_option`. **No registry migration is included** — left for the maintainer to decide whether to add `async_migrate_entry` or ship as a documented breaking change.

## Testing

Live-verified on **X3-Hybrid-G4 15kW + X3 EVC 11kW**: 23 of the 27 converted entities exist on this hardware and report correct on/off state (inverted ones included). The remaining 4 are `allowedtypes`-filtered for other generations and are produced by the same mechanical transform. Compile + ruff + device_info regression tests pass.

Scope is deliberately limited to the two **SolaX** plugins (hardware-testable). The other brands' plugins (Growatt, Solis, Sofar, AlphaESS) have their own binary selects but are out of scope here — no hardware to verify.

## Cross-checked against official SolaX Modbus protocol documents

**Hybrid X1/X3 ESS Modbus RTU V1.02** — 10 of the converted register pairs verified directly; write/readback addresses and `0: Disable / 1: Enable` polarity all match this PR:

| Register (write/read) | Official variable | HA entity name (switch) | Enum |
|---|---|---|---|
| 0x006C / 0x009B | SetChrg&DischrgPeriod2Enable | Charge and Discharge Period2 | 0: Disable, 1: Enable |
| 0x0062 / 0x0093 | SelfUseChargeFromGridEnable | Selfuse Night Charge | 0/1 |
| 0x009A / 0x00A2 | BatExtendFunction | Extend BMS Setting | 0: Disable, 1: Enable |
| 0x00CF / 0x00A3 | BatHeatingEn | Battery Heating | 0: Disable, 1: Enable |
| 0x00E1 / 0x010F | BatChargeToEVC | Battery to EV Charger | 0: Disable, 1: Enable |
| 0x00F0 / 0x0155 | PeakShavingPeriodAChargeFromGridEn | PeakShaving Charge from Grid | 0: Disable, 1: Enable |
| 0x00F4 / 0x00B4 | VPPExitIdleEn | VPP Exit Idle | 0: Disable, 1: Enable |
| 0x00F5 / 0x00B3 | CT Installation Check | Fast CT Check | 0: Disable, 1: Enable |
| 0x0104 / 0x0166 | GenPeriod2SetEnable | Generator Time 2 | 0: Disable, 1: Enable |
| 0x0109 / 0x016B | ChargeFromGenEnable | Generator Charge | 0: Disable, 1: Enable |

The remaining converted registers are earlier-generation (GEN2/GEN3) entries not present in the V1.02 document; their behaviour is covered by the live-hardware verification below.

**EVC_Modbus_V3.5** — register `0x0615` is officially named **ElectronicLock** (`0: unlock, 1: lock`), so the entity is renamed `device_lock` → `electronic_lock`; `0x0616` is **CardActivateSwitch** (`0: Turn off, 1: Turn on`) → `rfid_program` renamed `rfid_card_activation`. (These entities change domain in this PR anyway, so the key rename adds no extra breakage.)

### Redundant "Enable" suffix dropped from switch names

Four converted entities had names ending in " Enable" — meaningful for a dropdown, redundant on a switch (the toggle itself expresses enable/disable): `Charge and Discharge Period2 Enable`, `VPP Exit Idle Enable`, `Fast CT Check Enable`, `Selfuse Night Charge Enable` → suffix removed. Display-name change only; keys and unique_ids are untouched, and existing entity_ids are preserved as usual.

Docs folder reviewed: `solax-entity-description.md` describes converted entities by key (`battery_heating`, `hotstandby`) without referencing the entity domain, so no doc updates are required.


Bl00d-B0b:
**Inverted-polarity case cross-verified against SolaX Cloud on real hardware.**

`HotStandBy` is the one converted register with inverted polarity (`0: Enabled, 1: Disabled` — undocumented in the public V1.02 protocol doc; empirically corrected by @wills106 back in June 2023, one day after its introduction). Verified end-to-end on an X3-Hybrid-G4:

| Layer | Value |
|---|---|
| SolaX Cloud setting | HotStandby **Disabled** |
| Register 0xA1 raw | `1` |
| readback scale `{0: 1, 1: 0}` | `0` |
| `switch.<hub>_hotstandby` | **off** ✅ |

The switch write side mirrors it: on → writes `0` via `value_function_disable_enable`. The 3-year-standing inverted mapping is preserved exactly by this conversion.

TCWORLD:
Any entries with `scale={0: 0, 1: 1},`, you can probably remove the scale line entirely as it seems basically redundant.

Bl00d-B0b:
Good catch, thanks - the `scale={0: 0, 1: 1}` identity mappings were indeed redundant (the raw register value passes through unchanged without them). Removed all 26 of them; only the inverted `scale={0: 1, 1: 0}` on the HotStandBy readback remains, where the mapping actually does something. Verified on my live system after the change.

rosenrot00:
@Bl00d-B0b @wills106 since 2026.07.2 I'm getting spammed with this. I think it originates from this PR.

Unfortunately this is a show stopper for 07.2 and requires 07.3. This implementation doesn't respect the write command for switches.
```

2026-07-28T18:05:30.074725692Z [homeassistant] 2026-07-28 20:05:30.067 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor lease_mode corresponding to switch lease_mode bit 0 has no integer value None
2026-07-28T18:05:30.075068294Z [homeassistant] 2026-07-28 20:05:30.068 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor manual_mode_control corresponding to switch manual_mode_control bit 0 has no integer value None
2026-07-28T18:05:30.075426396Z [homeassistant] 2026-07-28 20:05:30.068 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor phase_power_balance_x3 corresponding to switch phase_power_balance_x3 bit 0 has no integer value None
2026-07-28T18:05:30.075770122Z [homeassistant] 2026-07-28 20:05:30.069 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor selfuse_mode_backup corresponding to switch selfuse_mode_backup bit 0 has no integer value None
2026-07-28T18:05:30.076120599Z [homeassistant] 2026-07-28 20:05:30.070 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor selfuse_night_charge_enable corresponding to switch selfuse_night_charge_enable bit 0 has no integer value None
2026-07-28T18:06:00.067038068Z [homeassistant] 2026-07-28 20:06:00.061 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor charge_and_discharge_period2_enable corresponding to switch charge_and_discharge_period2_enable bit 0 has no integer value None
2026-07-28T18:06:00.077641791Z [homeassistant] 2026-07-28 20:06:00.061 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor discharge_cut_off_point_different corresponding to switch discharge_cut_off_point_different bit 0 has no integer value None
2026-07-28T18:06:00.088346519Z [homeassistant] 2026-07-28 20:06:00.062 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor device_lock corresponding to switch device_lock bit 0 has no integer value None
2026-07-28T18:06:00.105794184Z [homeassistant] 2026-07-28 20:06:00.063 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor hotstandby corresponding to switch hotstandby bit 0 has no integer value None
2026-07-28T18:06:00.119541280Z [homeassistant] 2026-07-28 20:06:00.063 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor extend_bms_setting corresponding to switch extend_bms_setting bit 0 has no integer value None
2026-07-28T18:06:00.130601485Z [homeassistant] 2026-07-28 20:06:00.064 ERROR (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor battery_heating corresponding to switch battery_heating bit 0 has no integer value None
```

wills106:
I only tested by restarting HA. I didn't check the logs.

I'm getting similar:
```
This error originated from a custom integration.

Logger: custom_components.solax_modbus.switch
Source: custom_components/solax_modbus/switch.py:214
Integration: SolaX Inverter Modbus (documentation, issues)
First occurred: 17:43:50 (434 occurrences)
Last logged: 19:31:21

SolaX: Sensor backup_gridcharge corresponding to switch backup_gridcharge bit 0 has no integer value None
SolaX: Sensor discharge_cut_off_point_different corresponding to switch discharge_cut_off_point_different bit 0 has no integer value None
```

Bl00d-B0b:
@rosenrot00 @wills106 thanks for the reports - root cause found and fix up in #2210.

What happens: when a readback register cannot be read (device does not answer that address, or the register got quarantined after failed retries), the hub stores `None` for that key - this condition is old and the **select** platform always degraded silently to *unknown* in that case, which is why it was never visible before. The new switch code logged an ERROR on every poll for it, and worse, substituted 0 - showing the switch as off regardless of the real device state.

#2210 makes switches degrade exactly like the selects did: unreadable readback → state *unknown* (debug log only), recovering automatically once the register reads again. Writes and the post-write debounce are untouched.

@rosenrot00 regarding "doesn't respect the write command" - if you saw a switch failing to actually write (beyond the state display being wrong from the None-as-0 substitution), please share which entity and I'll dig into that separately; the write path itself is unchanged by both PRs.

rosenrot00:
I saw a FC16 write from here

`2026-07-28T18:08:53.976109745Z [homeassistant] 2026-07-28 20:08:53.975 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [281472191114720] Error during service call to switch.turn_on: SolaX: multi-function single-register write was rejected by device 1 at register 0x9e: ExceptionResponse(dev_id=1, function_code=144, exception_code=3)`

Bl00d-B0b:
@rosenrot00 that FC16 error nailed the second bug - thanks. `switch.py` called `async_write_registers_single` unconditionally, and that helper issues **function 0x10** despite its name; the select platform had dispatched on `write_method` and all converted selects used the default `WRITE_SINGLE_MODBUS` = **function 0x06**. So every converted switch silently changed its write function code from 0x06 to 0x10, and your inverter rejects 0x10 at 0x9E - precisely "doesn't respect the write command".

#2210 now fixes both issues: writes dispatch on `write_method` exactly like `select.py` (restoring 0x06 for all converted switches), and unreadable readbacks show *unknown* instead of ERROR-spam + fake off. Verified against the pre-conversion tree that no converted select used a non-default write method.

PR #2202, opened 2026-07-28
Adjust Sofar feed-in power limit for parallel systems

### LINKED ISSUE #2147: [Bug]: SOFAR HYDxxKTL-3P feedin_maximum_power: max value limited to inverter max power, breaks master-slave setup
### Describe the bug

**Description**: The value for `feedin_maximum_power` can only be set up to the maximum power of a single inverter. In master-slave mode, however, the value must be higher than the power of the master alone. 

**Actual behaviour:** The value is limited to the maximum power of a single inverter: The value for `feedin_maximum_power` can only be set up to the maximum power of a single inverter. In master-slave mode, however, the value must be higher than the power of the master alone, as all inverters feed into the grid together.

**Expected behaviour:** It should be possible to set the value higher than the power of a single inverter.

**Latest working version:** 2026.05.x

### Integration Version

2026.06.x

2026.06.3

### Inverter brand

Sofar Solar

### Plugin used

plugin_sofar.py

### Serial prefix

SP1ES120

### Connection Method

LAN Stick Logger LSE 03

### Detailed Error Log

_No response_

### Additional context

_No response_

rosenrot00:
Sofar parallel systems currently use the fixed 20 kW limit for the maximum feed-in power.

For a parallel master, this change uses the configured total system capacity (`inverter_power_kw`) as the upper limit, following the existing SolaX implementation. Single inverters and slaves keep the existing 20 kW limit.

Added tests for both parallel master and single/slave configurations.

Fixes #2147

### DIFF
--- a/custom_components/solax_modbus/plugin_sofar.py
+++ b/custom_components/solax_modbus/plugin_sofar.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from datetime import datetime
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


wills106:
Now we have moved away from sliders to input boxes for things like export limits etc. It might be worth changing these values to the maximum specified in the docs.

Or just removing the minimum and maximum completely if the docs specify min/max.

The limits were mostly there to made the sliders more manageable.

PR #2210, opened 2026-07-28
fix: switch writes honor write_method; unreadable readbacks no longer spam errors

Bl00d-B0b:
Fixes both regressions reported in #2177 (by @rosenrot00 and @wills106) after 2026.07.2.

## 1. Writes ignored `write_method` — the FC16 rejections

`_write_switch_to_modbus` always called `async_write_registers_single`, which despite its name issues a **function 0x10** (write multiple) request. Every converted select had used the default `WRITE_SINGLE_MODBUS` (**function 0x06**), and some inverters reject 0x10 on those registers — exactly the reported error:

```
multi-function single-register write was rejected by device 1 at register 0x9e: ExceptionResponse(..., function_code=144, exception_code=3)
```

Switches now dispatch on `write_method` the same way `select.py` does: `WRITE_SINGLE_MODBUS` → `async_write_register` (0x06), `WRITE_MULTISINGLE_MODBUS` → `async_write_registers_single` (0x10). Verified against the pre-#2177 tree: none of the converted selects declared a non-default write method, so restoring the dispatch restores the exact function codes used before the conversion. No entity declarations change.

## 2. Unreadable readbacks — the ERROR spam

When a readback register cannot be read (device does not answer that address, or the register was quarantined after failed retries), the hub deliberately stores `None` for the key. The select platform has always degraded silently to *unknown* in that case; the switch platform logged an **ERROR every poll cycle** and substituted `0` — actively showing the switch as off regardless of the real device state.

| Readback condition | Before | After |
|---|---|---|
| reads normally | on/off | on/off (unchanged) |
| unreadable (`None`) | ERROR every poll + shows **off** | **unknown**, debug log, auto-recovers |
| non-integer value | unhandled `ValueError` | **unknown**, debug log |

One file (`switch.py`), single commit.

### DIFF
--- a/custom_components/solax_modbus/switch.py
+++ b/custom_components/solax_modbus/switch.py
@@ -16,6 +16,7 @@ from .const import (
     DEFAULT_MODBUS_ADDR,
     DOMAIN,
     WRITE_DATA_LOCAL,
+    WRITE_MULTISINGLE_MODBUS,
     BaseModbusSwitchEntityDescription,
     matches_modbus_protocol,
 )
@@ -190,13 +191,21 @@ class SolaXModbusSwitch(SwitchEntity, RestoreEntity):
             return
 
         payload: int = self._value_function(self._bit, is_on, self._sensor_key, self._hub.data)
-        _LOGGER.debug(f"Writing {self._platform_name} {self._key} to register {self._register} with value {payload}")
-        await self._hub.async_write_registers_single(
-            unit=self._modbus_addr,
-            address=self._register,
-            payload=payload,
-            register_data_type=getattr(self.entity_description, "register_data_type", None),
-        )
+        _LOGGER.debug(f"Writing {self._platform_name} {self._key} to register {self._register} with value {payload} method {self._write_method}")
+        if self._write_method == WRITE_MULTISINGLE_MODBUS:
+            await self._hub.async_write_registers_single(
+                unit=self._modbus_addr,
+                address=self._register,
+                payload=payload,
+                register_data_type=getattr(self.entity_description, "register_data_type", None),
+            )
+        else:
+            await self._hub.async_write_register(
+                unit=self._modbus_addr,
+                address=self._register,
+                payload=payload,
+                register_data_type=getattr(self.entity_description, "register_data_type", None),
+            )
 
     @property
     def is_on(self) -> bool | None:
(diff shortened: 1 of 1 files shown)


rosenrot00:
Why does it look so weird? 
image

wills106:
Do the sensors return unknown?

These appeared as proper switches earlier.

rosenrot00:
`2026-07-29 12:39:00.844 DEBUG (SyncWorker_8) [custom_components.solax_modbus] saved modified persistent date: {'_version': 1, 'remotecontrol_active_power': 0, 'remotecontrol_reactive_power': 0, 'remotecontrol_duration': 20, 'remotecontrol_autorepeat_duration': 18000.0, 'remotecontrol_import_limit': 20000, 'remotecontrol_pv_power_limit': 30000, 'remotecontrol_push_mode_power_8_9': 0, 'export_first_battery_charge_limit_8_9': 22.0, 'remotecontrol_target_soc_8_9': 95, 'remotecontrol_minimum_soc_8_9': 10, 'remotecontrol_timeout': 10.0, 'config_export_control_limit_readscale': 10.0, 'config_max_export': 15000, 'measured_power_offset': 0, 'measured_power_gain': 100, 'remotecontrol_power_control': 'Disabled', 'remotecontrol_set_type': 'Set', 'remotecontrol_power_control_mode': 'Export-First Battery Limit', 'remotecontrol_timeout_next_motion': 161, 'charge_and_discharge_period2_enable': None, 'discharge_cut_off_point_different': None, 'device_lock': None, 'hotstandby': None, 'extend_bms_setting': None, 'battery_heating': None, 'battery_to_ev_charger': None, 'mppt': None, 'peakshaving_charge_from_grid': None, 'vpp_exit_idle_enable': None, 'fast_ct_check_enable': None, 'ct_cycle_detection': None, 'lease_mode': None, 'manual_mode_control': None, 'phase_power_balance_x3': None, 'selfuse_mode_backup': None, 'selfuse_night_charge_enable': None, 'energy_dashboard_pv_variants_enabled': 1, 'energy_dashboard_home_consumption_enabled': 1, 'energy_dashboard_grid_to_battery_enabled': 1}`

I see `None`

wills106:
I'm getting:
2026-07-29 11:44:31.638 DEBUG (MainThread) [custom_components.solax_modbus.switch] SolaX: Sensor backup_gridcharge for switch backup_gridcharge has no value yet, state unknown

I wonder if `sensor_key=` isn't working for these switches?

wills106:
I have just reloaded the integration (not restarted HA) and the switches have their values.

I think the switches are trying to populate their values before the sensor has done the initial read.

Bl00d-B0b:
@wills106 your instinct was right, and it uncovered a third regression - now fixed in the updated PR.

**Root cause of the startup `no value yet` / weird-looking switches:** `switch.py` registered every switch's `sensor_key` in `hub.writeLocals` - but that dict is reserved for `WRITE_DATA_LOCAL` entities (select/number/time all guard the registration with exactly that check). The pollution had three effects:

1. every readback key got persisted into `{hub}_data.json` (@rosenrot00's debug dump shows the file), 
2. stale persisted values were re-imported at startup, and
3. the first-time-init path seeded `data[key] = None` - which is precisely the `has no value yet` state you logged, present before the first real register read.

The updated PR now contains three fixes in one commit:
- writes dispatch on `write_method` (restores function 0x06 for all converted switches - the FC16 rejection @rosenrot00 caught),
- `writeLocals` registration guarded by `write_method == WRITE_DATA_LOCAL`, matching the other platforms,
- unreadable readbacks show *unknown* instead of ERROR-spam + fake off, and switches restore their last on/off state at startup so the short window before the first read shows the last known state instead of a weird unknown toggle.

After updating, one HA restart may still show restored (last-known) states for a few seconds until the first poll - that is by design and matches how most integrations behave.

Bl00d-B0b:
Correction to my previous comment: the branch update with the third fix was pushed after this PR had already been merged, so 2026.07.3 contains only the write-method dispatch and the unknown-state handling. The `writeLocals`/startup fix is now in a proper follow-up PR rebased on current main - see the link above this comment.

PR #2218, opened 2026-07-29
Refactor Energy Dashboard lifecycle coordination

rosenrot00:
Dashboard topology is now updated when inverters are loaded, unloaded, or report a changed parallel role. Refreshes are coalesced, fixed startup delays and entity-name fallbacks are removed, and different plugin families are no longer aggregated together.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -400,11 +400,15 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             entry,
         )
     try:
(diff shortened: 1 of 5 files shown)


wills106:
Just looking at this conflict. Am I correct in thinking I need to accept the current change:
image
So I keep hub_data and parallel_setting lines

rosenrot00:
Yes, exactly. We should keep the simple hub_data and parallel_setting lines from this PR and remove the old polling/retry fallback block. The coordinator now relies on the hub’s already published data and must not trigger additional Modbus reads here. I’ve resolved the conflict and rebased the PR; all checks are green.

PR #2220, opened 2026-07-29
Add ESPHome serial proxy support via tmodbus and SerialX

rosenrot00:
Home Assistant can now expose local serial ports and ESPHome Serial Proxies through its serial port selector. The previous Pymodbus serial client uses PySerial internally and cannot open `esphome-hass://` ports.

Changes made:

- Use Home Assistant's serial port selector
- Move serial Modbus RTU communication to tmodbus and SerialX
- Add support for ESPHome Serial Proxies
- Keep the existing serial port config key and support existing local serial ports without migration
- Keep TCP Modbus communication on Pymodbus

@wills106 there are some PRs addressing bigger architectural concepts. Please be prepared that some might break the integration in special cases I can't foresee. If we find an issue, we can react fast.


rosenrot00:
@wills106 thanks for merging. Let's find out what breaks :)

PR #2227, opened 2026-07-29
fix: switches no longer pollute local persistent data; restore last state at startup

Bl00d-B0b:
Follow-up to #2210 — the branch update containing this fix was pushed ~4 hours after the PR had been merged, so it never reached upstream (my comment there announcing "three fixes" was outdated the moment it was posted; apologies for the confusion). This PR carries the remaining fix, rebased onto current main (2026.07.3).

## Root cause of the startup `no value yet` / unknown switches (reported by @wills106 and @rosenrot00 on #2210)

`switch.py` registered **every** switch's readback `sensor_key` in `hub.writeLocals` — but that dict is reserved for `WRITE_DATA_LOCAL` entities (select/number/time all guard the registration with exactly that check). Effects:

1. every readback key was persisted into `{hub}_data.json` on each local-data save (visible in rosenrot00's debug dump),
2. stale persisted values were re-imported into the data dict at startup, masquerading as fresh register reads,
3. the first-time-init path seeded `data[key] = None` — precisely the `has no value yet, state unknown` condition logged before the first real register read.

## Fix

- `writeLocals` registration is now guarded by `write_method == WRITE_DATA_LOCAL`, matching the other platforms.
- Switches restore their last on/off state at startup (`RestoreEntity` was already inherited but only used for `WRITE_DATA_LOCAL`), so the short window before the first register read shows the last known state instead of an unknown toggle — addressing the "why does it look so weird" screenshot.

One file, single commit. After this, a stray `{hub}_data.json` may still contain the previously-persisted readback keys; they are ignored from now on and get dropped on the next save.

### DIFF
--- a/custom_components/solax_modbus/switch.py
+++ b/custom_components/solax_modbus/switch.py
@@ -43,7 +43,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, async_add_e
             switch = SolaXModbusSwitch(hub_name, hub, modbus_addr, hub.device_info, switch_info)
             if switch_info.value_function:
                 hub.computedSwitches[switch_info.key] = switch_info
-            if switch_info.sensor_key is not None:
+            if switch_info.write_method == WRITE_DATA_LOCAL and switch_info.sensor_key is not None:
                 hub.writeLocals[switch_info.sensor_key] = switch_info
             dependency_key = getattr(switch_info, "sensor_key", switch_info.key)
             if dependency_key != switch_info.key:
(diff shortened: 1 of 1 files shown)


wills106:
Just copied this over locally.
My "Backup Grid Charge" is now showing up as off on start-up and remains in that state.

It's a bit strange that "Discharge Cut Off Point Different" previously was showing as off, but with this change it's gone to On.
The problem is, I can't really remember what "Discharge Cut Off Point Different" actually does?

image

Bl00d-B0b:
@wills106 both observations explained:

**Why "Discharge Cut Off Point Different" flipped off → on:** before these fixes, an unread readback was silently substituted with 0 — the switch *fabricated* "off". Now the real readback (holding `0x111`) is read and **On is the true stored setting of your inverter**; nothing changed on the device, the display became truthful. Cross-check from another unit: my X3-Hybrid-G4 reads `0x111 = 0` and correctly shows off.

**What it does:** it isn't in the V1.02 protocol doc (checked the full text — the def dates back to the Gen3 documentation). It enables a **separate battery discharge floor for on-grid operation**: with it On, the inverter uses "Battery Minimum Capacity - Grid-tied" (the adjacent readback `0x112`) as the discharge cut-off while grid-tied, distinct from the general Battery Minimum Capacity that then applies off-grid/EPS. Typical use: keep a deeper backup reserve for outages while allowing normal cycling on-grid (or vice versa). Since yours is On, it's worth checking both minimum-capacity values so the grid-tied floor is what you expect.

**Backup Grid Charge showing off and staying off:** with the writeLocals fix the `0xFD` readback is read properly on each poll, so that off should simply be the genuine register value now (previously it could be the fabricated kind).

Issue #2230, opened 2026-07-30
[Bug]: X3-HAC EV Charger - charge_added_cumulative off by factor 65536, and sporadic read timeouts on varying registers


xzsg4d4khs-svg:
### Describe the bug

Two separate issues with `plugin_solax_ev_charger.py` on an X3-HAC 22kW (Gen2, ARM v2.03), polled as a dedicated hub over Modbus TCP directly to the charger's own LAN address.

Reads are otherwise excellent: `communication_success_rate` sits at 100.0 % and all 27 enabled sensors return plausible values every 10 s. Writes work too - see Additional context.

### 1. `charge_added_cumulative` is off by exactly a factor of 65536 (hi/lo word swap)

```
sensor.<hub>_ev_charger_charge_added_cumulative   21 141 913.6 kWh
```

The same charger read through the separate `solax_http` custom integration (local HTTP API, entirely independent code path) reports the lifetime total as:

```
sensor.solax_api_charge_added_total              322.6 kWh
```

And 322.6 x 65536 = 21 141 913.6 - an exact match, not an approximation. Classic symptom of the two 16-bit halves of a 32-bit register being combined in the wrong order.

Related: `charge_added_total` reads `0.0 kWh` at the same moment, while the real lifetime total is 322.6 kWh. So the register that actually holds the lifetime figure appears to be the one mapped to `charge_added_cumulative`, and `charge_added_total` looks mislabelled or mapped to a register this firmware does not populate.

PRs #2178 and #2227 touched EVC energy word order, but this sensor is still wrong in 2026.07.4.

### 2. Register `0x634` (Timer Boost Start Time) never responds

Repeating several times per hour, see the log section below.

In `plugin_solax_ev_charger.py`, `0x634` is declared twice - once as a `SolaXEVChargerModbusTimeEntityDescription` and once as a `SolaXEVChargerModbusSensorEntityDescription` - both with `wordcount=2`, for Timer Boost Start Time.

The charger answers every other register fine but never this one. Raising `time_out` from 5 to 15 seconds in the hub options changed nothing: the message still reads `after 1 retries`, so the retry count appears fixed at 1 and the extra latency budget is not the issue. Either this X3-HAC firmware does not implement that register block, or the address / wordcount is off for Gen2 hardware.

Side effect worth noting: `communication_health` flips to `Degraded` while `communication_success_rate` stays at `100.0 %`. Those two contradict each other.

### Integration Version

2026.07.4

[...]

Issue #2213, opened 2026-07-29
[Bug]: Solax X3-HAC 11kW - Sensor: "Charge Added - Cumulative" wrong values


RoterRaecher:
### Describe the bug

The sensor **Charge Added - Cumulative** (sensor.solax_wallbox_charge_added_cumulative)
is displaying a value that is much too high.

<img width="582" height="495" alt="Image" src="https://github.com/user-attachments/assets/c07804d0-a7f8-4c19-9094-09f707f8f948" />

### Integration Version

2026.07.2

### Homeassistant core version

2026.7.4

### Inverter brand

X3-HAC 11kW

### Plugin used

plugin_solax_ev_charger.py

### Serial prefix

5030B0

### Inverter firmware versions

ARM:  1.56-1.15 MDSP:  1.58

### Connection Method

Solax WiFi+LAN

### Dongle firmware

1.009.02

rosenrot00:
Could you run the following commands on a machine that has access to the wallbox? Please replace the `<WALLBOX-IP>` with the correct IP and report back what it responds.

```
mbpoll -m tcp -a 1 -p 502 -1 -0 -t 3:hex -r 16 -c 2 <WALLBOX-IP>
mbpoll -m tcp -a 1 -p 502 -1 -0 -B -t 3:int -r 16 -c 1 <WALLBOX-IP>
mbpoll -m tcp -a 1 -p 502 -1 -0    -t 3:int -r 16 -c 1 <WALLBOX-IP>
```

RoterRaecher:
Hello,

here are the responds:

```
Protocol configuration: ModBus TCP
Slave configuration...: address = [1]
                        start reference = 16, count = 2
Communication.........: 10.2.4.8, port 502, t/o 1.00 s, poll rate 1000 ms
Data type.............: 16-bit register, input register table

-- Polling slave 1...
[16]:   0x090B
[17]:   0x1387
```

```
Protocol configuration: ModBus TCP
Slave configuration...: address = [1]
                        start reference = 16, count = 1
Communication.........: 10.2.4.8, port 502, t/o 1.00 s, poll rate 1000 ms
Data type.............: 32-bit integer (big endian), input register table

-- Polling slave 1...
[16]:   151720839
```

```
Protocol configuration: ModBus TCP
Slave configuration...: address = [1]
                        start reference = 16, count = 1
Communication.........: 10.2.4.8, port 502, t/o 1.00 s, poll rate 1000 ms
[...]

PR #2235, opened 2026-07-30
fix: scope EVC cumulative-energy big word order to GEN1 only

Bl00d-B0b:
Fixes the word-order half of #2230; follow-up to the discussion on #2178.

The `order32="big"` override on `charge_added_cum` was hardware-verified on my **X3-EVC-11kW Gen1** (serial `C31103*`, **ARM v1.18**, register 0x25 = 118, classified GEN1 by the plugin): at the same instant, the input copy `0x10-0x11` read `[0, 3]` (high-word-first, 0.3 kWh only when decoded big) while its holding twin `0x619-0x61A` read `[3, 0]` and decoded to the correct 0.3 kWh little-endian. On that unit the documented order is demonstrably wrong for the input copy.

#2230 shows the opposite population: an **X3-HAC 22kW Gen2** (ARM v2.03) follows the documented little order, so the global override inflated its lifetime total by exactly a factor of 65536 (322.6 kWh became 21,141,913.6 - the reporter's cross-check against the independent HTTP integration makes it airtight).

**Fix:** the definition is split per generation - GEN1 keeps `order32="big"` (hardware-verified), GEN2 returns to the documented default order. If more firmware populations surface (e.g. #2213, where neither order looks plausible), the same split can be refined per firmware version.


TCWORLD:
I'd already opened a PR to fix this.

wills106:
Does this need closing down then?

Bl00d-B0b:
@wills106 yes - closing. #2234 implements the same generational split (and covers GEN3/GEN4 explicitly, which is the better version) and is already merged; I opened this one before spotting TCWORLD's. Verified #2234 against my Gen1 unit's case: GEN1 keeps the big-endian decode it needs, the HAC Gen2 reports in #2230 get the documented order. Nothing from this PR remains needed.

PR #2250, opened 2026-08-02
Add support for Solis S6-EH3P10K02-NV-YD-L

ikkentim:
Adds support for Solis S6-EH3P10K02-NV-YD-L. This is a Hybrid Gen6 inverter 48V; 10KW with 2MPPTs

Serial prefix: 103330

Manual: https://zonnigewinkel.nl/wp-content/uploads/2026/03/Handleiding-Solis-3-fase-laag-voltage-omvormer.pdf


ikkentim:
Question:

specs say: 
MPPT number=2
Max input strings number=3

Should i add MPPT3 attribute to this definition?

wills106:
You could try adding in the 3rd MPPT.
But it probably wont return anything based on the wording you described.

I have known other Inverters only having 2 MPPT's but 3 ports to connect to.

I don't know with the combined MPPT if you are meant to balance then both ie same number of panels? Or does it internally add them in series?

PR #2251, opened 2026-08-02
Added Charge cut-off SOC to limit battery charge SOC

flipperdk:
Adds a number entity (charge_cutoff_soc) to set an upper SOC limit for battery charging, plus a matching sensor to read the current value back.

**Why:**
On cloudy days in spring/autumn/winter, when the forecast suggests the battery won't fully charge from PV, you may want to partially charge from the grid during low-tariff periods. This entity sets the upper boundary.

**Testing:**
Tested and working on my own MHT-10K-25 (FW V08.03.00.04-V10.26.05.00).

**Note on the register:**
Register 52506 is undocumented in Solinteg's official Modbus documentation. I located it by changing the cut-off SOC in Solinteg's UI and diffing against a full register readout from the inverter. I've only been able to verify it on my own unit, so I can't confirm it applies unchanged across other Solinteg models/firmware.

### DIFF
--- a/custom_components/solax_modbus/plugin_solinteg.py
+++ b/custom_components/solax_modbus/plugin_solinteg.py
@@ -581,19 +581,19 @@ NUMBER_TYPES = [
         icon="mdi:import",
     ),
     SolintegModbusNumberEntityDescription(
-        name="Charge Cutoff SOC",         
+        name="Charge Cutoff SOC",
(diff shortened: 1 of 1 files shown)


braetkuegeli:
Works also on MHT-10K-40 (V07.04.00.00-V18.04.05.0)

PR #2262, opened 2026-08-05
Add Growatt SZG serial and TJ1 firmware recognition

### LINKED ISSUE #2261: [Bug]: Growatt unrecognized inverter type — add serial prefix SZG and firmware prefix TJ1
## Describe the bug

The Growatt plugin does not recognize an inverter whose **serial number starts with `SZG`** and whose **firmware version starts with `TJ1`**.

Modbus communication itself works (serial/firmware are read successfully), but detection fails with:

```text
unrecognized Growatt inverter type - firmware version : TJ1.0
```

As a result, the inverter is not classified correctly and entities are not created properly.

## Inverter details (known)

| Field | Value |
|---|---|
| Integration | `solax_modbus` (wills106 / HACS) |
| Plugin | `plugin_growatt.py` |
| Serial (as read / entity) | `SZGDCL505K` (prefix **`SZG`**) |
| Firmware (as logged) | **`TJ1.0`** (prefix **`TJ1`**) |
| Phases / topology observed | 3-phase grid-tied PV, 2 MPPT strings |
| Exact nameplate model / kW | **Unknown / not confirmed from nameplate** |
| Notes on rating | Integration reported `inverter_power_kw: 5` after local fix; plant peak AC output observed around **~13.8 kW**, so the 5 kW figure may be wrong or incomplete — please do not treat it as confirmed nameplate data |

## Local fix that worked

Adding the following mappings as **`PV | GEN2 | X3`** fixed detection and allowed the normal PV entity set to populate:

```python
# SERIAL_PREFIX_TYPES
"SZG": PV | GEN2 | X3,  # 3-phase PV, 2 MPPT (serial SZGDCL505K)

# FIRMWARE_PREFIX_TYPES
"TJ1": PV | GEN2 | X3,  # firmware TJ1.0 (SZG serials)
```

Suggested placement in current `plugin_growatt.py` (main):

- Near other serial prefixes such as `"RPH": PV | GEN2 | X3` in `SERIAL_PREFIX_TYPES`
- Near other firmware prefixes such as `"DM1": PV | GEN2 | X3 | MPPT4` in `FIRMWARE_PREFIX_TYPES`

## Optional secondary log (may be unrelated)

After entities started appearing, Home Assistant also logged:

```text
Platform solax_modbus does not generate unique IDs. ID Growatt_SZGDCL505K_inverter_temperature already exists
```

Included in case it helps triage duplicate entity creation / reload behavior.

## Expected behavior

`SZG` / `TJ1` should be recognized so the Growatt plugin classifies the inverter and creates the appropriate PV entities without a local patch.

## Actual behavior

Log shows unrecognized inverter type for firmware `TJ1.0`; classification fails until `SZG` and `TJ1` are added to the prefix maps.

## Additional context

Happy to provide more Modbus register dumps or entity lists if useful. Exact commercial model string from the nameplate is not available at filing time; classification as `PV | GEN2 | X3` was validated empirically (3-phase voltages/currents and 2 PV strings reading correctly after the local mapping).

hugotanure:
## Summary
- Recognize Growatt inverters with serial prefix `SZG` and firmware prefix `TJ1` (e.g. serial `SZGDCL505K`, firmware `TJ1.0`).
- Map both to `PV | GEN2 | X3` (3-phase PV, 2 MPPT), matching the local fix that cleared the `unrecognized Growatt inverter type` error.
- Fixes #2261.

## Test plan
- [ ] With an SZG / TJ1 Growatt on Modbus, confirm setup no longer logs unrecognized inverter type
- [ ] Confirm 3-phase PV entities populate (grid L1/L2/L3, PV1/PV2 strings)
- [ ] Review that classification as GEN2 remains appropriate if a named nameplate model is confirmed later

Made with [Cursor](https://cursor.com)


TCWORLD (changes requested):
It shouldn't require both of the entries to be detected. If the SZG entry alone is not enough , then only the TJ1 entry alone is required.

Could you confirm what your inverter model is - the comment should follow the existing entries - including the inverter model not your serial number.

wills106:
@hugotanure have you been able to look at TCWORLD's comment about duplication?

hugotanure:
Confirmed the SZG entry alone is sufficient: with both entries installed, the entity unique IDs are built from SZGDCL505K (e.g. Growatt_SZGDCL505K_inverter_temperature), which means detection matched in the SERIAL_PREFIX_TYPES loop and never reached the firmware fallback. I've dropped the TJ1 entry.

The inverter model is a MID 15KTL3-X (3-phase, 2 MPPT)

Issue #2269, opened 2026-08-07
[Bug]: Solax X3 Gen4 wrong battery SOH register


litinoveweedle:
### Describe the bug

Hello,

current implementation uses for Solax X3 Gen4 register 0xBF, which doesn't work:

```
    SolaXModbusSensorEntityDescription(
        name = "Battery State of Health",
        key = "battery_soh",
        icon = "mdi:battery-heart",
        native_unit_of_measurement = PERCENTAGE,
        register = 0xBF,
        register_type = REG_INPUT,
        entity_registry_enabled_default = False,
        allowedtypes = AC | HYBRID | GEN4 | GEN5,
    ),
```

battery BMS firmware 2.13:

0x00BF was reporting 0:

```
litin@desktop:~$ mbpoll -a 1 -r 191 -c 1 -t 3 192.168.3.63
mbpoll 1.0-0 - FieldTalk(tm) Modbus(R) Master Simulator
Copyright © 2015-2019 Pascal JEAN, https://github.com/epsilonrt/mbpoll
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type 'mbpoll -w' for details.

Protocol configuration: Modbus TCP
Slave configuration...: address = [1]
                        start reference = 191, count = 1
Communication.........: 192.168.3.63, port 502, t/o 1.00 s, poll rate 1000 ms
Data type.............: 16-bit register, input register table

-- Polling slave 1... Ctrl-C to stop)
[191]: 0
-- Polling slave 1... Ctrl-C to stop)
[191]: 0
-- Polling slave 1... Ctrl-C to stop)
[191]: 0
-- Polling slave 1... Ctrl-C to stop)
^C--- 192.168.3.63 poll statistics ---
4 frames transmitted, 3 received, 0 errors, 25.0% frame loss
```

but 0x003D was reporting meaningfull value of 99

```
litin@desktop:~$ mbpoll -a 1 -m tcp -p 502 -r 0x003D -c 1 -t 3:int 192.168.3.63
mbpoll 1.0-0 - FieldTalk(tm) Modbus(R) Master Simulator
Copyright © 2015-2019 Pascal JEAN, https://github.com/epsilonrt/mbpoll
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type 'mbpoll -w' for details.

Protocol configuration: Modbus TCP
[...]

PR #2270, opened 2026-08-07
Migrate SolaX GEN4 SoH Register for Proto >= 100

TCWORLD:
This was previously excluded from the 0x3D register change as other battery related sensors had issues and there was no evidence supporting its migration.

Based on recent testing it seems that the inverters may respond correctly to the new register.

Potential fix for #2269.


TCWORLD:
I'm going to close this for now as on closer inspection I think its not required.

PR #2275, opened 2026-08-11
EVC/HAC: rework against the current V3.6 and V1.00 documents — segmentation fix, firmware-version gating, Parallel/EMS/Datahub sub-devices

Bl00d-B0b:
## Summary

A complete rework of the SolaX EV charger plugin against the two current vendor documents — **EVC ModbusRTU V3.6** (Gen1) and **X1 X3 EV Charger Modbus TCP&RTU V1.00_EN** (Gen2 / HAC, 2025-06-17) — plus answers from SolaX support (ticket 703158, and Doris on the Gen1/Gen2 split).

Every register in both documents is now accounted for: **160 implemented, 12 deliberately out of scope, 0 undocumented, 0 unresolved document conflicts**. All Gen1 behaviour is verified on hardware (X3-EVC-11kW, ARM V1.18); the Gen2 side follows the document exactly — **that is where testers are needed, see below**.

## The original bug

SolaX support confirmed the EVC holding registers are internally segmented: `0x600–0x650` is one segment, `0x660–0x66F` another (master/slave parallel operation). An FC03 read spanning the boundary **silently returns zeros** for the second segment rather than an error — which is why Max Charge Current read 0. Fixed by anchoring a new read block at `0x668`, which also allowed removing the `block_size=32` workaround that predated this knowledge.

## Firmware-version gating

The plugin now publishes the firmware version (input `0x25`, e.g. V1.18 → 118) as `hub.modbus_protocol_version`, enabling `modbus_min`/`modbus_max` gating for EV chargers. Gates follow the V3.6 revision history and grow monotonically by address across `0x624–0x664` (fw 1.09 → 1.14). On a HAC the version source is input `0x78` (reads 100), so Gen2 variants are ungated.

Registers that **moved** across firmware versions are exposed as one entity whose backing register follows the firmware — no duplicate "Alt" entities:

| Entity | ≤ V1.11 | ≥ V1.12 | ≥ V1.14 |
|---|---|---|---|
| Charge Power L1–L3 | `0x08–0x0A` | `0x100–0x102` | |
| Run Mode | `0x1D` | `0x106` | |
| Charge Phase (X1) | `0x625` | `0x63B` | |
| Charge Mode | | `0x104` (sensor) | `0x641` (select) |

## Max Charge Current

Support suggested `0x64F` for the maximum current. Live testing showed it is the *persistent device limit*, not the setpoint — SolaXCloud itself writes `0x668`, and external writes to `0x64F` are acknowledged but ignored. So **Max Charge Current stays on `0x668`**, and a new optional `max_key` on number descriptions caps the Max/Min sliders dynamically from the `0x64F` readback.

Related fix: rated current is now taken from an explicit `(power rating, phases) → nameplate amps` map instead of `watts / (phases × 230)`, which truncated the **X1 7.2 kW unit to 31 A instead of 32 A**. Covered by new unit tests.

## New in this PR

- **Parallel sub-device** — `0x38–0x65` and `0x666–0x66F` as read-only diagnostics, gated by capability (`0x107 == 0xAA55`) plus the Parallel Mode option. Chargers without parallel support see only "Parallel Support = No".
- **EMS ∕ V2G sub-device** (Gen2) — the full `0xA000–0xA02A` status and `0xA100–0xA106` configuration blocks behind a new translated option (9 languages).
- **Datahub sub-device** — the `0x700–0x70D` power-control interface plus `0x624`/`0x643`, behind its own translated option. Exposed **read-only**: writes are ACKed and re-asserted by the firmware within ~0.1 s (measured), because this interface belongs to the paired inverter/Datahub.
- **Decoded fault descriptions** — the documented 32-bit map on `0x1E` and EMS `0xA028`, rendered as up to five names plus "(+N more)" instead of a raw integer.
- **Charge Start Time** (`0x31–0x36`, Gen2), **Auth Priority** (`0x650`), **Module/LCD/Modbus versions** (`0x71/0x73/0x78`), **Screen Fitted** (`0x24`), plus the doc-based entities for unbalance, Mode Button, Min Charge Current, Green 30s Delay and the serial numbers.
- **OCPP removed entirely** — a commercial-edition feature, out of scope for this integration.

## Correctness fixes found by auditing against both documents

- **Signed decode**: phase currents/powers followed each document (`u16` on EVC, `i16` on HAC). Previously unsigned everywhere, so a discharging HAC would have read −3000 W as 62536 W.
- **`0x23`** decoded per the current document: EVC editions (Home/Commercial/Integrated) vs HAC application scenes (Solar/Standard/OCPP).
- **`0x28`** renamed **Wiring Phase** — both documents describe the single-phase wiring assignment, not live activity; the phase controls are X1-only accordingly.
- **RTC**: verified on hardware that `0x61E–0x623` is a **UTC base clock** and the device renders display time as base + `0x61D` offset, with `0x61D` owned by the cloud (re-asserted from the app within a minute). Sync RTC now writes UTC only and leaves the timezone alone.
- **Display precision** (touches `sensor.py`, so it affects every plugin — see below): scaled sensors now publish `suggested_display_precision`, so a 0.01-scaled register shows its two decimals in the UI instead of being rounded to a whole unit by the frontend.

## Changes outside the EV charger plugin

Kept as small as possible; reviewers may want to look here first.

| File | Change |
|---|---|
| `sensor.py` | Publishes `suggested_display_precision` — **only** for sensors whose scale is a real fraction (< 1) or which explicitly ask for integers (`rounding=0`). Unscaled sensors on other plugins keep Home Assistant's current behaviour, so no inverter's readings change appearance. |
| `__init__.py` | `rounding=0` now yields an `int` instead of a float ending in `.0`; registers the two new device groups. |
| `number.py`, `switch.py` | Publish the accepted value locally after a successful write, so an entity does not report the stale value until the next poll. |
| `const.py`, `config_flow.py`, `strings.json`, 8 translations | The two new options (EMS ∕ V2G, Datahub), hidden for plugins that do not implement them. The options dialog now also scans `SENSOR_TYPES` when deciding whether a plugin supports a device group — previously a sensors-only group could never show its option. |

## Testing

- Hardware: X3-EVC-11kW Gen1, ARM V1.18 — all Gen1 entities exercised live, including write tests on the current/mode/RTC registers and the parallel and Datahub blocks.
- Automated: new `tests/unit/test_evc_rated_current.py` (11 cases); full suite green on Python 3.12 and 3.13; ruff + mypy clean.

## 🙏 Gen2 (HAC) testers wanted

I only have a **Gen1 (EVC)** charger. The Gen2 side is implemented strictly from the V1.00 document and cannot be verified here. If you own an **X1/X3-HAC**, testing feedback would be very welcome — especially:

1. **EMS ∕ V2G hub** (enable the option) — do the `0xA000` block values look sane, and does charge control behave?
2. **`0x664` Main Breaker Limit Switch** — the polarity is an informed assumption (switch **on** writes 0 = "Closed"). If it acts inverted on real hardware, that is a one-line fix.
3. **`0x21` power rating** — the two HAC documents contradict each other on the value map; V1.00 is implemented. Does your model report the expected rating?
4. **Registers V1.00 dropped but V1.0.0 documented** (PE measurements `0x03`/`0x07`, grid currents `0x12–0x14`, boost and unbalance settings): does your hardware still answer them?

Bug reports are the intended mechanism here — the policy throughout is *implement per the documents, fix on report*.

Full register-by-register audit (both documents, live values, decode, gating): available on request.


wills106:
Review output:

## Review of `2275_diff.txt` (SolaX EV Charger rework)

### 1. F-strings — PASS ✅
The diff contains **no f-strings in logging statements**, so it does not reintroduce the Ruff G004 violation that the earlier `060fe322` commit ("Replace SolaX f-string logs") removed.

The only 4 f-strings in the diff are all in **return statements** (non-logging), which is consistent with the existing codebase convention (e.g. [`plugin_solax_ev_charger.py`](homeassistant-solax-modbus/custom_components/solax_modbus/plugin_solax_ev_charger.py:1738) already uses `return f"ARM v{fw}"`):
- Line 320: `return f"Unknown fault 0x{raw:08X}"`
- Line 322: `return ", ".join(names[:5]) + f" (+{len(names) - 5} more)"`
- Line 336: `return f"Unknown fault 0x{raw:08X}"`
- Line 338: `return ", ".join(names[:5]) + f" (+{len(names) - 5} more)"`

All `_LOGGER.*` calls in the diff use printf-style (`%s`/`%04X`), matching the established convention. No f-strings were reintroduced.

### 2. Reverting later PRs — PASS ✅
The diff does **not revert any later-PR changes**. Key evidence:
- The diff's base blobs (`__init__.py` = `98587e2c4`, `config_flow.py` = `8078406bf`) are **older** than the current HEAD (`bb8064c3e`), which already includes later PRs (2283 EPS sub-device, 2284 parallel sub-device, 2285 external-generator, 2286, 2296).
- The diff **never references** any later-PR content: no `external_generator`, `read_gen`, `CONF_READ_GEN`, `device_group_of`, `_plugin_supports_energy_dashboard`, or `_plugin_uses_feature_flag`.
- The diff does **not touch** `plugin_solax.py` (the file targeted by later PRs 2283/2284) — 0 occurrences.
- Of the 419 removed content lines, none correspond to later-PR additions. The only `ems`/`datahub`/`parallel`/`eps` matches in removed lines are the diff's **own** additions (e.g. the single `key="datahub_charge_current"` removal is the diff's own reorganization, replaced by its own new version in the same diff).

### Note (not a defect)
The diff does not apply cleanly to the current HEAD via `patch` — `__init__.py` hunk #3 fails because the current tree already has `external_generator` in `DEVICE_GROUP_NAMES`/`GATED_DEVICE_GROUPS` (added by later PR 2285). This is expected since the diff was generated against an older base; it is a **merge-ordering** consideration, not a revert. When rebasing 2275 onto current HEAD, the `ems`/`datahub` entries should be added alongside the existing `external_generator` entry rather than replacing the `pm` block.

### Summary
Both requested checks pass: no f-strings were reintroduced (logging stays printf-style), and no later-PR changes are reverted.

PR #2276, opened 2026-08-11
Fix X1 charge current key collision on EV charger

Bl00d-B0b:
## Problem

In `plugin_solax_ev_charger.py`, two sensor descriptions share the key `charge_current`:

- the internal readback sensor at holding register `0x628` (backs the **Charge Current** number entity)
- the X1 diagnostic measurement sensor at input register `0x4` (disabled by default)

Both write to `hub.data["charge_current"]`. If an X1 user enables the diagnostic sensor, the Charge Current number entity alternates between the written setpoint and the live measured current on each poll cycle.

## Fix

Rename the input-register measurement sensor key to `charge_current_measured` so it no longer collides with the setpoint readback. The entity display name ("Charge Current") is unchanged.


Bl00d-B0b:
Superseded: the key-collision fix is folded into #2275 so all EVC changes ship in one PR.

PR #2277, opened 2026-08-11
SolaX: align entity names with the cloud app, unify time schedules and sensor precision

Bl00d-B0b:
> Stacked on #2286. Review the last commit only — the diff includes the parent until it merges.

Three related pieces of naming work on the main device and the smaller SolaX plugins:

1. hybrid battery-mode settings aligned with the SolaX cloud app,
2. one consistent scheme for every time schedule,
3. glued and inconsistently spelled display names split and normalised.

The equivalent work for the EPS, Parallel and External Generator sub-devices lives in #2325, alongside the decoding fixes for those same sub-devices.

**Option values are deliberately untouched** — `Feedin Priority`, `Enabled Feedin Priority` and the SoC-mode options keep their exact strings, since automations select options and match states by string. Keys are unchanged, so sensor/number/select/switch entity ids do not move either. Time entities are the one exception; see the note at the end.

## 1. Battery-mode settings follow the app

| Before | After |
| --- | --- |
| Selfuse Night Charge | Self Use Charge from Grid |
| Selfuse Discharge Min SOC | Self Use Min SoC |
| Selfuse Nightcharge Upper SOC | Self Use Charge Battery to |
| Feedin Discharge Min SOC | Feed-In Priority Min SoC |
| Feedin Nightcharge Upper SOC | Feed-In Priority Charge Battery to |
| Backup Discharge Min SOC | Backup Min SoC |
| Backup Nightcharge Upper SOC | Backup Charge Battery to |

Self Use Charge Battery to only applies while Self Use night charge is enabled, so it uses `active_when` (from #2279) to follow that switch.

Every remaining entity name that spelled the state of charge as `SOC` now uses `SoC` — Min SoC, Selfuse Backup SoC, Switch On SoC, Battery Charge Upper SoC, Peak Shaving Max/Reserved SoC, the Dual Battery Charge Upper SoC readbacks, the Remote Control Target/Minimum SoC numbers, and `ForceCharge SOC` in the A1-J1 plugin, which the earlier sweep missed. Option values such as `Reference SOC` are again left alone.

`Charge and Discharge Period2` becomes **Charge/Discharge P2**, and the four Gen4/Gen5/Gen6 times it controls — Force Charge P2 Start/Stop (`0x6D`/`0x6E`) and Allowed Discharge P2 Start/Stop (`0x6F`/`0x70`) — now use `active_when` so they only exist while that switch is on, matching how the app hides the second period. Gen2/Gen3 keep their old behaviour; those models have no such switch.

The four Battery Heating times (`0xD0`–`0xD3` → `0xA4`–`0xA7`) only exist while the Battery Heating switch (`0xCF` → `0xA3`) is on. A heating level (Low/Medium/High) setting exists only for dual-battery systems (`0x20A`/`0x20B` per the V1.02 protocol document, already exposed as Gen5/Gen6 readback sensors); a Gen4 rejects that whole register block, verified live.

## 2. One scheme for every time schedule

The time entities had grown several conventions — `End` vs `Stop`, a redundant `Time` suffix, and bare period numbers in different positions. They now all read:

**`<Function> [P<n>] Start | Stop`**

- **`Stop`, never `End`** — so each pair sorts Start-then-Stop in an alphabetical list (`Sta` < `Sto`). `Stop` is also what SolaX's own registers are called (`work_stop_time`).
- **No `Time` suffix** — the entity is already a time; `Timer Boost Start Time` was saying it twice.
- **`P<n>` for periods**, so a period's entities group together instead of every Start grouping before every Stop.

37 time entities across three plugins, with no exceptions:

| Plugin | Entities |
| --- | --- |
| `plugin_solax.py` (22) | Force Charge P1/P2 · Allowed Discharge P1/P2 · Battery Heating P1/P2 · Backup Charge · Peak Shaving Discharge P1/P2 · Work Time P1/P2 |
| `plugin_solax_lv.py` (12) | Charge P1–P3 · Discharge P1–P3 |
| `plugin_solax_ev_charger.py` (3) | Timer Boost Start/Stop · Smart Boost Stop |

The same `P<n>` notation is applied to the period-scoped settings that are not times, so a period reads as one block: **Peak Shaving Discharge P1/P2 Limit**, **Force Time P1/P2 Max Capacity**, **Charge/Discharge P2**.

The A1-J1 plugin has no time entities; its eight schedule readback *sensors* follow the same scheme (`Charger Start Time 1` → `Charger P1 Start`), as do the 16 Dual Battery heating readbacks (`Heating Period 1 End Hour` → `Heating P1 Stop Hour`).

## 3. Glued and inconsistent names

| Before | After |
| --- | --- |
| `ActivePower L1` | `Active Power L1` |
| `Reactive or ApparentPower L1` | `Reactive or Apparent Power L1` |
| `PeakShaving …` | `Peak Shaving …` |
| `PowerControlMode Trigger` | `Power Control Mode Trigger` |
| `Remotecontrol …` **and** `RemoteControl …` (two spellings) | `Remote Control …` |
| `Remotecontrol TimeOut` | `Remote Control Timeout` |
| `ForceTime` / `ForceCharge` | `Force Time` / `Force Charge` |
| `HotStandBy` · `PowerFactor` | `Hot Standby` · `Power Factor` |
| `MateBox enabled` · `Safety code` · `Inverter DSP hardware version` | Title Case |
| `wAS4777 Power Manager` | `WAS4777 Power Manager` (typo) |

Proper nouns are deliberately left intact: SolaX, MateBox, WiFi, SoC, SoH, BMS, EPS, PV, MPPT. Other vendors' plugins have similar glued names but are out of scope for a SolaX PR.

## 4. Sensor rounding matches the register scale

A value is `round(raw * scale, rounding)`. A rounding coarser than the scale throws away digits the register provides; a finer one can never produce them. Since #2275 the sensor platform derives Home Assistant's display precision from the same `rounding`, so a mismatch is visible in the UI as well.

| Scale | Rounding |
| --- | --- |
| 0.1 | 1 |
| 0.01 | 2 |
| 0.001 | 3 |

**Digits that were being discarded** — Battery Voltage/Current Charge, Battery Charge Float Voltage, Battery Discharge Cut Off Voltage, Inverter Frequency L1/L2/L3 (scale 0.01, rounded to 1); Battery Input/Output Energy Total, Total Solar Energy, Today's Yield on MIC, and Power Factor in the LV and Mega/Forth plugins (scale 0.001, rounded to 1 or 2). An inverter frequency of 49.97 Hz was being reported as 50.0.

**Digits that could never exist** — Today's Solar Energy, Today's/Total Yield, E Charge Today/Total, Total Grid Export/Import all asked for 2 decimals on a 0.1 scale. One carried the comment `# GEN4 | GEN5 might be 1`; with a scale of 0.1 no second decimal is arithmetically possible, so 1 is correct for every generation.

**Left alone** — `Time Count Down` keeps its explicit `rounding=0`. An integer countdown is a deliberate display choice, so explicit zeros are preserved rather than overwritten.

29 entities in total. No register mappings, scales or keys change — only how many decimals each sensor reports. (The EPS entity with the same issue is handled in #2325, alongside the other fixes for that sub-device.)

## 5. MIC Power Limits reads the right register on MIC-G2

The control (write register `0x060F`, `PowerLimitsPercent`) had a readback that existed only for the first MIC generation, so on a **MIC-G2 the number showed `unknown`** — it had no value to display. Two causes:

- the readback sensor was gated `MIC | GEN | X3` while the number is `MIC | X3`, so no MIC-G2 ever got one;
- the register differs between generations — reading `0x0332` on a MIC-G2 returns **60000**, which is not a percentage.

| Document | Register | Meaning |
| --- | --- | --- |
| MIC V1.05 | `0x0332` | `PowerLimitsPercent`, R, 1 %, Uint16 — correct for the **first** generation |
| MIC-G2 V4.0 | `0x034F` | low byte `MpptScanMode`, **high byte `PowerRatio`** — percent of power limits |

So the first generation keeps `0x0332`, and MIC-G2 gains a readback on `0x034F` decoded as the high byte (`REGISTER_U8H`). The write register `0x060F` is unchanged and correct for both — it is **write-only**, which is why it cannot be read back directly.

**Verified on hardware** (X3-MIC G2, `MC208TIA808030`):

```
0x034F = 0x6400 → high byte 100 → app shows Power Limit 100 %
0x034F = 0x6300 → high byte  99 → after setting the limit to 99 %
low byte (MpptScanMode) unchanged in both reads
```

The entity now tracks the value instead of reporting `unknown`.

Renamed **"PV Limit" → "Power Limits"**, matching the SolaX app and the register name. ⚠️ Unlike the other renames in this PR, this one **does move the entity id** (`number.<hub>_pv_limit` → `number.<hub>_power_limits`), observed on the test system.

**Entity ids:** time entities are not registry-backed in this integration, so their entity ids derive from the display name — a rename therefore moves the id too (e.g. `time.<hub>_charge_1_start` → `time.<hub>_charge_p1_start`). That is unavoidable for any time-entity rename and is why they are grouped into one PR rather than spread across several. Sensors, numbers, selects and switches are unaffected: their ids come from their keys, which do not change.

Verified on hardware (X3 Hybrid 15kW Gen4, X3-EVC-11kW, HA 2026.9.1): all renamed entities load and report their live values.


TCWORLD:
> Automations that select that option by string need updating.

Please don't break everyone's automatons for something that is purely stylistic...

Bl00d-B0b:
Fair point — agreed and fixed. The select option values (`Feedin Priority`, `Enabled Feedin Priority`) are back to their original strings, so nothing that automations reference changes: entity ids, keys and option values are all untouched by this PR. What remains is purely display names (friendly names shown in the UI), which HA doesn't use for automation matching. The description's breaking-change note is removed accordingly.

PR #2278, opened 2026-08-11
Unpin tmodbus and bump to 0.5.0

balloobbot:
tmodbus 0.5.0 made tenacity a required dependency and dropped the now empty `smart` extra, so the requirement becomes `tmodbus[async-serial]`.

The current pin is breaking compatibility with the pre-release of solaredge which is adopting [`modbus-connection`](https://home-assistant-libs.github.io/modbus-connection/): https://github.com/WillCodeForCats/solaredge-modbus-multi/issues/1031#issuecomment-5257798589

_btw, I've been including this integration as a test bed to see how modbus connection works in real world use cases. The code can be found [here](https://github.com/balloobbot/homeassistant-solax-modbus/tree/migrate-to-modbus-connection)._

### DIFF
--- a/custom_components/solax_modbus/manifest.json
+++ b/custom_components/solax_modbus/manifest.json
@@ -9,6 +9,6 @@
   "integration_type": "hub",
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/wills106/homsassistant-solax-modbus/issues",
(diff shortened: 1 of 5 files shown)


balloob:
I manually created this PR. Repo rules prevented PRs that are not from collaborators, so I couldn't open a PR with this user.

PR #2284, opened 2026-08-13
Parallel: own sub-device, entities removed when the option is off

Bl00d-B0b:
> Stacked on #2279 (sub-device infrastructure). Review the last commit only — the diff includes the parent until it merges.

The 54 parallel-mode entities move from the main inverter device to `<hub> Parallel`, linked to the inverter with `via_device`. The redundant `PM` prefix is dropped from the names, since the device now provides that context. **Keys and entity ids are unchanged.**

With Parallel Mode (Master-Slave) switched off the entities are not created and the device is removed from the registry.

Verified on an X3-Hybrid-G4 15 kW.


wills106:
Review output:

Reviewed 2284_diff.txt against the homeassistant-solax-modbus source (plugin_solax.py) and the sibling logic in init.py/sensor.py/config_flow.py.

Result: the diff is correct and complete.

Details:

Touches only custom_components/solax_modbus/plugin_solax.py, in the "Input - Parallel Mode" section.
Applies cleanly to current main (82e072fd).
Exactly 54 parallel-mode entities changed, matching the PR comment: each drops the PM prefix from name= and gains device_group="pm".
No key= lines changed, so entity ids are unchanged (verified by grep). The only test reference, pm_inverter_count in test_energy_dashboard_coordinator.py, uses the key and is unaffected.
The device_group="pm" wiring is already supported: it maps to the "Parallel" sub-device via DEVICE_GROUP_NAMES, links via via_device in group_device_info(), is gated by CONF_READ_PM (read_pm, default False) in GATED_DEVICE_GROUPS/device_group_enabled(), and the device is removed from the registry by async_cleanup_disabled_devices() when the option is off. No other files need changes.
The change is a clean, self-contained rename+grouping that relies on existing infrastructure. No issues found.

PR #2285, opened 2026-08-13
External Generator: own sub-device with dependency-driven availability

Bl00d-B0b:
> Stacked on #2279 (sub-device infrastructure). Review the last commit only — the diff includes the parent until it merges.

The 33 generator entities were gated by the **Dry Contact Box** option, which is a different feature — in the SolaX app the external generator is its own tree. They move to a `<hub> External Generator` device behind a new **External Generator Support** option (`read_gen`, off by default, translated in all shipped languages), and each entity only exists while its parent in that tree is active:

```
External Generator (0xC7): Disabled | ATS Control | Dry Contact
  Start Gen Method (0xE3): only with Dry Contact
    Switch On SoC (0xE4) / Switch Off SoC (0xE5): only with Reference SOC
  Max Run Time (0xE6), Max Rest Time (0xE7), Hold Min Power (0x10E)
  Allow Work Start Time (0xE8) / Allow Work Stop Time (0xE9)
  Max Charge (0xC8)
  Charge Battery from Generator (0x109 Gen4, 0x10C Gen5/Gen6)
    Charge Battery To (0x10A Gen4, 0x10D Gen5/Gen6)
  Schedule 2 (0x104 Gen4, 0x107 Gen5/Gen6)
    unlocks the second Charge Period / Allowed Discharge Period pair
```

Names follow the cloud app now that the device provides the context: `Generator Max Charge` → **Max Charge** (in kW, `0.01` scale, 0–300 kW instead of a raw 10 W register), `Generator Charge` → **Charge Battery from Generator**, `Generator Min Power` → **Hold Min Power**, `Generator Time 2` → **Schedule 2**, `Generator Charge SOC` → **Charge Battery To**. **Keys and entity ids are unchanged.**

Every register mapping was verified live on an X3-Hybrid-G4 15 kW by changing each setting in the SolaX app and tracking the readbacks.


wills106:
Review output of #2285, I think we can ignore the comments about CRLF and mojibake. I think I introduced those errors generating the diff change files in Windows, using the default settings which generate UTF-16 CRLF files:

## Review of 2285_diff.txt (homeassistant-solax-modbus)

### What the diff does
This is the last commit of PR #2285, stacked on #2279. It relocates 33 generator entities from the Dry Contact Box gating to a new `<hub> External Generator` sub-device behind a new `read_gen` option (off by default), renames them to match the cloud app, and adds the `read_gen` translation to all 8 shipped languages. The diff bundles the parent #2279 changes (the 54 PM entities from 2284) so it applies on top of that base.

### Verification results

**Applies cleanly:** The Python files (`__init__.py`, `config_flow.py`, `const.py`, `plugin_solax.py`) and `strings.json`, `de.json`, `en.json`, `nl.json` all apply cleanly. NOTE: the file ships with CRLF line endings, so `git apply` needs them stripped first (`tr -d '\r'`).

**33 generator entities confirmed:** 33 `name="Generator ..."` lines removed, matching the PR's count.

**Keys/entity ids unchanged:** All `key=` values (`generator_max_charge`, `generator_charge`, `generator_time_2`, etc.) and `sensor_key=` values are untouched.

**active_when conditions match the described tree exactly:**
- `generator_control: ("ATS Control", "Dry Contact")` — the parent gate for the whole tree
- `generator_control: ("Dry Contact",)` — Start Gen Method, Max Run Time, Max Rest Time, Hold Min Power, Allow Work Start/Stop Time
- `+ generator_start_method: ("Reference SOC",)` — Switch On/Off SoC (only with Reference SOC)
- `+ generator_charge: (1,)` — Charge Battery To (only when Charge Battery from Generator is on; verified `value_function_enable_disable` stores 1 for on)
- `+ generator_time_2: (1,)` — Schedule 2 / Charge Period 2 (unlocks second period pair)

**Wiring consistency:** `CONF_READ_GEN`/`DEFAULT_READ_GEN` added to const.py; imported in `__init__.py` and `config_flow.py`; registered in `GATED_DEVICE_GROUPS` as `"external_generator": (CONF_READ_GEN, DEFAULT_READ_GEN)`; added to `DEVICE_GROUP_NAMES` as "External Generator"; the option is hidden via `_plugin_supports_device_group` when the plugin doesn't implement it. All consistent with the existing eps/pm pattern.

**read_gen in all shipped languages:** All 8 shipped translation files (cs, da, de, en, fr, it, nl, se) are touched.

**Max Charge fmt i→f and scale/unit change:** Verified correct. The internal sensor `generator_max_charge` (0x132) now has `scale=0.01`, so `hub.data` holds kW. The number entity's `read_scale` defaults to 1 (never reassigned), so `fmt="f"` is required — with `fmt="i"` and a fractional `native_step=0.1`, the read path at number.py:241 would fall through to `float(value)` anyway, but `fmt="f"` is the correct, explicit choice for a 0.01-scaled kW register. Round-trip write (`_scale_native_value_to_register`) is consistent.

### Issue found: translation files contain mojibake
The 5 non-ASCII translation files (cs, da, fr, it, se) **fail to apply** because their non-ASCII content is mojibake — double-encoded UTF-8. The added `read_gen` lines and the surrounding context lines contain corrupted characters (e.g. `Podpora extern├¡ho gener├ítoru` instead of proper Czech diacritics, confirmed via byte inspection: bytes like `e2 94 9c` are box-drawing artifacts, not valid Czech/German/French text). The ASCII-only files (de, en, nl) apply fine.

This is a **corruption in the diff itself**, not a logic error. The English strings ("External Generator Support") are correct, and the structure is right, but the localized strings in those 5 files are garbled and would produce invalid-looking translations if force-applied. The localized `read_gen` strings need to be regenerated from the correct source text before this diff can be merged.

### Verdict
The code changes are correct, complete, and consistent with the PR description and existing patterns. The only defect is mojibake corruption in the 5 non-ASCII translation files, which prevents them from applying and would produce broken localized strings. Fix the localized `read_gen` translations (proper UTF-8) and the diff is ready.

PR #2286, opened 2026-08-13
Hybrid Gen4: Dry Contact sub-device with dependency-driven availability

Bl00d-B0b:
Based on `main` — #2279, which provided the sub-device infrastructure, has since merged, so this is a single self-contained commit.

The 13 Gen4 dry-contact load-management controls become their own HA device (`<hub> Dry Contact`), and each entity only exists while its full parent chain in the app's dependency tree is active:

```
Dry Contact Mode (0xC3 -> 0x12F): Generator Control | Load Management
  Load Management (0xC2 -> 0x12E): Disabled | Manual Mode | Smart Save
    Manual Mode switch (0xB6 -> 0x122): only in Manual Mode
    Threshold On Feed-in Power (0xB7 -> 0x123)  \
    Threshold Off Consumption  (0xB9 -> 0x125)  |
    Threshold Off Battery SoC  (0xBA -> 0x126)  |  only in Smart Save
    Minimum Per On Signal      (0xBB -> 0x127)  |
    Maximum Per Day            (0xBC -> 0x128)  |
    Schedule                   (0xBD -> 0x129)  /
      Work Time P1/P2 Start/Stop (0xBE-0xC1 -> 0x12A-0x12D):
      only while Schedule is enabled
```

Left of the arrow is the write register, right of it the readback. Every mapping was verified live on an X3-Hybrid-G4 15 kW by changing each setting in the SolaX app and tracking which register moved.

The **Dry Contact Box** option now only decides whether the sub-device exists; it no longer sets the `DCB` feature bit on individual entities, and its label drops the model suffix since #2279 hides the option for plugins without dry-contact entities.

Display names follow the app (Work Mode → Load Management, Manual Mode Control → Manual Mode, Feedin On Power → Threshold On Feed-in Power, Consume Off Power → Threshold Off Consumption, Switch Off SOC → Threshold Off Battery SoC, Maximum Per Day On → Maximum Per Day).

The four schedule controls are renamed to **Work Time P1 Start / Work Time P1 Stop / Work Time P2 Start / Work Time P2 Stop**, so each period groups together and sorts Start before Stop (`Sta` < `Sto`). "Stop" matches the register's own name (`work_stop_time`).

**Entity-id note:** all other keys and entity ids are unchanged, but the two readback sensors at 0x12B/0x12D (`work_stop_time_1/2`, disabled by default) become `internal=True` and are therefore removed as entities. This is required, not cosmetic: `time.py` resolves its value from `hub.data[self._key]` and ignores `sensor_key`, so a time control only shows a value when its readback publishes under the same key — which is how the Start side (0x12A/0x12C, already internal) has always worked. Making the Stop readbacks internal under the matching key fixes the Stop controls and mirrors the Start pattern.


wills106:
Review output, still having issues with file encoding...
Was `work_stop_time_1` → `work_end_time_1` intentional, not sure how many people use these entities?

## Review of PR #2286 (Dry Contact sub-device)

### Applies cleanly
The diff is CRLF-encoded. Against the #2279 base it applies cleanly for all Python files, `strings.json`, and ASCII-only translations (de, en, nl). Only the 5 non-ASCII translation files (cs, da, fr, it, se) fail — same mojibake corruption as #2285/#2283 (double-encoded UTF-8). The diff is stacked on #2285, so it does not apply to the current working tree (which already has #2285) — this is expected for stacked PRs.

### Dry Contact sub-device wiring — correct
- `dry_contact` added to `DEVICE_GROUP_NAMES` ("Dry Contact") and `GATED_DEVICE_GROUPS` (`(CONF_READ_DCB, DEFAULT_READ_DCB)`).
- `config_flow.py`: both `dry_contact` and `external_generator` are now gated by `_plugin_supports_device_group` (the loop that was only on `external_generator` in #2285 is now fixed).
- 13 entities get `device_group="dry_contact"` — matches the PR's count.

### active_when parent chains — correct
Verified against the app's dependency tree:
- `dry_contact_mode: ("Load Management",)` — base gate for the whole tree
- `+ work_mode: ("Manual Mode",)` — Manual Mode switch
- `+ work_mode: ("Smart Save",)` — Threshold On/Off entities + Schedule
- `+ work_mode: ("Smart Save",) + schedule: (1,)` — all 4 Work Start/End Time entities (0xBE/0xBF/0xC0/0xC1)
- `schedule` and `manual_mode_control` keys exist as internal sensors (verified).

### Register mappings — correct
The write→readback split is consistent with the existing `work_start_time` pattern (write at 0xBE, readback at 0x12A). The `dry_contact_mode` internal readback already exists at 0x12F and `work_mode` at 0x12E in the working tree (old commits), so the PR's "0xC3 → 0x12F" notation describes the existing split, not a new change.

### DCB feature-bit removal — correct
`DCB` removed from all 83 `allowedtypes` (0 re-added). The `read_dcb`/`invertertype | DCB` block in `async_determineInverterType` is removed alongside the `CONF_READ_DCB`/`DEFAULT_READ_DCB` imports, so the import removal is safe (verified no other usage remains).

### Label change — correct
`read_dcb` label drops the model suffix across all 8 languages ("Dry Contact Box (SolaX Hybrid Gen4 & Gen5)" → "Dry Contact Box").

### f-strings — none introduced
Zero added f-strings inside `_LOGGER(...)` calls, and zero added f-strings anywhere in the diff (excluding `fmt=`).

### One discrepancy to flag (not a blocker)
The PR states "Keys and entity ids are unchanged," but the diff renames the sensor keys `work_stop_time_1` (0x12B) → `work_end_time_1` and `work_stop_time_2` (0x12D) → `work_end_time_2`. This is intentional and consistent with the `work_start_time` pattern (the 0xBF/0xC1 write entities already use `work_end_time_1/2`), so the rename aligns the readback key with the write key. However:
- The unique_id is `{platform}_{key}`, so the renamed sensor (sensor platform) and the existing 0xBF time entity (time platform) do **not** collide on unique_id.
- Both now share `hub.data["work_end_time_1"]` — but that's the intended behavior (matching the write entity), same as the existing `work_start_time` pattern.

So this is a deliberate, consistent change rather than a bug, but the PR description is slightly inaccurate — the keys for these two readback sensors **do** change. Worth either correcting the PR text or confirming this is intended.

wills106:
> Was `work_stop_time_1` → `work_end_time_1` intentional, not sure how many people use these entities?

@Bl00d-B0b once you have had a look at the above comment I am ready to merge and do a new release containing all of these backlogged PR's I have been approving.

If you need more time, let me know and I will release the other changes and then release this PR at a later date.

Bl00d-B0b:
I make this decision. But of course I can rework. For rework I need next week. Currently on holidays. But this is a trick. To be sure Start time be first in sort order before End. I use "E" from Cyrillic alphabet. As result IDs I do not touch but now in GUI I see first Start time then End Time. Works for me. If you do not allow then I brewer it back.

wills106:
Won't they be ordered like:
"Work End Time 1"
"Work End Time 2"
"Work Start Time 1"
"Work Start Time 2"

As per https://github.com/wills106/homeassistant-solax-modbus/discussions/2140
Or did you actually want the End Times first?

Bl00d-B0b:
Thanks for the careful review — and sorry for the delay, I was away.

Two corrections to what I wrote earlier, after testing this properly on my hardware.

**1. There is no Cyrillic character in this PR.** I described an approach I had experimented with previously and then abandoned; it never made it into the branch. I tested it again this week to be sure, and it is genuinely a bad idea: Home Assistant's slugifier strips the non-ASCII letter, so `Work Time P1 Еnd` produced the entity id `..._work_time_p1_nd`. Please disregard that comment.

**2. The key rename was not cosmetic — it was required.** `time.py` looks up `hub.data[self._key]` directly and ignores `sensor_key`, so a time entity only shows a value if its readback sensor publishes under the *same* key. The Start controls worked because the internal sensor at 0x12A already used `work_start_time_1`; the End controls had no matching publisher. I verified this by reverting the rename: both End entities immediately went to `unknown`, and restoring the shared key brought them back to `23:59:00`.

**What I changed in response to your question**

- The readback sensors at 0x12B/0x12D are now `internal=True`, exactly mirroring their 0x12A/0x12C siblings on the Start side. They keep feeding the controls, but they are no longer entities — so there are no duplicate display names, and the disabled-by-default sensors that previously carried the `work_stop_time_*` keys are simply gone rather than renamed.
- The four time controls are renamed to **Work Time P1 Start / Work Time P1 Stop / Work Time P2 Start / Work Time P2 Stop**.

That naming answers your sort-order question without any trickery: `Sta` sorts before `Sto`, so each period reads Start then Stop, grouped together, in plain ASCII with clean entity ids (`..._work_time_p1_start`, `..._work_time_p1_stop`). "Stop" also matches SolaX's own register name (`work_stop_time`), so the vocabulary is the vendor's rather than mine.

To answer the question you actually asked: no, I did not want the End times listed first — that was the problem I was trying to solve, and Start/Stop solves it properly.

**Tested on hardware** (X3 Hybrid 15kW, Gen4, HA 2026.9.1):

```
Work Time P1 Start = 00:00:00     Work Time P1 Stop = 23:59:00
Work Time P2 Start = 00:00:00     Work Time P2 Stop = 23:59:00
```

The PR description has been corrected accordingly — the sensor keys **do** change, and I have described what happens to them.

Ready to merge from my side, so please go ahead with the release whenever it suits you.

PR #2299, opened 2026-08-31
Add X3-MIC Gen1 lock controls

### LINKED ISSUE #2282: [Bug]: X3-MIC Gen1 — all writes silently fail because Lock State is not exposed for GEN1
## Summary
On the X3-MIC Gen1, writes to the settings registers are rejected while `0x600` LockState is at its default `0`. They succeed once `0x600` is set to `6868` ("Unlocked - Advanced"). `2014` ("Unlocked") is **not** sufficient on this model.

The `lock_state` select is currently limited to `MIC | GEN2 | GEN4`, so Gen1 users cannot unlock the inverter from within the integration. The `System On` / `System Off` buttons and the `PV Limit` number entity are created and appear to work, but nothing reaches the inverter.

Request: add `GEN` to `allowedtypes` of the `lock_state` select (`register=0x600`), with an `initvalue` so it does not depend on a read-back register that Gen1 does not have.

## Environment
| | |
|---|---|
| Inverter | X3-MIC Gen1, 5 kW |
| Serial | `MU502T…` (returned byte-swapped as `UM05T2…` when read from `0x300`) |
| Firmware | DSP 1.40 / ARM 1.36 |
| Connection | RS485 → Ethernet gateway, Modbus RTU over TCP, port 8899, unit 1 |
| Integration version | 2026.07.1 |
| Home Assistant | Core 2026.7.2, Supervisor 2026.07.5, HA OS 18.1 |

## Steps to reproduce
1. Leave `0x600` at its default value of `0` (there is no way to change it for Gen1 in the UI).
2. Press `System Off`, or set `PV Limit` below 100.
3. Nothing happens. The read-back mirror at `0x332` keeps its previous value and the inverter stays in Run Mode 2.

## Evidence
All tests were done with a standalone pymodbus script and the integration disabled, to keep a single master on the bus. Each write was retried up to six times with 2.5 s between attempts.

Registers that read fine on this firmware:

```
0x332 PV Limit mirror = 100
0x33D Firmware DSP    = 140
0x33E Firmware ARM    = 136
0x40F Run Mode        = 2 (Normal Mode)   [input register]
```

Reads of `0x600`, `0x60F` and `0x610` return exception 2 (illegal data address). That appears to be normal for this model: the `0x6xx` registers are write-only, with read-back mirrored in the `0x3xx` range.

Write attempts, each writing back the value that was already active, so no actual change and no EEPROM cycle:

| Attempt | Result |
|---|---|
| `0x610` = 1, FC 0x06 | exception 6 (slave device busy) |
| `0x610` = 1, FC 0x10 | exception 6 |
| `0x60F` = 100, FC 0x06 | exception 6 |
| `0x60F` = 100, FC 0x10 | exception 6 |
| `0x60F` = 100 after `0x600` = 2014 | exception 6 |
| **`0x60F` = 100 after `0x600` = 6868** | **accepted** |

Worth noting for anyone debugging this later: a locked inverter answers with exception 6 (busy), not exception 4 or 3. That makes it look like a timing problem, which sends you down the wrong path. Writes to `0x600` itself are accepted while the inverter is locked, so only the registers behind the lock are affected.

Verification that the write actually lands, run twice, accepted on the first attempt both times:

```
unlock 0x600 = 6868 → accepted
write  0x60F = 99   → accepted, mirror 0x332: 100 → 99
write  0x60F = 100  → accepted, mirror 0x332: 99  → 100
```

So ARM 1.36 does have this register set. The blocker was purely the lock level.

## Suggested change

`custom_components/solax_modbus/plugin_solax.py`, in `SELECT_TYPES`:

```python
        name="Lock State",
        key="lock_state",
        register=0x600,
        option_dict={
            0: "Locked",
            2014: "Unlocked",
            6868: "Unlocked - Advanced",
        },
        allowedtypes=MIC | GEN | GEN2 | GEN4,   # GEN added
        initvalue=0,                            # Gen1 has no read-back register
        icon="mdi:lock-question",
    ),
```

If `initvalue` on a non-`WRITE_DATA_LOCAL` select is unwanted, a separate entry for `MIC | GEN | X3` would work just as well — whichever fits the plugin's conventions better.

## No read-back register exists on Gen1

The existing `lock_state` read-back sensors both return exception 2 on this inverter:

```
holding 0x367 (MIC | GEN2 | X3)        → exception 2, illegal data address
holding 0x39A (MIC | GEN2 | GEN4 | X1) → exception 2, illegal data address
```

So no read-back sensor should be added for `MIC | GEN` — it would only get quarantined, which is what #2133 describes for `0x39A` on the X1-Boost Gen3.

Since `current_option` falls back to `entity_description.initvalue` when the key is absent from `hub.data`, adding `initvalue=0` gives the select a sensible starting state ("Locked") without any read attempt. The displayed value then reflects what was last written from HA rather than the inverter's actual state, which seems acceptable given the register cannot be read at all.

## Possibly related

- #2231 `[Bug]: Device Lock`
- #2133 X1-Boost Gen3, control not working, `lock_state 0x39a`
- #1449 X3-Hybrid G4 cannot be unlocked after firmware upgrade

Not duplicates as far as I can tell — different models and generations — but lock handling seems to be a recurring theme.

## Documentation suggestion

It may be worth stating in the docs that `Lock State` has to be `Unlocked - Advanced` rather than `Unlocked` before settings can be written. On this Gen1 unit `2014` was consistently rejected, and there is nothing in the FAQ about it. I do not know whether Gen2 and Gen4 behave the same way.

### Integration Version

2026.07.1

2026.7.2

### Inverter brand

Solax Power

### Plugin used

plugin_solax.py

### Serial prefix

MU502T

DSP 1.40 / ARM 1.36

### Connection Method

S485 → Ethernet gateway, Modbus RTU over TCP, port 8899

### Dongle firmware

Not using dongle but a PUSR RS485 to WiFi Converter Modbus Gateway

### Detailed Error Log

_No response_

### Additional context

_No response_

rosenrot00:
SolaX X3-MIC Gen1 inverters reject setting changes while locked. Unlike newer MIC generations, they do not expose a readable lock-state register and require the advanced unlock value `6868` to be written to register `0x600`.

- Add Gen1-only **Unlock Advanced Settings** and **Lock Settings** buttons
- Use the write-only register `0x600`
- Allow changing **PV Limit** and using **System On/Off** after unlocking
- Keep the existing GEN2/GEN4 lock-state controls unchanged

Fixes #2282


wills106:
Confirmed working in https://github.com/wills106/homeassistant-solax-modbus/issues/2282#issuecomment-5480298395

Issue #2302, opened 2026-09-02
[Bug]: Unable to select Modbus Power Control (direct) option


richardmx5guys:
### Describe the bug

Returns an error when selecting a Modbus Power Control (direct) option from Disabled to Enable SOC Target Control Mode.

If an option for Modbus Power Control (direct) is made without reloading SolaX Integration first, it will return an error as above.
But, the change was actually made on an Inverter and so home assistant will show incorrect value of the Modbus Power Control (direct) option.

BUT! if I reload SolaX integration prior making a Selection of Modbus Power Control Mode (direct) then no error is returned, the change is made in the inverter and no ERROR is returned.

### Integration Version

2026.08.2

### Homeassistant core version

2026.8.3

### Inverter brand

Solax X1-G4 Hybrid

### Plugin used

plugin_solax.py

### Serial prefix

H4372A

### Inverter firmware versions

ARM 1.52-1.05 DSP 1.57

### Connection Method

SolaX WiFi+LAN 2.0

### Dongle firmware

006.06

### Detailed Error Log

Error: SolaX: multi-function single-register write was rejected by device 1 at register 0x7c: ExceptionResponse(dev_id=1, function_code=144, exception_code=4)

2026-09-02 22:15:16.356 INFO (MainThread) [custom_components.solax_modbus.select] writing SolaX select register 124 value 0 with method 2
2026-09-02 22:15:16.668 ERROR (MainThread) [homeassistant.components.automation.solax_soc_control_on_off] SolaX SOC Control ON OFF: Choose at step 1: Turn SOC OFF: Error executing script. Error for device at pos 3: SolaX: multi-function single-register write was rejected by device 1 at register 0x7c: ExceptionResponse(dev_id=1, function_code=144, exception_code=4)
2026-09-02 22:15:16.669 ERROR (MainThread) [homeassistant.components.automation.solax_soc_control_on_off] SolaX SOC Control ON OFF: Error executing script. Error for choose at pos 1: SolaX: multi-function single-register write was rejected by device 1 at register 0x7c: ExceptionResponse(dev_id=1, function_code=144, exception_code=4)
2026-09-02 22:15:16.672 ERROR (MainThread) [homeassistant.components.automation.solax_soc_control_on_off] Error while executing automation automation.solax_soc_control_on_off: SolaX: multi-function single-register write was rejected by device 1 at register 0x7c: ExceptionResponse(dev_id=1, function_code=144, exception_code=4)

### Additional context

[...]

PR #2315, opened 2026-09-05
Fix SolaX direct VPP register block writes

rosenrot00:
Send complete, mode-specific FC16 register blocks for SolaX direct VPP controls instead of individual register writes.

Parameter changes continue to apply immediately while the relevant mode is active. Read the current inverter mode before updates to avoid reactivating expired commands. Preserve the selected Set/Update behavior and leave autorepeat controls unchanged.

Addresses #2302


rosenrot00:
Wait before merging this one, we need to verify the issues around VPP as this PR would break automations as it sends all relevant direct registers in one go.

PR #2323, opened 2026-09-07
SolaX: tidy entity names in the EPS, Parallel and External Generator sub-devices

Bl00d-B0b:
> Stacked on #2286. Review the last commit only — the diff includes the parent until it merges.

Cosmetic follow-up to the already-merged sub-device PRs (#2283 EPS, #2284 Parallel, #2285 External Generator): their display names are brought in line with the conventions used elsewhere in the plugin.

**No functional change.** Keys, entity ids of sensors/numbers/selects/switches, option values, register mappings and gating are all untouched — nothing an automation can reference moves.

### EPS (3)

`Min SOC` → `Min SoC`, `Restart SOC` → `Restart SoC`, `Min Esc SOC` → `Min Esc SoC` — the state of charge is spelled `SoC` everywhere else in the integration.

### Parallel (18)

Glued words gain their space, for the shared entities and their per-inverter I2/I3 variants:

| Before | After |
| --- | --- |
| `ActivePower L1/L2/L3` | `Active Power L1/L2/L3` |
| `I2 ActivePower L1/L2/L3` | `I2 Active Power L1/L2/L3` |
| `I3 ActivePower L1/L2/L3` | `I3 Active Power L1/L2/L3` |
| `Reactive or ApparentPower L1/L2/L3` | `Reactive or Apparent Power L1/L2/L3` |
| `I2 / I3 Reactive or ApparentPower L1/L2/L3` | `I2 / I3 Reactive or Apparent Power L1/L2/L3` |

### External Generator (11)

The schedules adopt the `<Function> [P<n>] Start|Stop` convention:

| Before | After |
| --- | --- |
| `Allow Work Start Time` / `Allow Work Stop Time` | `Allow Work Start` / `Allow Work Stop` |
| `Charge Period Start Time` / `Charge Period End Time` | `Charge P1 Start` / `Charge P1 Stop` |
| `Charge Period 2 Start Time` / `Charge Period 2 End Time` | `Charge P2 Start` / `Charge P2 Stop` |
| `Allowed Disc Period Start Time` / `... End Time` | `Discharge P1 Start` / `Discharge P1 Stop` |
| `Allowed Disc Period 2 Start Time` / `... End Time` | `Discharge P2 Start` / `Discharge P2 Stop` |
| `Schedule 2` | `Schedule P2` |

Three reasons for that shape:

- **`Stop` rather than `End`**, so each pair sorts Start-then-Stop in an alphabetical list (`Sta` < `Sto`). It is also what the registers are called (`generator_stop_time`).
- **No `Time` suffix** — the entity is already a time.
- **`P<n>`**, so a period's entities group together instead of every Start grouping before every Stop.

The wordy prefixes are dropped because the **External Generator** sub-device already supplies that context. The shortened names do not collide with the battery schedules on the main device, which are `Force Charge P1 Start` and `Allowed Discharge P1 Start`.

**Entity ids:** the External Generator time entities are not registry-backed in this integration, so their ids follow the display name (`time.<hub>_charge_period_start_time` → `time.<hub>_charge_p1_start`). That is unavoidable for any time-entity rename. Every other entity keeps its id, since those come from the keys, which do not change.

Verified on hardware (X3 Hybrid 15kW Gen4, HA 2026.9.1).


Bl00d-B0b:
Consolidated into #2325, which now covers both the decoding fixes and the display-name tidy-up for the same three sub-devices (EPS, Parallel, External Generator). That PR is based directly on main, so it no longer carries the Dry Contact commit this one did and can be merged independently.

PR #2326, opened 2026-09-08
SolaX: match sensor rounding to the register scale

Bl00d-B0b:
> Stacked on #2325. Review the last commit only — the diff includes the parent until it merges.

A sensor's value is `round(raw * scale, rounding)`. Where `rounding` is finer than the scale it can never produce those digits; where it is coarser it throws away digits the register actually provides. Since #2275 the sensor platform also derives Home Assistant's **display precision** from the same `rounding`, so a mismatch is now visible in the UI, not just in the stored value.

**30 sensors across three plugins** are brought in line with their register scale:

| Scale | Rounding |
| --- | --- |
| 0.1 | 1 |
| 0.01 | 2 |
| 0.001 | 3 |

### Decimals that were being discarded

The register carries them; the sensor rounded them away.

| Entity | Scale | Rounding |
| --- | --- | --- |
| Battery Voltage Charge · Battery Current Charge | 0.01 | 1 → **2** |
| Battery Charge Float Voltage · Battery Discharge Cut Off Voltage | 0.01 | 1 → **2** |
| Inverter Frequency L1 / L2 / L3 | 0.01 | 1 → **2** |
| EPS Yield Total | 0.01 | 1 → **2** |
| Battery Input / Output Energy Total · Total Solar Energy | 0.001 | 1 → **3** |
| Today's Yield (MIC GEN) | 0.001 | 2 → **3** |
| PowerFactor (LV) · Power Factor (Mega/Forth) | 0.001 | 1 → **3** |

A 50.00 Hz inverter frequency, for example, was being reported to one decimal despite the register providing hundredths.

### Decimals that could never exist

`round(raw * 0.1, 2)` cannot produce a second decimal, so asking for one only misleads the display layer.

| Entity | Scale | Rounding |
| --- | --- | --- |
| Today's Solar Energy · Today's Yield · Total Yield | 0.1 | 2 → **1** |
| E Charge Today · E Charge Total | 0.1 | 2 → **1** |
| Total Grid Export · Total Grid Import | 0.1 | 2 → **1** |

One of these carried the comment `# GEN4 | GEN5 might be 1`. With a scale of 0.1 no second decimal can ever be produced, so **1 is correct for every generation** — the uncertainty in that comment is resolved by the arithmetic rather than by hardware.

### Left alone

`Time Count Down` keeps its explicit `rounding=0`. An integer countdown is a deliberate display choice, not a mismatch, so explicit zeros are preserved rather than overwritten.

No register mappings, scales, keys or entity ids change — only the number of decimals each sensor reports.


Bl00d-B0b:
Splitting these into the existing PRs instead: the EPS entity goes to #2325 alongside the other EPS decode fixes, and the main-device, LV and Mega/Forth entities go to #2277 with the rest of the presentation work. Closing to avoid a fourth open PR touching the same file.

PR #2327, opened 2026-09-08
Fix via_device_id TypeError on HA < 2026.8 (retro-compatible fallback)

SimoneB79:
Follow-up to #2308, fixes the `via_device_id` breakage on HA < 2026.8.0 reported by @TCWORLD ([comment](https://github.com/wills106/homeassistant-solax-modbus/pull/2308#issuecomment-5590775788)).

## Problem

The fallback logic introduced in #2308 chooses `via_device_id` vs `via_device` based on whether the parent device is **found**. But on HA < 2026.8 the legacy `async_get_device` lookup still works (the parent is always found), so the code sets `via_device_id` — a keyword `DeviceRegistry.async_get_or_create()` only accepts from HA 2026.8.0 onwards:

```
TypeError: DeviceRegistry.async_get_or_create() got an unexpected keyword argument 'via_device_id'
```

Every sub-device fails to register (Energy Dashboard, dry contacts, battery packs) on e.g. HA 2026.7.4.

## Fix

Decide based on whether the registry API **supports** the keyword, not on lookup success:

- `supports_via_device_id()` — one-shot capability probe via signature inspection (version-proof, no version-string parsing)
- `link_parent_device()` — shared helper used by all three call sites:
  - `sensor.py` (battery packs)
  - `__init__.py` (`group_device_info`)
  - `energy_dashboard.py` (`create_energy_dashboard_device_info`, `hass is None` guard kept)

HA ≥ 2026.8 keeps the scoped `via_device_id` behaviour; older versions fall back cleanly to the deprecated-but-working `via_device` tuple. Also removes the small logic duplication left by #2308.

## Test

Adds `test_supports_via_device_id_detection` to the existing regression suite.

### DIFF
--- a/custom_components/solax_modbus/__init__.py
+++ b/custom_components/solax_modbus/__init__.py
@@ -162,7 +162,7 @@ from .const import (
 from .const import (
     matches_active_when as matches_active_when,
 )
-from .device_registry_lookup import get_device_by_identifier
+from .device_registry_lookup import link_parent_device
 from .modbus_transport import CoreModbusTransport, ModbusTransport, NativeModbusTransport, UnavailableModbusTransport
 from .pymodbus_compat import DataType, convert_from_registers, convert_to_registers, pymodbus_version_info
 from .sensor import SolaXModbusSensor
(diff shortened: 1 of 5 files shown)


TCWORLD:
I can confirm this resolves the compatibility issue - with this patch applied the integration now loads again correctly on 2026.7.4.

Fixes #2328.

Issue #2179, opened 2026-07-09
[Bug]: Solis - S6-EH1P10K-L-PLUS (FB00) - single-register writes to timed discharge/charge SOC, current & volt don't persist


Kriskit:
Been chasing this for a while with Predbat and finally worked out what's going on, so figured I'd write it up in case anyone else hits it.

### Environment
- **Integration version:** 2026.06.7
- **HA core:** 2026.7.1 (HAOS)
- **Inverter:** Solis S6-EH1P10K-L-PLUS (hybrid), ~1-2 months old, current firmware as of July 2026
- **Plugin:** `plugin_solis_fb00.py`
- **Connection:** Waveshare RS485-to-Ethernet adapter

(Side note — the "Non SolaX Inverter" bug template dropdown only lists `plugin_solis.py` / `plugin_solis_old.py`, there's no `plugin_solis_fb00.py` option, so I've filled this in manually.)

### Describe the bug
On this FB00 firmware the per-slot config number entities that write to a single register don't actually stick. For discharge slot 1:

- 43750 — Timed Discharge SOC
- 43751 — Timed Discharge Current
- 43752 — Timed Discharge Volt

If I set any of these (either from the HA UI or via `number.set_value`), the entity shows the new value for a second or two, then the next poll reads the old value straight back over the top. The charge slot behaves the same way.

The time fields (start/end hours/minutes) are fine because those go through the "Update Discharge Times" button, which does a block write (FC 0x10) starting at 43753. The SOC/current/volt registers only ever get a single-register write (FC 0x06), and the FB00 firmware seems to just ignore those — it ACKs on the bus so HA thinks it worked, but nothing changes internally.

### Steps to reproduce
1. Write SOC 20 → shows 20 → reverts to the previously committed value within 1–2 poll cycles.
2. Write SOC 20, then press "Update Discharge Times" → still reverts. The button's `value_function` only pushes the four time registers, so it never touches 43750.
3. Same story for the current register (43751) — stuck at whatever was last committed.

So it looks like the FB00 firmware wants these written as part of a block rather than poked individually. The Solis Modbus notes floating around also warn against single-register access on these, which lines up.

### Where it bites
Anything driving the inverter programmatically (Predbat in my case) can set the discharge/charge *times* but can't actually set the SOC floor or the current/rate — those silently stay at whatever was last committed manually.

### Possible fix
[...]

PR #2330, opened 2026-09-09
Solis FB00: Remote Dispatch support (Ver3.4 44100 block) — full-depth discharge on fw 4F0052

philicibine:
## Solis FB00: Remote Dispatch support (Ver3.4 44100 block) — live-verified full-depth discharge on fw 4F0052

Adds HA-native support for the Solis **Remote Dispatch** protocol (real-time dispatch block, protocol Ver3.4, registers 44100-44112) to `plugin_solis_fb00.py`, following the existing timed charge/discharge "local numbers + block-write button" pattern.

### Why

Issue #2179 established that firmware 4F0052 enforces the locked timed-discharge SOC (43750=50) as a hard floor on **every** scheduled/forced export path. Remote Dispatch is a separate RAM-only control path that **bypasses that floor** — live-verified: dispatch `battery discharge` at 1-2kW sustains below 50% SOC down to the commanded soc_min (tested at 43-49%), while timed slots and RC force stop dead at 50%. It is the only full-depth export path on affected firmware. Dispatch-capable inverters expose the block when input 34502 reads 0xAA55.

### What's added (FB00 hybrid only)

**Buttons**
- `Apply Remote Dispatch` — block-writes the realtime block 44105-44112 (mode, S32 power, function, SOC window) from the staged local entities
- `Disable Remote Dispatch` — single-block stop (master off, failsafe back to default); live-verified to leave the block at clean state with status 34504 → 0

**Config entities**
- `Dispatch Control Mode` select (Battery Hold / Battery Charge / Grid Import / Grid Port Import / Self Consumption / Feed-in Priority — direction from signed power)
- `Dispatch Power` number (W, signed; negative = discharge/export; raw = watts/10 per protocol)
- `Dispatch Failsafe Interval` number (minutes; the inverter self-reverts if no fresh write arrives)
- `Dispatch SOC Minimum` / `Dispatch SOC Maximum` numbers
- `Dispatch Master` switch (44100)

**Diagnostic sensors** (disabled by default): capability flag (34502, input), running status (34504, input), and readbacks of master/mode/power/SOC window.

### Live verification (S5-EH1P6K-L, fw 4F0052, this inverter on the bench)

Driven purely through HA services, battery at 43-44% SOC (below the locked 50 floor):

1. Stage mode=Battery Charge, power=−1000W, soc_min=30, failsafe=10 → `number.set_value` / `select.select_option`
2. `Dispatch Master` on → register 44100 = 1
3. Press `Apply Remote Dispatch` → block lands exactly: `[1, 5, 0, 0xFFFF, 0xFFFF, 2, 0xFFFF, 0xFF38, 85, 30, 100, 0, 0]`, dispatch status 34504 = 2 (running)
4. Battery discharges at commanded power (~1kW incl. house load), grid exports the surplus — sustained below the 50% floor
5. Press `Disable Remote Dispatch` → block reverts to defaults `[0, 5, 0, 0xFFFF, 0xFFFF, 1, ...]`, status → 0, battery resumes self-use

Also fixed en route: a switch with no `value_function` silently no-ops in `_write_switch_to_modbus` (line 214 returns early) — the master switch now uses a direct on/off value function.

### Notes for maintainers

- The block **must** be written as contiguous FC16 writes, global before realtime ("the function field is re-initialised unless the master is already on") — hence the master-switch-then-apply-button flow rather than a single service. The same ordering is documented in Pho3niX90/solis_modbus, whose protocol notes matched this firmware exactly (credit in code comments).
- Capability gating: entities are HYBRID-only for now; the capability sensor (34502) is the honest gate — non-dispatch-capable inverters read 0 and the buttons simply won't do anything useful. Happy to add `active_when` gating on the capability sensor if preferred, or a select for charge vs discharge direction as separate options.
- The schedule block (44116+, 6 inverter-resident periods) is **not** included yet — kept this PR minimal; can follow up if there's interest.

Closes nothing by itself (the 43750 write-lock needs a Solis fw fix), but gives 4F0052 owners a working full-depth path and everyone else a real-time control surface.

Link to docs: #2179 (issue trail with the full gate matrix and test data).


custom_components/solax_modbus/plugin_solis_fb00.py
```diff
@@ -175,7 +175,64 @@ def value_function_battery_control_override(initval: int, descr: Any, datadict:
...
+    6: "Feed-in Priority",
+}
+# Modes 2/3/4 encode battery charge/discharge and grid import/export as register
+# values 2/3/4; direction comes from the sign of the S32 power value (negative =
+# battery discharge / grid export). The select exposes merged names.
+DISPATCH_MODE_OPTIONS_INV = {v: k for k, v in DISPATCH_MODE_OPTIONS.items()}
+
+
 # ============================================= Charging ===========================================================
+
```
wills106, line 233: Can you remove one of these Blank lines. The CI/CD is failing as there is 3 blank lines, the standard is 2.
Thanks

